PR #12, opened 2014-10-03
Add test for enclosed-area() function

ukanga:
Do not merge yet: non standard ODK Validate jar.


MartijnR:
I believe this can be closed. (area() was added to public JavaRosa and ODK Validate)

pld:
Does the JAR version in pyxform need to be upgraded?

pld:
Was it already?

MartijnR:
Not sure actually. I can't tell from the ODK Validate history in which commit it was added. However, both the XLSForm Offline Desktop App and http://opendatakit.org/xiframe support the area() function.

MartijnR:
pyxform in KoBo doesn't so that makes it a bit suspicious.

ukanga:
The jar version is at 1.4.6, so it may not need to be upgraded. Will rebase and see what changes if any need to be applied.

ukanga:
I dropped the custom jar and rebased with master, tests are passing.

PR #28, opened 2015-02-24
Adding a pyxform_test_case and "tests_v1/" directory

dorey:
This will be useful for testing the inputs and outputs of this
tool as strings, avoiding the confusion of having complicated
features buried in fixtures and XLS files in the repository.

More examples of the PyxformTestCase _in action_ coming soon!

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -50,6 +50,8 @@ class Survey(Section):
     )
 
     def validate(self):
+        if self.id_string in [None, 'None']:
+            raise PyXFormError('Survey cannot have an empty id_string')
         super(Survey, self).validate()
         self._validate_uniqueness_of_section_names()
 
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -432,7 +432,7 @@ class Survey(Section):
             return result, not result == xml_text
         return text, False
 
-    def print_xform_to_file(self, path="", validate=True, warnings=None):
+    def print_xform_to_file(self, path=None, validate=True, warnings=None):
         """
         Print the xForm to a file and optionally validate it as well by
         throwing exceptions and adding warnings to the warnings array.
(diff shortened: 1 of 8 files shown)


MartijnR:
:thumbsup:

dorey:
Notes on the PyxformTestCase pull request--

This PR changes very little in the main pyxform code, but it allows us to change how we test the functionality by writing tests with [readable string markdown input](https://github.com/XLSForm/pyxform/commit/5dd9837eeeabab81bcae3077e345e1a6f1dd02fc#diff-3de307fc85dda2dbece176d156153627R23), and a number of pattern matching assertions on the output.

I have created a new branch with a few examples of tests which have been rewritten in this format:
 * [Rewriting geo widgets tests](https://github.com/XLSForm/pyxform/commit/f61515268a374aeda414d6ec33e0fb904a67caf5)
 * [For loops](https://github.com/XLSForm/pyxform/commit/99e637a98f05ded961ede2727e8efa1d4520de24)
 * [Testing caps of labels](https://github.com/XLSFortestsm/pyxform/commit/0d90be2c8448b2915f3333e82a749196c8fba6cc)

The goal is to get away from having a number of tests which rely on separate files (xls, xml, csv) and tend to give confusing failure messages when you try to make small changes. This has been a problem since the earliest pyxform tests, as @amarder and @nathanathan can attest.

I haven't built this out to support external datasets yet (or any zip file output) but if we agree this is a good direction for the tests to go, we could extend it in the future.

Thoughts, @prabhasp @ukanga @yanokwa @mberg @MartijnR ?

dorey:
One observer has suggested using a slightly different markdown syntax for the tables that looks more like:

    # sheetname
    | col1 | col2 | col3 |
    |------+------+------|
    | val1 | val2 | val3 |

The important thing is that plug-ins (like [this for sublime-text](https://github.com/vkocubinsky/SublimeTableEditor)), are available to do the formatting, and that devs are happy to read and write it. This syntax will **only** be used in tests.

MartijnR:
I can see it would make sense to go for the full (unofficial) markdown syntax for tables. I believe the `+` should be a `|` though in most implementations (e.g. GitHub).

ukanga:
:+1: Looks good

PR #35, opened 2015-04-23
Support or_other with choice_filter

### LINKED ISSUE #17: Support select one or_other when we have a choice_filter

nathanathan:
This resolves #17

The fix may result in the instance containing an unused choice list because it does not attempt to remove the original choice list without the other option if all the questions that use it include an other option.

This PR also changes the "specify other" question to include the original question label, if the label has no translations. E.g.

```
Which country are you from?
Specify other:
```

Instead of just:

```
Specify other.
```


ukanga:
@nathanathan this seems to only show the `other` choice only when someone skips `state` and `county`, is that the intention?
This seems to work when I add other column for each `state` and `county` combination.
image
Should that always be the case?

nathanathan:
The test form I added is not a good example. To always see the other option, it would be better to add a `"or name="other"` condition to the choice filter rather than modify the choice list.
I considered automatically appending that condition to the choice filter, so the other option would always show, but then there would be no way to filter it out. On the other hand, someone could create their own other choice if they need to filter it out. Which way of doing it makes more sense to you?

ukanga:
I am not keen on `or name="other"` option on the choice_filter, the simple `or_other` should be all that is required. Both cases do require that other is available as an option for each likely choice filter, note the `county` and `state`  which are part of the choice filter.
```xml
          <item>
            <itextId>static_instance-cities-11</itextId>
            <county>king</county>
            <state>texas</state>
            <name>other</name>
          </item>
```

KeynesYouDigIt:
its 4 years old ATP - should we close this? or rebase and propose again? @ukanga

yanokwa:
Closing until we have a clear description of or_other current behavior and what desired behavior in the context of a dynamic choice list is. https://github.com/XLSForm/pyxform/issues/369

PR #37, opened 2015-05-06
add type=file

### LINKED ISSUE #33: Support PDF File Upload question type

ukanga:


ukanga:
@MartijnR @cagulas you can now test this on stage, deployed the branch.

cagulas:
Ok, I am able to upload pdf files, only pdf files.  Is that the intention?  Am I supposed to be able to upload other types of files as well, like .doc, etc.?

ukanga:
You should be able to upload any type of file, the implementation does not restrict the type of file to be uploaded. That restriction will be something that would be controlled by the aggregate server.

cagulas:
Ok, then, it's not functioning correctly, because only pdf files can be uploaded.  All other file types are greyed out and cannot be selected for upload.

MartijnR:
Ah thanks. A long time has passed since this was feature was developed.... It seems `application/*` no longer works in Chrome (still works in FF). We should find another media type.

MartijnR:
I updated the issue (#33) with a new proposed mediatype value.

PR #41, opened 2015-07-20
Previously, a badly formed calculate produced an unhandled error in

jkpr:
the validation stage. This error is now handled so that the program
does not crash unexpectedly. The problem was implicitly decoding a non-ASCII
character with the ASCII codec.

Furthermore, using `xls2xform.py` as main does not crash with a badly formed
calculate. Errors are handled and informative messages are displayed to the
user.

Test files that produce the error are added, both .xlsx and the corresponding
.xml. A test case that checks for this error is added.
 Changes to be committed:
    modified:   pyxform/odk_validate/**init**.py
    new file:   pyxform/tests/bug_example_xls/bad_calc.xlsx
    modified:   pyxform/tests/odk_validate_tests.py
    new file:   pyxform/tests/test_output/bad_calc.xml
    modified:   pyxform/xls2xform.py


jkpr:
I was unable to run all the test cases. I tried to run `nosetests` and there was an immediate error. Something about `ImportError: No module named 'survey'. I tried to run the relevant test cases individually.

dorey:
Note @lognaturel -- this closed PR is following @jkpr 's `master`, so it's got a lot more changes than were originally intended. I think most of their changes were in [this commit](https://github.com/XLSForm/pyxform/pull/41/commits/05dd2d021ee3cb073ce9d8cb0cb7c8dc5f77de4b)

lognaturel:
Ah, makes sense! Thanks, @dorey.

PR #42, opened 2015-07-24
Support writing settings attributes to first node inside instance

yanokwa:
We'd like add a 'attribute::something' to the settings sheet and have that something show up in the XML.

form_id  |  attribute::my_number  |  attribute::my_string
------------ | ------------ | -------------
my_form | 1234567890 | lorém ipsum

```xml
<instance>
  <test id="attribute_test" my_number="1234567890" my_string="lorém ipsum" />
</instance>
```

### DIFF
--- a/pyxform/constants.py
+++ b/pyxform/constants.py
@@ -27,6 +27,7 @@ DEFAULT_LANGUAGE = u"default_language"
 LABEL = u"label"
 HINT = u"hint"
 STYLE = u"style"
+ATTRIBUTE = u"attribute"
 
 BIND = u"bind"  # TODO: What should I do with the nested types? (readonly and relevant) # noqa
 MEDIA = u"media"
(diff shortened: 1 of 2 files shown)


yanokwa:
@ukanga and @MartijnR. This look like a reasonable addition to XLSForm to you?

MartijnR:
yep :+1: for the feature, for developers or to enable private XForm features like the existing bind:: feature does (I'm pretty sure that's your intention but mention it just in case).

I didn't look at pyxform code. 

So, we agree that this should stay out of the xslform.org et. al. documentation, and public fora, right? As soon as a worthwhile attribute XForm extension is proposed and agreed for the benefit of all, we should agree on a proper dedicated XLSForm column for it.

ukanga:
@yanokwa I think this is reasonable. Could have benefited from a test in the PR, will merge it once I have done testing.

PR #48, opened 2015-09-15
Include UTF-8 and translation tests in 'tests_v1' format

dorey:


dorey:
Merging own PR because it's been up for a week and only contains improvements to tests in the `tests_v1` directory

PR #55, opened 2015-12-16
Allows "value" as an alias for "name"

dorey:
* allows "value" as an alias for "name" (as in old behavior)
* moves existing tests away from xls fixtures


dorey:
@ukanga @yanokwa @pld @mberg 

This PR restores `value` as an accepted alias for the `name` column of the survey sheet as discussed in [odk-developers](https://groups.google.com/forum/#!topic/opendatakit-developers/BZKVfQkotjo) thread.

pyxform/tests_v1/test_sheet_columns.py
```diff
@@ -0,0 +1,134 @@
...
+                    ],
+                model__contains=[
+                    '<bind nodeset="/aliases/q1" type="string"/>',
+                    ],
+                xml__contains=[
+                    '<input ref="/aliases/q1">',
+                    '<label>Question 1</label>',
+                    '</input>',
+                    ])
+''' # uncomment when re-implemented
```
ukanga: Any specific reason why we should keep this commented code in?

ukanga:
:+1:

pyxform/tests_v1/test_sheet_columns.py
dorey: I was thinking of fixing it in this PR but i didn't get to it. I kept it in as a note about some incomplete behavior associated with aliases. I could attach it to a new issue to fix when-time.

ukanga: Attaching to a new issue would be the ideal way I think, otherwise this is good to merge.

PR #60, opened 2016-03-21
use built-in xml.etree rather than lxml

sheppard:
This makes installing on various platforms a bit easier since there's one less thing to compile.  The API for `xml.etree.ElementTree` covers most of what `lxml.etree` provides, with some minor differences.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -9,8 +9,9 @@ import re
 from odk_validate import check_xform
 from survey_element import SurveyElement
 from errors import PyXFormError
-import constants
+from pyxform import constants
 import os
+import xml.etree.ElementTree as ET
 
 
 nsmap = {
@@ -22,6 +23,9 @@ nsmap = {
     u"xmlns:orx": u"http://openrosa.org/xforms"
     }
 
+for prefix, uri in nsmap.items():
+    prefix = prefix.replace("xmlns", "").replace(":", "")
+    ET.register_namespace(prefix, uri)
 
 class Survey(Section):
 
--- a/pyxform/tests/builder_tests.py
+++ b/pyxform/tests/builder_tests.py
@@ -1,5 +1,5 @@
 import re
-from lxml import etree
+import xml.etree.ElementTree as ET
 from unittest import TestCase
 from pyxform.builder import SurveyElementBuilder, create_survey_from_xls
 from pyxform.xls2json import print_pyobj_to_json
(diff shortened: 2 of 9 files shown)


sheppard:
Any thoughts on this?  It takes a full minute off of the build time.

pld:
does this affect performance when running pyxform? if not, 👍

sheppard:
Good question.  I rebased and used `timeit` to run the test suite with and without the change and in both Python 2.7 and 3.5.  It looks like performance is slightly better in 3.5 and slightly worse in 2.7 - though this may be within the margin of error.  My guess is that 2.7 could be made a bit faster by using `cElementTree` instead of `ElementTree` (which is effectively what 3.5 is doing).

As [this page discusses](http://lxml.de/performance.html), `lxml.etree` is still often faster than `ElementTree` (and `cElementTree`), but not in every case, and the performance difference is not nearly as large as it was historically (i.e. in Python 2.6 and earlier).

```bash
git checkout master
python -m timeit -n 10 -r 3 -s 'import os' 'os.system("python2 setup.py nosetests")'
# (test output...)
10 loops, best of 3: 11.9 sec per loop

git checkout xml
python -m timeit -n 10 -r 3 -s 'import os' 'os.system("python2 setup.py nosetests")'
# (test output...)
10 loops, best of 3: 12.2 sec per loop

git checkout master
python -m timeit -n 10 -r 3 -s 'import os' 'os.system("python3 setup.py nosetests")'
# (test output...)
10 loops, best of 3: 11.9 sec per loop

git checkout xml
python -m timeit -n 10 -r 3 -s 'import os' 'os.system("python3 setup.py nosetests")'
# (test output...)
10 loops, best of 3: 11.7 sec per loop

```

pld:
this is acceptable to me, @dorey @ukanga leaving this to y'all as the pyxform pyexperts

dorey:
The tests run with `nosetests` run so slowly because it's spawns a java process to run each xform through ODK validate.

Not that it adds to the discussion a whole lot, but I confirmed that there's no slowdown when running the lightweight tests (in the `tests_v1` directory).

```
python -m timeit -n10 -r3 -s 'import os' 'os.system("python2.7 setup.py nosetests --tests pyxform/tests_v1/")'
# test output
10 loops, best of 3: 805 msec per loop
git checkout xml
python -m timeit -n10 -r3 -s 'import os' 'os.system("python2.7 setup.py nosetests --tests pyxform/tests_v1/")'
# test output
10 loops, best of 3: 811 msec per loop
```

👍  for moving away from lxml!

PR #63, opened 2016-04-29
"--skip_validate" arg for xls2xform.py

lindsay-stevens:
c758940 Exposes the existing option in survey.print_xform_to_file to not run ODK_Validate on the output XML file, as an optional argument "--skip_validate" on the command line.
- Allows packaging the xls2xform.py script as onefile pyinstaller exe for easy distribution. (I can provide this if there is interest, it's just ```pyinstaller xls2xform.py --onefile".
- ODK_Validate can still be run as a separate step; sorting out the java side of things is easier for less technical users that way.

94c759c Before I could get started on this there were a couple of Windows-specific problems preventing the nosetests from running. In each case I've tried to make it so that the original behaviour will be attempted, then try the Windows route, and then fail after that. All the tests pass for me on Windows 7 x64 with these changes. It's passing in travis-ci so I guess it's Linux compatible too. https://travis-ci.org/lindsay-stevens/pyxform/builds/126578491.

a57ab51 I also had trouble with the bundled ODK_Validate.jar; when running the tests, the ```java -jar ``` call would error out with NoClassDefFound exceptions. I think it is something to do with the "jar inside a jar" structure. So I re-packaged it from the validate repo as of 2016-04-29 https://github.com/opendatakit/validate/commit/65444de and it worked.


dorey:
@lindsay-stevens, thank you for getting this working in windows.

PR #67, opened 2016-05-13
We should not add default hints into forms

yanokwa:
This is behavior is not great for three reasons.

1. It's not transparent. When a user converts the form, it's not clear where the hint comes from. It took way too long to find which tool in the ecosystem was adding this (e.g., Javarosa, Collect, xlsform.opendatakit.org, pyxform)
2. It's not easy to localize. We should make things easy to localize and putting this in the source of a library makes that hard.
3. It's not that accurate. Geopoint isn't just for GPS. It has Wi-Fi and cell localization that works indoors. Also outdoors isn't sufficient to get a lock. You need to have a clear view of the sky.

### DIFF
--- a/pyxform/question_type_dictionary.py
+++ b/pyxform/question_type_dictionary.py
@@ -381,8 +381,7 @@ QUESTION_TYPE_DICT = \
             },
             "bind": {
                 "type": "geopoint"
-            },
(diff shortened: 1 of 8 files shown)


yanokwa:
What do you think, @dorey?

dorey:
👍 Looks good

Issue #68, opened 2016-05-31
Non-ascii characters in a question name cause an exception


spatialdude:
When non-ascii characters are in the question name, the conversion fails with an exception such as -

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe1' in position 1: ordinal not in range(128)
```

instead of reporting that the question name is invalid.

spatialdude:
Solution https://github.com/XLSForm/pyxform/issues/68#ref-commit-fbe6b8e is to encode **question_name**

```python
 error_message += " Invalid question name [" + question_name.encode('utf-8') + "] "
```

lognaturel:
I think this can be closed? (Sorry, I commented on the PR before!)

PR #71, opened 2016-06-14
#68 encode question_name

dorey:
@spatialdude 's fix for #68


lognaturel:
I think this can be closed?

PR #75, opened 2016-08-05
add additional namespaces to XForm

### LINKED ISSUE #65: Add ability to add additional namespaces
Referring to [Esri's additions](http://doc.arcgis.com/en/survey123/create-surveys/esricustomcolumns.htm) and to custom features I am working on, it would be great if we could add additional space-separated `prefix="namespace"` e.g. in the settings sheet under a namespaces column that result in the corresponding xmlns attributes on the html element.

| namespaces |
--------------------
|esri="http://esri.com/xforms" enk="http://enketo.org/xforms" naf="http://nafundi.com/xforms"|

This would allow us to use survey columns such as: `bind::esri:fieldType` and `bind::nk2:myattribute` and not be tempted to either use no namespace or an incorrect (but already existing) one.

So the above would ensure the following namespace declarations are in the XForm output:

```xml
<h:html .... 
    xmlns:esri="http://esri.com/xforms" 
    xmlns:enk="http://enketo.org/xforms" 
    xmlns:naf="http://nafundi.com/xforms"
>
```

ukanga:
closes #65

### DIFF
--- a/pyxform/constants.py
+++ b/pyxform/constants.py
@@ -61,3 +61,5 @@ CASCADING_CHOICES = u"cascades"
 
 OSM = u"osm"
 OSM_TYPE = u"binary"
+
+NAMESPACES = u'namespaces'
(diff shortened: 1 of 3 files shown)


MartijnR:
Thanks! 👍

PR #76, opened 2016-08-07
add support for external_instances

### LINKED ISSUE #30: Add support for external instances
[earlier discussion on ODK dev forum](https://groups.google.com/forum/#!topic/opendatakit-developers/9_VZoe7crVY)

* Support for both *select_one_from_file \[FILENAME\]* and *select_multiple_from_file \[FILENAME\]* 
* The above types always create regular `<select>` or `<select1>` with a regular `<itemset>` and always add a single `<instance id="" src="" />` to the model (unless that instance with that id already exists).
* Filenames can have either .csv or .xml extensions.
* instance `id` attribute = lowercased filename minus the extension: countries.csv => `<instance id="countries">`
* instance `src` attribute for XML sources gets `jr://file/` prefix and for CSV sources gets `jr://file-csv/` prefix (similar to jr://image, jr://video and jr://audio).
* the label ref of an itemset is always `ref="label"` because itext references couldn't work.

#### 1a. XLSForm input (CSV file, no choice filter):

| type  | name   | label | choice_filter
|-------|--------|-------|-------------------
| select_one_from_file countries.csv | country | Pick one country |

#### 1b. XForm output:

Very similar to a regular select_one with choice filter (but in this case the choice filter is empty)

```xml
<model>
    <instance>
        ....
    </instance>
    <instance id="cities" src="jr://file-csv/cities.csv" />
</model>
<h:body>
    <select1 ref="/myform/country">
        <label>Country</label>
        <itemset nodeset="instance('countries')/root/item">
            <value ref="name"/>
            <label ref="label"/>
        </itemset>
    </select1>
</h:body>

```

#### 2a. XLSForm input (XML file, no choice filter):

| type  | name   | label | choice_filter
|-------|--------|-------|-------------------
| select_one_from_file countries.xml | country | Pick one country | 

#### 2b. XForm output:

_Exactly the same syntax as used by CommCare for their many XML external instances_

```xml
<model>
    <instance>
        ....
    </instance>
    <instance id="cities" src="jr://file/cities.xml" />
</model>
<h:body>
    <select1 ref="/myform/country">
        <label>Country</label>
        <itemset nodeset="instance('countries')/root/item">
            <value ref="name"/>
            <label ref="label"/>
        </itemset>
    </select1>
</h:body>

```

#### 3a. XLSForm input (CSV file, with choice filter):

| type  | name   | label | choice_filter
|-------|--------|-------|-------------------
| select_multiple_from_file cities.csv | city | Pick some cities | country = ${country}

#### 3b. XForm output:

```xml
<model>
    <instance>
        ....
    </instance>
    <instance id="cities" src="jr://file-csv/cities.csv" />
</model>
<h:body>
    <select ref="/myform/country">
        <label>Country</label>
        <itemset nodeset="instance('cities')/root/item[country= /myform/country ]">
            <value ref="name"/>
            <label ref="label"/>
        </itemset>
    </select>
</h:body>
```

Notes:

* this feature is not compatible with ODK Collect (and vice versa select_external and search are not compatible with Enketo)
* the external data referred to in an XLSForm needs to always have a label column (that's an XLSForm restriction) 
* csv column headers may not contain a colon (this not relevant to Pyxform but to the xlsform.org documentation)
* these XForms do not pass ODK Validate validation (I suspect it checks whether itemset nodeset paths exist), so to use this feature ODK Validate would have to be switched off or tweaked so it bypasses checks for paths that refer to external instances (i.e. that refer to instance(ID)/path/to/node when the targeted instance has a `src` attribute)

ukanga:
closes #30

### DIFF
--- a/pyxform/tests_v1/test_support_external_instances.py
+++ b/pyxform/tests_v1/test_support_external_instances.py
@@ -1,0 +1,105 @@
+from pyxform_test_case import PyxformTestCase
+
+
+class ExternalCSVInstancesTest(PyxformTestCase):
+    def test_external_csv_instances(self):
+        # re: https://github.com/XLSForm/pyxform/issues/30
+        self.assertPyxformXform(
+            name="ecsv",
+            md="""
+            | survey |                                              |                |                |
+            |        | type                                         | name           | label          |
+            |        | select_one_from_file cities.csv              | city           | City           |
+            |        | select_multiple_from_file neighbourhoods.csv | neighbourhoods | Neighbourhoods |
+            """,  # noqa
+            xml__contains=[
+                """<instance id="cities" src="jr://file-csv/cities.csv">
+        <root>
+          <item>
+            <name/>
+            <label/>
+          </item>
+        </root>
+      </instance>""",  # noqa
+                '<select1 ref="/ecsv/city">',
+                """<instance id="neighbourhoods" src="jr://file-csv/neighbourhoods.csv">
+        <root>
+          <item>
+            <name/>
+            <label/>
+          </item>
+        </root>
+      </instance>""",  # noqa
+                '<select ref="/ecsv/neighbourhoods">',
+            ], debug=True
+        )
+
+
+class ExternalXMLInstancesTest(PyxformTestCase):
+    def test_external_xml_instances(self):
+        # re: https://github.com/XLSForm/pyxform/issues/30
+        self.assertPyxformXform(
+            name="exml",
+            md="""
+            | survey |                                              |                |                |
+            |        | type                                         | name           | label          |
+            |        | select_one_from_file cities.xml              | city           | City           |
+            |        | select_multiple_from_file neighbourhoods.xml | neighbourhoods | Neighbourhoods |
+            """,  # noqa
+            xml__contains=[
+                """<instance id="cities" src="jr://file-xml/cities.xml">
+        <root>
+          <item>
+            <name/>
+            <label/>
+          </item>
+        </root>
+      </instance>""",  # noqa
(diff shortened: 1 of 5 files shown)


pyxform/tests_v1/test_support_external_instances.py
```diff
@@ -0,0 +1,138 @@
...
+            xml__contains=[
+                """<instance id="cities" src="jr://file-csv/cities.csv">
+        <root>
+          <item>
+            <name/>
+            <label/>
+          </item>
+        </root>
+      </instance>""",  # noqa
+                '<select1 ref="/ecsv/city">',
```
MartijnR: maybe add, to ensure an itemset is output without a choice-filter/predicate: `<itemset nodeset="instance(\'cities\')/root/item">'

pyxform/tests_v1/test_support_external_instances.py
```diff
@@ -0,0 +1,138 @@
...
+            xml__contains=[
+                """<instance id="cities" src="jr://file-xml/cities.xml">
+        <root>
+          <item>
+            <name/>
+            <label/>
+          </item>
+        </root>
+      </instance>""",  # noqa
+                '<select1 ref="/exml/city">',
```
MartijnR: maybe add, to ensure an itemset is output without a choice-filter/predicate: `<itemset nodeset="instance(\'cities\')/root/item">'

pyxform/tests_v1/test_support_external_instances.py
ukanga: will do.

MartijnR:
thanks! 👍

PR #84, opened 2016-09-16
Py3 compatibility and misc. fixes

lindsay-stevens:
Summary of changes:
- Baked-in Python 3 compatibility (e.g. not via 2to3), mostly:
    - print function
    - absolute imports
    - dict iterables to list
- Fix complaint / warning on py3 with Windows java detection comparing bytes to str
- Windows chokes on os.path.exists if the str is > 255 chars, so refactor to try parsing XML first instead of trying to see if XML is a valid file
- PEP8 all the things
- Remove unused / duplicate variables, functions, modules
- Patch minidom's XML serialization of text characters to not escape double quotes, only [&<>], for compatibility with libxml and expat behaviour (e.g. if modifying output from pyxform).

I started out just doing the last one (escaping), but it kind of snowballed into all the other stuff. Hope it looks alright. The trade-offs for adopting this Py3 support:
- Use absolute imports, cast dict item/key views to lists, use \__future\__.print_function;
- If using unicode, basestring or unichr, import them from the pyxforms.utils module instead of builtins.


ukanga:
Awesome, looks good. 

Should this pass PEP8?

lindsay-stevens:
I applied quite a few PEP8 changes but there's still some naming (a bit of camelCase functions/variables here and there) and line length violations (largely in test fixture strings). 

If you're open to / would prefer strict PEP8 let me know and I'll update the PR.

ukanga:
Apologies, went away for a bit.

I had assumed when you stated PEP8 that you had resolved all instances. I am not keen on PEP8 but that would be a very welcome PR. If you can fix the merge conflict, I can review this PR again before merging.

lindsay-stevens:
All done I think. Not sure what happened with the changes going into two commits there though - I'm blaming jetbrains for that :)

PR #104, opened 2017-03-28
Fix #87 validation of external instances

### LINKED ISSUE #87: Validation error with external instances
Hi,

I'm getting a validation error in `xls2xform` when using `select_one_from_file` in my XLSForm. If I skip validation it loads through ODK Aggregate and works with Enketo.

XLSForm content is attached (see [select_from_file_test.xlsx](https://github.com/XLSForm/pyxform/files/488142/select_from_file_test.xlsx)) and shown below.

survey:

| type | name | label | choice_filter |
| --- | --- | --- | --- |
| select_one_from_file states.csv | State | Select State |  |

choices:

| list_name | name | label |
| --- | --- | --- |
| states | name | label |

settings:

| form_title | form_id |
| --- | --- |
| select_from_file test | select_from_file_test |

The validation error is shown below.

```
python xls2xform.py select_from_file_test.xlsx select_from_file_test.xml
  File "xls2xform.py", line 94, in <module>
    main_cli()
  File "xls2xform.py", line 86, in main_cli
    args.path_to_XLSForm, args.output_path, args.skip_validate)
  File "xls2xform.py", line 23, in xls2xform_convert
    survey.print_xform_to_file(xform_path, validate=validate, warnings=warnings)
  File "/src/pyxform/pyxform/survey.py", line 507, in print_xform_to_file
    warnings.extend(check_xform(path))
  File "/src/pyxform/pyxform/odk_validate/__init__.py", line 133, in check_xform
    'ODK Validate Errors:\n' + _cleanup_errors(stderr))
odk_validate.ODKValidateError: ODK Validate Errors:
Selection choice label text and image uri are both missing for: ${State} choice: 1.

Selection value is missing for: ${State} choice: 1. The XML is invalid.

>> Xform is invalid! See above for errors and warnings.

Result: Invalid
```

Below is the XML generated if I run with `--skip_validate`. This loads without error in ODK Aggregate and the form works as expected in Enketo. So should the validation pass?

``` xml
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <h:head>
    <h:title>select from file test</h:title>
    <model>
      <instance>
        <select_from_file_test id="select_from_file_test">
          <State/>
          <meta>
            <instanceID/>
          </meta>
        </select_from_file_test>
      </instance>
      <instance id="states" src="jr://file-csv/states.csv">
        <root>
          <item>
            <name/>
            <label/>
          </item>
        </root>
      </instance>
      <bind nodeset="/select_from_file_test/State" type="select1"/>
      <bind calculate="concat('uuid:', uuid())" nodeset="/select_from_file_test/meta/instanceID" readonly="true()" type="string"/>
    </model>
  </h:head>
  <h:body>
    <select1 ref="/select_from_file_test/State">
      <label>Select State</label>
      <itemset nodeset="instance('states')/root/item">
        <value ref="name"/>
        <label ref="label"/>
      </itemset>
    </select1>
  </h:body>
</h:html>
```

rhunwicks:


ukanga:
Looks okay, any chance you can add a test for this?

rhunwicks:
I have added `run_odk_validate=True` to the existing tests, which causes them to fail against master. Then I have updated them to match the slightly altered XML output (with the included `_`)

MartijnR:
Thanks @rhunwicks. This is great. 

External instances have been accepted by the ODK community recently, so in a future ODK Validate update, we should be able to remove the trick (the complete dummy instance content) to fool it into validating.

rhunwicks:
@MartijnR please can you point me to the discussion about accepting external instances - I want to use `select_one_from_file` with Kobocat and there are some changes to Kobocat that would be required?

dorey:
@rhunwicks, in one of the kobocat py files there's a method imported called `from pyxform import survey_from`. If you comment out this import , then you can switch kobocat's pyxform to this one, and only lose spss features. We are working on a fix to kobocat's pyxform situation that should be pushed in the next week, at which point `select_one_from_file` will be supported.

MartijnR:
Excellent news @dorey (probably also activates pulldata())

@rhunwicks the discussion I was referring to in order to in the future clean up the pyxform code by removing th dummy instance content was an [XForms Spec discussion](https://github.com/opendatakit/xforms-spec/issues/50).

PR #116, opened 2017-05-31
Add ODK Validate 1.5.0 jar.

lognaturel:
Closes #114. Confirmed working on macOS, Linux and Windows 10.

<strike>I have verified that I can run this jar from where it is with `java -jar ODK_Validate.jar`. The UI shows the version as 1.5.0 and a form such as [contains_starts-with_ends-with.txt](https://github.com/XLSForm/pyxform/files/1042485/contains_starts-with_ends-with.txt) that uses new XPath functions validates. I can also run Validate through the CLI and get the same results.

However, if I run `python ./xls2xform.py` on the corresponding XLSForm -
[contains_starts-with_ends-with.xlsx](https://github.com/XLSForm/pyxform/files/1042489/contains_starts-with_ends-with.xlsx) - I get the same ODK Validate errors as I got with the old jar. Interestingly, if I then run this same jar on the resulting xml file, it validates as expected.

When I run `nosetests`, I get the same behavior as on master. 117 passes but there is `usage: nosetests [-h] [--json] [--skip_validate] path_to_XLSForm output_path nosetests: error: too few arguments` printed part way through.

Is there something special that needs to be done to update the jar? Paging @lindsay-stevens because I see you were the last to do this update and had to fix some previous issues.</strike>

Issue #99, opened 2017-01-03
Add setup step to readme.md?


MartijnR:
For python noobs such as myself, that would like to use this, this should maybe go somewhere in the readme.md:

> `python setup.py install`
> `python setup.py develop`

alxndrsn:
Another python ignorant here.  What do these commands do?

ukanga:
@MartijnR the readme has undergone some updates recently, can you please check and see if it warrants closing this issue.

ukanga:
@alxndrsn https://stackoverflow.com/questions/19048732/python-setup-py-develop-vs-install does have some answers for you. Development mode or production mode.

MartijnR:
Looks good. Thanks! Can be closed I think.

PR #119, opened 2017-06-01
Update local execution instructions.

lognaturel:
Mostly from @dorey [here](https://github.com/XLSForm/pyxform/pull/116#issuecomment-305521611). The key thing is to replace the explicit install of `xlrd` with a run of `setup.py`. Additionally, I think including the `virtualenv` strategy explicitly is very helpful. Suggestions appreciated!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,24 +20,38 @@ https://github.com/modilabs/pyxform
 
 pyxform is a major rewrite of `xls2xform <http://github.com/mvpdev/xls2xform/>`_.
 
-Running pyxform as a Python script:
-===========================
+Running pyxform from local source
+=================================
 
-1. install xlrd::
+Note that you must uninstall any globally installed `pyxform` instance in order to use local modules.
 
-    # On ubuntu these terminal commands should do it:
+From the command line::
 
-    easy_install pip
-
-    pip install xlrd
-
-2. Run this command::
-
+    python setup.py develop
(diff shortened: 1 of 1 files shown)


alxndrsn:
A couple of comments on this, although I have no experience with Python:

1. these new instructions look helpful for people who are planning to make changes to pyxform, but for general use it seems much cleaner just to directly `pip install` (_without_ the `-e`), and not worry about virtualenv.  If there are reasons why that's a bad idea I'd be really interested to learn more.  Otherwise, I think it would be good to move the simple, general-user setup instructions to the top of the doc.
2. on https://github.com/XLSForm/pyxform/issues/99#issuecomment-305733981, @ukanga linked https://stackoverflow.com/questions/19048732/python-setup-py-develop-vs-install, which suggests that `pip install` should be preferred over invoking `setup.py` directly

lognaturel:
Hmm, good points, @alxndrsn. Typically I would say non-developers really shouldn't be coming to GitHub to get instructions but it seems like currently there isn't really anywhere else to get them from, probably because most end-users use pyxform through various wrappers. That said, there is category of people (me prior to yesterday) who want the command-line tool but don't need to modify it. So I agree with putting those instructions at the top.

alxndrsn:
> That said, there is category of people (me prior to yesterday) who want the command-line tool but don't need to modify it.

This is the category I'd choose to be in too 😄

PR #120, opened 2017-06-02
Link to specific repo in install instructions

alxndrsn:


alxndrsn:
N.B. may overlap with __ and __.  If this change is accepted, let me know and I'll fix the mess before merge.

lognaturel:
I understood this differently. I thought the point was that if @alxndrsn has a branch I'm interested in trying out, I could install it that way and make comments on the corresponding pull request. Sometimes I may also want to pull from master but I think keeping it generic is more helpful.

lognaturel:
I would prefer something like `git+https://github.com/<GitHub username>/pyxform.git@<branch name>#egg=pyxform` but I suppose that if someone wants to pull code this way, they probably get it. So I don't feel strongly about this.

PR #121, opened 2017-06-02
Don't pip-install in developer mode

alxndrsn:
From `pip help install`:

> -e, --editable <path/url>  Install a project in editable mode (i.e. setuptools
>                            "develop mode") from a local project path or a VCS
>                            url.

As the instructions above this section are for installing for developers, I
don't think it's helpful for general users to be encouraged here to install in
dev mode.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,7 @@ Installing pyxform from remote source
 =====================================
 `pip` can install from any GitHub repository::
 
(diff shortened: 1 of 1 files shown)


alxndrsn:
N.B. may overlap with __ and __.  If this change is accepted, let me know and I'll fix the mess before merge.

PR #122, opened 2017-06-02
Move user-focussed installation instructions higher in README

alxndrsn:


alxndrsn:
N.B. may overlap with #120 and #121.  If this change is accepted, let me know and I'll fix the mess before merge.

ukanga:
I think you may need to rebase and have one PR that also includes #121, seem related.

alxndrsn:
👍 as per comment above!

README.rst
```diff
@@ -17,6 +17,16 @@ https://github.com/modilabs/pyxform
 
 pyxform is a major rewrite of `xls2xform <http://github.com/mvpdev/xls2xform/>`_.
 
+Installing pyxform from remote source
+=====================================
+`pip` can install from any GitHub repository::
+
+	pip install -e git+https://github.com/XLSForm/pyxform.git@master#egg=pyxform
```
ukanga: Thought the idea was to change this to be without the `-e` option.

alxndrsn: No, that's at https://github.com/XLSForm/pyxform/pull/121!

lognaturel:
I would do this differently. I would have a section called something like "Installing the latest release of pyxform for command-line use" that just uses `pip install pyxform` and then shows how to use `xls2xform`. Ideally above that there'd also be a little narrative about how `pyxform` is within several tools with GUIs.

Then there'd be another section for pyxform developers with two suggestions -- run from local source or install in dev mode from an arbitrary git repo.

alxndrsn:
@lognaturel sounds good to me.  I don't know this tool well enough to do more than incremental improvements to documentation, so I'll leave this for someone else.

PR #127, opened 2017-06-08
Add user facing install instructions

lognaturel:
I tried to make it clear that most users will want to use GUI-based tools but that those who want to use `pyxform` but not develop for it can install through `pip`.


ukanga:
Looks good to me.

Issue #79, opened 2016-08-12
Support for numeric range questions


spatialdude:
Add support for numeric range questions.

The XForms specification describes the `<range>` control https://en.wikibooks.org/wiki/XForms/Range which might look like 

![image](https://cloud.githubusercontent.com/assets/7235583/17612613/553dee12-6097-11e6-8bd1-c56b5c37d315.png)

This would be provide more flexibility than the **distress** appearance as supported in [Enketo](http://blog.enketo.org/distress/) which has a limited range of 0 to 10.

![image](https://cloud.githubusercontent.com/assets/7235583/17612634/a46d27a0-6097-11e6-9e21-3a148b806f3b.png)

The `<range>` control is also mentioned in the JavaRosa roadmap https://bitbucket.org/javarosa/javarosa/wiki/xform

Two approaches could be considered -
### Option 1

A **range** question type, which would be supported by additional columns to define the parameters for start, end, step and increment and appearance type to control presentation (eg **compact**  for spinner control or **horizontal** for a horizontal bar)
### Option 2

Support a **range** appearance type for **integer** and **decimal** question types. 

e.g **`range(start, end, step, increment)`** with logical default for all parameters (e.g. 1, 10, 1, 1)

Option 2 has some simplicity advantages in that does not require additional columns to be recognised nor the addition of new control types. It could also be combined with other appearance for additional styling control. As its just an appearance type, there would no change to pyxform.

e.g. `range(1, 200) compact`

Which might look like - 

![image](https://cloud.githubusercontent.com/assets/7235583/17613437/881904b2-60a0-11e6-9852-64f900b94d8b.png)

An enhanced **range(...**) appearance bears some similarity in concept to the **select(...)** expression appearance type http://xlsform.org/#dynamic-selects-from-pre-loaded-data

MartijnR:
I would be in favor of adding support for the `<range>` control in our XForms implementation, according to the [XForms spec](https://www.w3.org/TR/xforms/#ui-range). Perhaps just omitting the `incremental` property.

I think our first concern is adding it to JavaRosa because of ODK Validate. It may be better to move this discussion (as an XForms discussion) to the ODK Dev forum first, but we can see what happens here. 

I'm guessing there are already some JavaRosa ports that have added this functionality but they might have been inflexible, specific solutions like Enketo's distress widget.

(On the XLSForm side, I like option 2 more but option 1 also seems fine with me.)

MartijnR:
[...]

PR #131, opened 2017-06-19
Add support for numeric range questions

ukanga:
See details in #79 

- [x] introduce a new column called parameters, attributes or properties
- [x] use name/value pairs separated by `=` in those cells
- [x] separate name/value pairs by spaces or comma or semicolon.
- [x] for `range`, allow `min`, `max` and `step` with defaults `1`, `10` and `1` respectively.

### DIFF
--- a/pyxform/aliases.py
+++ b/pyxform/aliases.py
@@ -91,9 +91,7 @@ survey_header = {
     u"required_message": u"bind::jr:requiredMsg",
     u"required message": u"bind::jr:requiredMsg",
     u"body": u"control",
-    u"properties": u"properties",
-    u"parameters": u"properties",
-    u"attributes": u"properties",
+    u"parameters": u"parameters",
 }
 list_header = {
     u"caption": constants.LABEL,
--- a/pyxform/aliases.py
+++ b/pyxform/aliases.py
@@ -90,7 +90,8 @@ survey_header = {
     u"requiredMsg": u"bind::jr:requiredMsg",
     u"required_message": u"bind::jr:requiredMsg",
     u"required message": u"bind::jr:requiredMsg",
-    u"body": u"control"
+    u"body": u"control",
+    u"properties": u"properties",
 }
 list_header = {
     u"caption": constants.LABEL,
--- a/pyxform/aliases.py
+++ b/pyxform/aliases.py
@@ -92,6 +92,8 @@ survey_header = {
     u"required message": u"bind::jr:requiredMsg",
     u"body": u"control",
     u"properties": u"properties",
+    u"parameters": u"properties",
+    u"attributes": u"properties",
 }
 list_header = {
     u"caption": constants.LABEL,
--- a/pyxform/builder.py
+++ b/pyxform/builder.py
@@ -1,16 +1,16 @@
 import copy
-from pyxform import file_utils
 import os
-from pyxform import utils
 
+from pyxform import file_utils, utils
 from pyxform.errors import PyXFormError
-from pyxform.section import RepeatingSection, GroupedSection
-from pyxform.survey import Survey
-from pyxform.question import Question, InputQuestion, TriggerQuestion, \
-    UploadQuestion, MultipleChoiceQuestion, OsmUploadQuestion
+from pyxform.question import (InputQuestion, MultipleChoiceQuestion,
+                              OsmUploadQuestion, Question, RangeQuestion,
+                              TriggerQuestion, UploadQuestion)
 from pyxform.question_type_dictionary import QUESTION_TYPE_DICT
-from pyxform.xls2json import SurveyReader
+from pyxform.section import GroupedSection, RepeatingSection
+from pyxform.survey import Survey
 from pyxform.utils import unicode
+from pyxform.xls2json import SurveyReader
 
 
 def copy_json_dict(json_dict):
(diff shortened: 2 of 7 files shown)


lognaturel:
Awesome, @ukanga! This is great!

I hadn't noticed that in #79 we switched from `start` and `end` which are the XForms names to `min` and `max`. I would tend to keep `start` and `end` for consistency unless there's a good reason not to.

One thing I just learned is that `body::attribute` is an existing mechanism that could be used to get the attributes in. Is the idea that it would be deprecated in favor of this new column?

I'm not sure whether this is because I write code, because I'm a little obsessive or what but I really want to put spaces around equal signs (e.g. `step = 3`)!! This might not be important but I did want to mention it. Maybe the split could be on an equal sign surrounded by space? The current error is `ValueError: need more than 1 value to unpack`. Making that a little more helpful is another idea.

I'm not sure I like that the new column can have three different names. I think that's likely to be more confusing than anything if people are sharing spreadsheets and seeing different names for the same thing. Including more than one of the same column (both `properties` and `attributes` for example) results in `AttributeError: 'dict' object has no attribute 'split'`.

@spatialdude @MartijnR @yanokwa Perhaps you have other thoughts?

MartijnR:
re `body::attribute` (and `bind::attribute`), I think we should probably maintain them. I see them as lower level features for developers (and custom apps). 

The new column will presumably only pick the attributes it's looking for (unlike `body::attribute` and `bind::attribute`), and could magically place them on either a bind or form control depending on the feature, and magically change the attribute name from user-friendly to spec-friendly.

ukanga:
I used https://github.com/XLSForm/pyxform/issues/79#issuecomment-308898575 as a point of reference, I will make the change to `start` and `end`.

I will look into the spacing `step = 3` issue as well as the other separator options, comma or semicolon. Might be easier just to deal with `body::attribute` and probably add validation for this instead. I lean more towards `body::attribute` than the extra columns, and if I were to pick one column to add from `parameters`, `properties` or `attributes` I would probably choose `properties`. A consensus on this will be helpful.

ukanga:
About the `AttributeError` exception, is there a test case I am not covering?

lognaturel:
I do think @MartijnR makes a good point that adding a new column makes it easier to use different attribute names between XForms and XLSForm which could help make XLSForm more user-friendly. I also think it's nice to limit the number of columns a form would have. On the other hand, using the more generic `body::attribute` strategy means pyxform doesn't have to change as much if new constructs with attributes are added. I think whether to go with a new column or the existing mechanism should be decided by folks who know XLSForm and how it is used better than I do (I really just learned about `body::attribute`). @yanokwa I'm sure you have opinions.

@ukanga regarding the `AttributeError`, the case is to include both a `properties` and an `attributes` column. But it sounds like it's unlikely both will continue to be accepted.

yanokwa:
Great discussion so far! 

Agreed on the rationale to add a new column. I think @MartijnR's framing of `body::attribute` as a developer-level feature makes sense to me.

Given that we have this direct mapping to XML attributes, I do not think it's a good idea to also call the new column attributes because that overloads the term. Ditto with properties. I like parameter as a term. It's my least favorite of the options, but I think it's the correct one.

I think we should have one option for the separator and it should be `=` (e.g., `param1=100 param2="team odk")`. 

I do not think we should allow spaces around the (e.g., `param1 = 100 param2 = 400` because it looks funny without a separator.

ukanga:
Thanks, I will implement using the `parameters` column.

ukanga:
This is now ready for another look, I have `parameters` as the only additional column. 

There is support for comma, space and semicolon separators, all the below are supported:
- `start=1 end=10 step=1`
- `start=1;end=10;step=1`
- `start = 1 , end = 10 , step = 1`

`ValueError` exceptions with `unpack` in the messages should no longer be present.

lognaturel:
Looking good! 😊

The `parameters` column seems to currently be case-sensitive. I think `Parameters` should also be accepted.

On the other hand, I'm not sure I like the three different separator options. Like we discussed when considering different names for the new column, I feel like it's confusing to have different people's XLSForm documents using different conventions. I don't really mind which one is picked, I would just prefer a single one. I suppose I'd also be ok with documenting and encouraging one but silently accepting others for a little more flexibility.

Other than that, it looks ready to go to me.

lognaturel:
Oh, and I'll add that the error messages are a lot friendlier, thank you! 🙏

ukanga:
All column headers tend to be case sensitive in XLSForm, could be a separate issue to separate it.
If it's to be documented we should stick with space separator to be similar to how the `appearance` column behaves. I am going to leave the other options in place.

lognaturel:
Thanks, @ukanga! 🎉

JohnHasthorpe:
Hi @ukanga @lognaturel 

We are looking at using the new **body::Parameters** column for a barcode question type, in order to pass in the formats that we want to scan against (e.g. Param1 = "QR, EAN, UPC, Code_39, Code_93, DataMatrix"). From a validation perspective would this work. From the above it looks like it will be validating separators only, but is this likely to change?

lognaturel:
@JohnHasthorpe I'm afraid I don't really understand your question. It might be helpful to start by proposing an XForms approach and working forward from there to see whether the parameter column is the best choice for what you are trying to achieve.

JohnHasthorpe:
@lognaturel Thanks for the reply. 

I was really trying to find out what kind of validation would be applied to body::Parameters column, and to work out if we could safely use it for other non XForms parameters in our app. However we have subsequently decided to use our own body::esri::parameters column for such parameters.

Thanks again for looking at this.

PR #132, opened 2017-06-23
Fixed #96: Whitespace after ${} should not be trimmed

### LINKED ISSUE #96: Whitespace after ${} should not be trimmed
In the attached form, I get input and use it in a label as output. The XML that pyxform  0.9.24 outputs trims the whitespace after the input, and that breaks my form.

That is, when I write `${in} test`, pyxform reads that as `${in}test` and so my XML looks like this:

Actual: `<label><output value=" /whitespace/in "/>test</label></input>`
Expected: `<label><output value=" /whitespace/in "/> test</label></input>`

[whitespace.xlsx](https://github.com/XLSForm/pyxform/files/627663/whitespace.xlsx)

batkinson:
This fixes corrupted labels when they contain variable references. The problem is due to pretty printing (and subsequent attempts at whitespace compacting) as described in #96.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -425,9 +425,9 @@ class Survey(Section):
         # TODO: check out pyxml
         # http://ronrothman.com/public/leftbraned/xml-dom-minidom-toprettyxml-and-silly-whitespace/
         xml_with_linebreaks = self.xml().toprettyxml(indent='  ')
-        text_re = re.compile('>\n\s+([^<>\s].*?)\n\s+</', re.DOTALL)
+        text_re = re.compile('(>)\n\s*(\s[^<>\s].*?)\n\s*(\s</)', re.DOTALL)
         output_re = re.compile('\n.*(<output.*>)\n(  )*')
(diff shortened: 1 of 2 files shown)


ukanga:
Cool. Any chance you can add a test for this in `pyxform/tests_v1/`? The below code should do.

```python
+            name='whitespace',
+            md="""
+            | survey  |                 |          |      |
+            |         | type            | label    | name |
+            |         | text            | A1       | a1   |
+            |         | note            | ${a1} A1 | n1   |
+            """,
+            xml__contains=[
+                '<label><output value=" /whitespace/a1 "/> A1 </label>',
+            ])
```

batkinson:
@ukanga thank you for pointing out where the tests were and the example test. I added the test in a separate commit for convienience. Those reviewing can reset to the prior commit and test without the fix to confirm the issue.

One additional note, the nature of this change affects the output format such that you may want to review this on other (larger) forms. While I believe the logic behind this behavior is sound (you can't safely collapse all whitespace without mistakenly running some items together), it may affect other form designs.

alxndrsn:
This change introduces a single space after `<value>` and before `</value>` in generated XML.

PR #134, opened 2017-06-27
Support `list_name` in OSM worksheet.

### LINKED ISSUE #109: 'list_name' in OSM Sheet
`'NoneType' object is not iterable` gets thrown for forms that use `list_name` instead of `list name` in the osm sheet

urbanslug:
Dealias and group headers for OSM worksheets.
In the OSM worksheet headers resolve all aliases and group them before generating the OSM tags.

closes #109


dorey:
In the [`xml__contains`](https://github.com/XLSForm/pyxform/pull/134/files#diff-1357f2cbf777840e9218ba8465217b60R20) strings in your tests, could you slim down the provided array of strings to test just the aspects of the XML document relevant to the OSM features?

An example is in the other tests, like the [`test_groups.py`](https://github.com/urbanslug/pyxform/blob/6f7168d652d780dfad72e878d4a1ef22550d1af8/pyxform/tests_v1/test_groups.py#L42) file, where individual XML changes are tested.

urbanslug:
@dorey I have reduced the right side of the test/equality to the minimum.

ukanga:
@urbanslug confirming the OSM tags are present is equally important as well.

Issue #105, opened 2017-03-31
Generate ODK XForms spec-compliant meta block


lognaturel:
BLOCKED:
- ~Enketo likely doesn't support this (see https://github.com/XLSForm/pyxform/issues/105#issuecomment-959796769)~
- Servers would need a way to identify/communicate that the meta block namespace has changed through form updates (https://github.com/XLSForm/pyxform/issues/105#issuecomment-578983800)

The [ODK Xforms spec](https://opendatakit.github.io/xforms-spec/#metadata) describes specific node names for metadata and requires that they be in the `meta` block and in the OpenRosa XForms namespace. The implementations tend to be more lenient but sticking to consistent naming has advantages for downstream analysis tools.

pyxform currently places metadata elements such as start time and end time in the `widgets` block and allows the user to specify an arbitrary node name. Instead, it would be preferable to generate something like 

```
<orx:meta>
    <orx:deviceID/>
    <orx:timeStart/>
    <orx:timeEnd/>
    <orx:userID/>
    <orx:instanceID/>
</orx:meta>
```

MartijnR:
Indeed! I've split this up into a few points and added a related instanceID syntax improvement:

*  not requiring a `name` value (and ignoring it if users do add it)
*  use the fixed nodeNames as specified in the spec for the individual meta nodes
*  adding the required namespace on both the meta block and the individual meta nodes
*  make sure `instanceID` is put under the same namespaced meta block
* don't add a `calculate` attribute on the instanceID bind, but  add a `preload="uid"` attribute instead

nribeka:
Checking to generate this meta block. Still poking around the code phase.

lognaturel:
Good news, @MartijnR's last bullet about the instanceID bind was done: https://github.com/XLSForm/pyxform/issues/94

craigappl:
Notes from the breakout session:

### Problem
We currently generate the meta tags. According to the standard specifications, we should put them in a different namespace instead of the way we use them now. When we generate them now, we generate them outside of the meta block. 

### Discussion
- The challenge is that people have downstream analysis dependent on the current configuration
- It allows the user to define the name of how this field is made available to the client. So, a user can define what they want to name these.
- The names come from the xform spec, which implies that they are static. Therefore, the same way you can always rely on the instanceID you can rely on these fields.
- We found that the xform spec defines it in the meta block.
- This is disruptive in the existing forms as well as downstream organizations who have analysis scripts. There are many users who use those fields.
[...]

PR #136, opened 2017-07-06
Add audit

### LINKED ISSUE #128: Add setting for audit
opendatakit/xforms-spec#94

The agreed-upon XForms syntax is:
```xml
<orx:meta>
    <orx:audit/>
</orx:meta>

<bind nodeset="/orx:meta/orx:audit" type="binary"/>
```

For now, including this would mean a csv file with detailed timing information is included with each form submission. In the future, there will also be the option to customize the contents of the csv (e.g. add location information) through attributes.

More about the feature and how it works is at opendatakit/collect#257

I presume this would be an `audit` column in the settings sheet. For now it could just contain yes/no true/false and eventually can have other values as the audit concept is expanded. Alternately the column could be named `audit_type` and currently only accept the value `timing`. 

@MartijnR I bet you have ideas for this.

lognaturel:
Closes #128. @MartijnR are you ok with this moving forward without #105 for now?

I tried to do this in the least invasive way I could think of. I moved the generation of the meta block to the top so that the audit field could be removed from `survey_sheet` but those changes to `xls2json.py` don't look great to me. @dorey @ukanga would be great to get your expert feedback!

### DIFF
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -513,6 +513,59 @@ def workbook_to_json(
         r"(?P<osm_command>(" + '|'.join(aliases.osm.keys()) +
         ')) (?P<list_name>\S+)')
 
+    # Create and populate meta block
+    meta_children = []
+
+    if aliases.yes_no.get(settings.get("omit_instanceID")):
+        if settings.get("public_key"):
+            raise PyXFormError(
+                "Cannot omit instanceID, it is required for encryption.")
+    else:
+        # Automatically add an instanceID element:
+        meta_children.append({
+            "name": "instanceID",
+            "bind": {
+                "readonly": "true()",
+                "calculate": settings.get(
+                    "instance_id", "concat('uuid:', uuid())"),
+            },
+            "type": "calculate",
+        })
+
+    if 'instance_name' in settings:
+        # Automatically add an instanceName element:
+        meta_children.append({
+            "name": "instanceName",
+            "bind": {
+                "calculate": settings['instance_name']
+            },
+            "type": "calculate",
+        })
+
+    # audit goes in meta block
+    if {u'type': u'audit', u'name': u'audit'} in survey_sheet:
+        meta_children.append({
+            "name": "audit",
+            "bind": {
+                "type": "binary"
+            },
+            "type": "audit"
+        })
+        survey_sheet.remove({u'type': u'audit', u'name': u'audit'})
+
+    if len(meta_children) > 0:
+        meta_element = \
+            {
+                "name": "meta",
+                "type": "group",
+                "control": {
+                    "bodyless": True
+                },
+                "children": meta_children
+            }
+        noop, survey_children_array = stack[0]
+        survey_children_array.append(meta_element)
+
     for row in survey_sheet:
         row_number += 1
         prev_control_type, parent_children_array = stack[-1]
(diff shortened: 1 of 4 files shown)


lognaturel:
Right, this also breaks a whole lot of the tests with fixed structures so there has to be another approach.

ukanga:
@lognaturel What has happened is you have changed the order in which the `meta/instanceID` is created in the `XForm`, it now appears first instead of last. To get the tests passing you will need to update all failing tests to match up or maintain the order by leaving the code after `for row in survey_sheet:` loop.

ukanga:
I lean towards maintaining the order since the alternatives may mean tests in https://github.com/onaio/onadata and https://github.com/kobotoolbox/kobocat will also need to be updated.

lognaturel:
The reason I moved it up is because I couldn't think of another good way to make sure `audit` wasn't both in `meta` and outside it. It seems it needs to first be identified as something that goes into `meta` and then removed from `survey_sheet`. If you like this approach, I'm happy to change the order in the test `xml` files. But if you all can think of a better way that doesn't require changing the order, do let me know. 

One thing I considered is creating an array of `meta` elements that would be removed from `survey_sheet` before the `survey_sheet` loop and then adding them to `meta_children`. That might be less invasive.

lognaturel:
I hadn't noticed your latest message. Yes, maintaining the order sounds better. Let me try what I described with introducing another list.

pyxform/xls2json.py
```diff
@@ -513,6 +513,59 @@ def workbook_to_json(
...
+        meta_children.append({
+            "name": "instanceName",
+            "bind": {
+                "calculate": settings['instance_name']
+            },
+            "type": "calculate",
+        })
+
+    # audit goes in meta block
+    if {u'type': u'audit', u'name': u'audit'} in survey_sheet:
```
ukanga: This actually should be handled inside the `for row in survey_sheet:` loop, you may actually not to add any code in this `xls2json.py` file. It should behave the same way `start` and `end` do, in such a manner that a user can have their own name for the `audit` type filed e.g `my_audit`.

lognaturel: Ahh, I forgot about the name, thanks. 

I'll look at `start` and `end`. The thing about them is that they're not currently put in the `meta` block so they're not a direct analog, I believe.

lognaturel: (they should be, that's what #105 is about. And actually, the name should eventually be ignored, I think.)

lognaturel:
@ukanga I'm not sure why I didn't go in this direction in the first place! This also gives a very clear path to eventually putting `timeStart`, `timeEnd`, etc in the meta block (part of #105). What do you think?

ukanga:
I suppose I missed the fact that this is a meta block field. This seems better.

I think we may still have an issue here, we are likely going to have a user name it any name they prefer and as such we will not have the `meta/audit` in the meta block. I see now why the settings sheet was being considered for this value.

Should we validate that the user entered `audit` for both `name` and `type` or simply default `audit` to `meta/audit`?

ukanga:
The test below does show how you could name the field differently and as a result probably have a different meaning than intended.

```python
self.assertPyxformXform(
    name="meta_audit",
    md="""
    | survey |        |          |       |                     |
    |        | type   |   name   | label | parameters          |
    |        | audit  |   aydit  |       |                     |
    """,
    xml__contains=[
        '<meta>',
        '<audit/>',
        '</meta>',
        '<bind nodeset="/meta_audit/meta/audit" type="binary"/>'],
)
```

MartijnR:
I think the output of that test is correct. Maybe pyxform should output a warning to the user that name is ignored if a name is entered? (or that could be part of #105, I've just added that the bullet points).

lognaturel:
I've forced the name to always be `audit` whether it is specified or not and added tests for those cases. 

I propose documenting `audit` in the XLSForm spec with a blank name column and making it clear that any name specified is ignored (I can do this). I don't see a big need to give the user a warning if the name is replaced although I suppose someone could want to try to force a particular name for downstream analysis. What do you think? I can add one if you'd like @MartijnR or @ukanga.

One other thing that just occurred to me -- should there be some kind of validation in case a user includes `audit` more than once?

MartijnR:
yes, that validation would be good if ODK Validate doesn't accidentally pick that up already. We could not do the warning until we learn that users are often confused about this despite documentation.

lognaturel:
Right, there's already a validation for two nodes of the same name and the error message is plenty descriptive, thanks @MartijnR! 

@ukanga @dorey This is ready for review. Don't hesitate to let me know if you don't think this is the right approach.

dorey:
Late to the topic, but I probably would have opted to enforce that the `name` is one of `[None, '', 'audit']` before setting it to `'audit'`, with an error message if they try to call it something else. Users would figure out pretty quick.

But this PR looks great. Nice, concise, and properly tested! 👍

lognaturel:
Thanks, @dorey! I added an error of the form "ERROR: '[row : 2] Audits must always be named 'audit.' The name column should be left blank.'"

dorey:
Awesome. Alright I think this is good to go!

PR #140, opened 2017-07-27
Refactor PyxformTestCase

ukanga:
This is to allow use of `PyxformTestCase.md_to_pyxform_survey(...)`  to create pyxform survey objects to use in tests in other projects without overriding the `assertContains` and `assertNotContains` functionality provided. Also, in cases where the test infrastructure is not required.

### DIFF
--- a/pyxform/tests_v1/pyxform_test_case.py
+++ b/pyxform/tests_v1/pyxform_test_case.py
@@ -1,26 +1,100 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
-from __future__ import print_function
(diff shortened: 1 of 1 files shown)


dorey:
I could see this being useful.

> Also, in cases where the test infrastructure is not required.

Is this something we will eventually want to rework into a `pyxform.utils` of its own --not in this PR, but in the future?

ukanga:
I will say yes to potentially moving it to `pyxform.utils`.

ukanga:
Thanks @dorey .

PR #147, opened 2017-08-17
[for review] Fix whitespace trimming of outputs

alxndrsn:


dorey:
The markdown representation is not very good when testing whitespace. In the future, I would recommend looking into copying one of the tests that uses a `ss_structure` representation.

alxndrsn:
Thanks, @dorey - tests updated.

alxndrsn:
@dorey I've fixed the problem described.  However, the fix also introduces the following change:

### Old
```
<label>
  <output value=" /xml_escaping/a "/>&lt; <output value=" /xml_escaping/b "/></label><item>
```
### New
```
<label><output value=" /xml_escaping/a "/> &lt; <output value=" /xml_escaping/b "/></label><item>
```

I think the new behaviour's an improvement, but would appreciate your opinion.

alxndrsn:
This change also gives more consistent formatting to `<text><value>` elements that contain an `<output>`, e.g.

## Current
```
<text id="/delivery/some:label">
  <value><output value=" /delivery/other "/> asdf </value></text>
```

## With this patch
```
<text id="/delivery/some:label">
  <value><output value=" /delivery/other "/> asdf</value>
</text>
```

dorey:
> the new behaviour's an improvement, but would appreciate your opinion.

Agreed. The old way changed the content inadvertently. This is the right way of doing it. (Haven't looked at the code yet though)

batkinson:
Take a look at my comment here https://github.com/XLSForm/pyxform/issues/96#issuecomment-310744459. The problem is not with the regular expression. The problem is a conceptual one. You can't introduce spaces where none existed by pretty-printing, then collapse them without also eliminating spaces that are essential. You can play around with half-measures, but the real solution is to use a pretty-printer that doesn't add whitespace where none existed. It should only *extend* whitespace. Similarly, when collapsing, you can't safely eliminate whitespace where some existed. That was the problem with #96, it's the mirror image of this.

batkinson:
If you take a look at the raw output before pretty printing, the forms look good are perfect with regards to this. It's making them human-readable that's causing the problem. Minidom's formatter just isn't up for the task.

alxndrsn:
@batkinson thanks for the response.  I don't follow what you're saying about #96 though - this PR fixes the example in the original issue #96 and eliminates the extra leading and trailing whitespace that were introduced by #132.  Can you give an example of what would be handled incorrectly?

batkinson:
> this PR fixes the example in the original issue #96 and eliminates the extra leading and trailing whitespace that were introduced by #132.

I assume you understand what I meant about the conceptual issue. 

For an example, you already gave one yourself. You need a parser to solve this, regular expressions are not capable of comprehending the recursion. That's normally why pretty printers are implemented as parsers and not scanners. Scanners can't identify whitespace leading and trailing an element.

We're going to play whack-a-mole with this until the root cause is addressed.

alxndrsn:
I do understand that regular expressions aren't ideal for parsing XML, and that to get _correct_ results you'd use a proper parser.

> For an example, you already gave one yourself.

This PR handles the example that I gave.  On first reading I thought you were objecting to the PR being merged, but perhaps that's not the case?

batkinson:
> perhaps that's not the case?

Correct, I just wanted to note that this likely isn't solved. I'll take a look at what you did differently. No matter what, it's likely not "right". The reason being that "right" isn't this approach (regular expressions cleaning up a stupid pretty printer).

batkinson:
Yeah, you've captured the idea well in the issue and pointing to that article sums up the problems with using minidom's formatter well. Thanks.

yanokwa:
Closing because 
1. Pretty printing doesn't happen by default anymore
1. It'd be better to fully address #150 
1. @alxndrsn isn't here to defend himself.

PR #148, opened 2017-08-18
Remove no-op XML formatting code

alxndrsn:
From what I can see, this code doesn't do anything.  Possibly I just can't work out a test to demonstrate its current behaviour.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -432,8 +432,6 @@ class Survey(Section):
         output_re = re.compile('\n.*(<output.*>)\n(  )*')
         pretty_xml = text_re.sub(lambda m: ''.join(m.group(1, 2, 3)), xml_with_linebreaks)
         inline_output = output_re.sub('\g<1>', pretty_xml)
-        inline_output = re.compile('<label>\s*\n*\s*\n*\s*</label>')\
-            .sub('<label></label>', inline_output)
         return '<?xml version="1.0"?>\n' + inline_output
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


pyxform/tests_v1/test_whitespace.py
```diff
@@ -13,3 +13,13 @@ def test_over_trim(self):
             xml__contains=[
                 '<label><output value=" /issue96/var "/> text </label>',
             ])
+
+    def empty_label_squashing(self):
```
ukanga: For this test to run it should be named `def test_empty_label_squashing(self):`.

pyxform/survey.py
```diff
@@ -432,8 +432,6 @@ def _to_pretty_xml(self):
         output_re = re.compile('\n.*(<output.*>)\n(  )*')
         pretty_xml = text_re.sub(lambda m: ''.join(m.group(1, 2, 3)), xml_with_linebreaks)
         inline_output = output_re.sub('\g<1>', pretty_xml)
-        inline_output = re.compile('<label>\s*\n*\s*\n*\s*</label>')\
-            .sub('<label></label>', inline_output)
```
ukanga: I think at some particular point the XML generated included a blank label, it does not seem like this is possible anymore, but this functionality was to remove any white space in the label generated. As such, removing this may be of no consequence, I can merge it if the test is removed or an appropriate test is in place.

alxndrsn:
As you say, empty label tags don't seem to be included now.  I've removed the test.  Please let me know if you want commits squashed.

Issue #130, opened 2017-06-14
Convert files with leading numbers in their name


yanokwa:
Files with leading numbers in their name error out with `pyxform.errors.PyXFormError: The name '1-basic' is an invalid xml tag. Names must begin with a letter, colon, or underscore, subsequent characters can include numbers, dashes, and periods.`

This is because pyxform uses the name of the file (e.g., basic.xlsx) as the name of the node inside the instance.
```
    <h:title>basic</h:title>
    <model>
      <instance>
        <basic id="basic">
          <note/>
          <my_name/>
```
So when the name of the file starts with 1, the name of the node will start with 1 and that is not valid XML. I'm not sure what the best fix here would be, but some options to consider:
 * Prepend numbers in the node name with _
 * Use fixed name (e.g., ODK Build uses `data`)

ukanga:
I do like the second option, having a fixed name `data`.  Where I have had to change this is when trying to merge data from two different forms with the same fields using ODK Briefcase. I ended up using `find` and `sed` to search and replace node paths to match up.

ukanga:
Looking at https://github.com/XLSForm/pyxform/blob/master/pyxform/xls2json.py#L900 I think this ought to be something determined by the application using this library i.e 
```python
 data = parse_file_to_json(path, default_name='data')
```

Would this work for you?

yanokwa:
My desired outcome is that [xls2xform.py#L18](https://github.com/XLSForm/pyxform/blob/master/pyxform/xls2xform.py#L18) always uses `data` as the default_name. So if you are proposing that, it'd work great for me! 

@MartijnR does that break anything ecosystem wise? Is `data` a reasonable name?

MartijnR:
Yes, it's fine with me. 

The only disadvantage is that we currently have some protection when people try to edit a record in an amended version of the form. Enketo will show an error that the rootnodes are not the same. This is a much-hated message but it's also nice-to-have protection (against server issues when updating a form). So now it will be mostly up to Ona/KoBo to make sure records are edited in the right form @ukanka @dorey!

ukanga:
We do have such a check in Ona for form replacements, so this will simply apply to all new forms when this is implemented.

dorey:
I agree that `<data>` is a good root node name.

yanokwa:
Sounds like we have consensus! 🎉

PR #152, opened 2017-08-22
Use "data" as Default Node Name

moshthepitt:
Implement use of "data" as root node name that one can optionally override.

Ensure existing tests do not break by having them use a custom node name as expected in each test.

Fix: #130

### DIFF
--- a/pyxform/builder.py
+++ b/pyxform/builder.py
@@ -21,7 +21,7 @@ def copy_json_dict(json_dict):
     items = None
 
     if type(json_dict) is list:
-        json_dict_copy = [None]*len(json_dict)
+        json_dict_copy = [None] * len(json_dict)
         items = enumerate(json_dict)
     elif type(json_dict) is dict:
         json_dict_copy = {}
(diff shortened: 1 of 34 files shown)


lognaturel:
I remember @ukanga saying something about why this hadn't been merged but now I can't remember. Maybe because it's risky? It would be helpful to get a reminder of what's blocking it because I do agree it would be a wonderful change.

ukanga:
@moshthepitt can you rebase this with the latest master so that we can give this another shot at merging it?

moshthepitt:
@ukanga will do

lognaturel:
Would it help if I could arrange for some manual QA on this change? Please let us know what can be done to move this forward! More users have been exploring advanced uses of XPath recently and it would be much easier if the document root name were stable rather than being based on the spreadsheet filename.

moshthepitt:
@ukanga @lognaturel let me try to rebase this by the end of the week then we can take stock on where we are.  Sounds good?

jkpr:
As I wrote over at #335, 

I was looking through the code for the XML root / default node name. I don't know why I missed it before. In the settings, I can simply use `name` as a key and then set the value to what I would like, such as `LS` in my example above.

In the conversations at #130 and #152, I do not see conversation about this setting, so maybe everyone knows about it. One way fix #152 would be to set the `name` key to `data` in the settings tab. This isn't published on xlsform.org, though.

yanokwa:
@ukanga was uncomfortable with the existing solution and now it's very out of date. Closing in favor with #368.

PR #153, opened 2017-08-26
Fixed import error calling for missing module pyxform

ivangayton:
Hi there,

I advised a friend to use PyXform to convert xlsforms to xml locally for testing, and when she cloned the repo and attempted to use the xls2xform.py script, the following error resulted:

```Traceback (most recent call last):
  File "xls2xform.py", line 7, in <module>
    from pyxform import builder, xls2json
ImportError: No module named pyxform
```

This was on OSX. I pulled the latest code to my Ubuntu 16.04 machine, and found the same result: any attempt to run xls2xform, with or without arguments, generates this ImportError.

Looking through the code, it appears that a number of modules call for an import from the module pyxform. I'm not sure why it's been done this way; it appears to make it impossible to use the script directly from the command line in either OSX or Ubuntu, but perhaps there's another use-case for which this import structure works, or is necessary.

In any case, I've removed all import statements directly referencing the pyxform module, and it now works on both my Ubuntu machine and my friend's Mac.

After making these changes in my fork, calling:

```
python xls2xform path/to/infile.xlsx path/to/outfile.xml
```

Results in a success message and a working output file in the specified folder.

However, I have no idea if this somehow compromises other uses of the script! I haven't tested it on a Windows machine (I would think that Python is Python, but I don't know this, so safer not to asume). Furthermore, I don't know that in some embedded application this import structure isn't somehow necessary.

If this doesn't break things for others, it certainly helps those of us who use the script standalone from OSX or Linux command line to develop and test forms, and we'd be grateful if you'd merge it. If it breaks things for other people, no worries, we'll continue forking as necessary.

If it's necessary to keep the import calls to pyxform, I'd appreciate knowing why this is; perhaps I can try to learn what's necessary to make the changes that accommodate both that need and my desire to have a working command-line script to convert forms locally.

Thanks, as always, for creating and maintaining this extraordinary project, which I use on a daily basis for humanitarian and development work.

Ivan

### DIFF
--- a/pyxform/aliases.py
+++ b/pyxform/aliases.py
@@ -1,4 +1,4 @@
-import constants
+from . import constants
 
 # Aliases:
 # Ideally aliases should resolve to elements in the json form schema
--- a/pyxform/aliases.py
+++ b/pyxform/aliases.py
@@ -1,4 +1,4 @@
-from . import constants
+import constants
 
 # Aliases:
 # Ideally aliases should resolve to elements in the json form schema
--- a/pyxform/aliases.py
+++ b/pyxform/aliases.py
@@ -1,4 +1,4 @@
-from pyxform import constants
+import constants
 
 # Aliases:
 # Ideally aliases should resolve to elements in the json form schema
--- a/pyxform/builder.py
+++ b/pyxform/builder.py
@@ -1,7 +1,7 @@
 import copy
 import os
 
-from . import file_utils, utils
+import file_utils, utils
 from errors import PyXFormError
 from question import (InputQuestion, MultipleChoiceQuestion,
                               OsmUploadQuestion, Question, RangeQuestion,
--- a/pyxform/builder.py
+++ b/pyxform/builder.py
@@ -1,16 +1,16 @@
 import copy
 import os
 
-from pyxform import file_utils, utils
-from pyxform.errors import PyXFormError
-from pyxform.question import (InputQuestion, MultipleChoiceQuestion,
+import file_utils, utils
+from errors import PyXFormError
+from question import (InputQuestion, MultipleChoiceQuestion,
                               OsmUploadQuestion, Question, RangeQuestion,
                               TriggerQuestion, UploadQuestion)
-from pyxform.question_type_dictionary import QUESTION_TYPE_DICT
-from pyxform.section import GroupedSection, RepeatingSection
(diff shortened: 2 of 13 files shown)


ivangayton:
Ok, fails CI tests. I guess the imports are somehow necessary. I have no idea how to fix this, so I'll withdraw the pull request and maybe post an issue!

dorey:
Hi Ivan. Looks like it's almost there. [The tests are failing when run in python3](https://travis-ci.org/XLSForm/pyxform/builds/268650658) which no longer imports paths relatively by default. I think that importing the packages with an explicitly relative path will fix this.

For example, this line in your PR:

> `import constants`

can be rewritten as

`import .constants`

(which is the same as `from . import constants`)

And, if you want to go up a directory, you can do `import ..constants`


And if you want to test this out locally, you can create a new virtualenv with python3, which should be quite painless. Something like this should demonstrate the problem locally.

```
virtualenv -p python3 pyxform_python3_env
. pyxform_python3_env/bin/activate
python setup.py develop
pip install nose
nosetests --tests pyxform/tests_v1/
```

ivangayton:
Hi Dorey,

Many thanks for your feedback and suggestion. Had a go using the `from . import constants` syntax, and it broke comprehensively! 

I'm now reading my Stack Overflow on the intricacies of relative imports in Python3, which turns out to be rather complicated. I imagine I'll figure it out, though probably not right away, and in the meantime nobody is inconvenienced (my colleagues and I can use my fork, which works in Python2.x, and whoever is using it the way it is here can continue to do so)! 

So putting this pull request back to sleep for now, and I'll get back to it when I've got a bit of time to figure out the relative import thing.

Thanks again!

ivan

lognaturel:
@ivangayton Did you follow the instructions in the readme exactly and in particular run `setup.py`? I haven't totally understood what issue you're having but I can say that I work from source on OS X and don't have this problem. Also consider that if you don't need to modify the source, you can use `pip` for easier installation.

ivangayton:
Hi @lognaturel, no, sorry, didn't read the instructions! Just assumed when it stopped working the way I'd been using it for years (without installing, just calling the script from the repo folder) that something was wrong. 

Sorry for wasting people's time, thanks for help and responses, I'll install properly when time permits!

Ivan

PR #155, opened 2017-09-07
Upgrade xlrd

yanokwa:
xlrd changelog is at https://xlrd.readthedocs.io/en/latest/changes.html.

To test his change, I ran `nosetests` and got this output.

 ```
$ nosetests
........................................................................usage: nosetests [-h] [--json] [--skip_validate] path_to_XLSForm output_path
nosetests: error: too few arguments
..........................................................
Ran 130 tests in 16.332s

OK 
```

I think this is a safe change, but I don't know pyxform that well. It'd be prudent for the maintainers (or maybe @sheppard) to consider the implications of this change before merging.

### DIFF
--- a/requirements.pip
+++ b/requirements.pip
@@ -1,6 +1,6 @@
 nose==1.3.7
 FormEncode==1.3
 modilabs-python-utils==0.1.5
-xlrd==1.0.0
+xlrd==1.1.0
 unittest2==1.1.0
 unicodecsv==0.14.1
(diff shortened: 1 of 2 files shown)


sheppard:
I think the nosetests error was there before, though it would be good to sort out eventually.

What do you think about loosening the spec to `xlrd>=1.0.0` (or `xlrd>=1.0.0,<2.0.0`)?  I doubt xlrd's API is going to change much in the future, at least not with out a major version bump.  We could probably do the same for unicodecsv.

yanokwa:
I want to keep this PR targeted, so I'm going to merge just this point releases. 

Further, xlrd seems to update once a year (https://github.com/python-excel/xlrd/releases) so I think manually changing the versions shouldn't be too big of a burden.

PR #156, opened 2017-09-21
Upgrade from xlrd 1.0.0 to 1.1.0

jpic:
This fixes version conflicts with other packages which rely on the latest release of xlrd, such as wq/wq#27


lognaturel:
@jpic I believe this is not quite complete but that #155 does the full update!

jpic:
Awesome, thanks !

PR #164, opened 2017-12-12
Python 3,unichr not defined in util.py error

### LINKED ISSUE #161: Python 3,unichr not defined in util.py error
image
in util.py line no 18 , unichr is not defined error is coming when I am using pyxform in my plugin in QGIS 3 (i.e python 3 environment)

yjacolin:
Fix #161

### DIFF
--- a/pyxform/utils.py
+++ b/pyxform/utils.py
@@ -15,7 +15,7 @@ except NameError:
 else:
     unicode = unicode
     basestring = basestring
-    unichr = unichr
+    unichr = chr
 
 SEP = "_"
 


ukanga:
This change breaks the current functionality as the failing tests in one python2.7 version indicates. I would expect for python 3.6 the code to be executed will be in the except clause https://github.com/XLSForm/pyxform/blob/master/pyxform/utils.py#L11 which does exactly as the change you are proposing. I suspect there maybe something else that does inject `unicode()` in your version of python 3.6 ans as such ends up in the else clause. That should be looked into on your end. Otherwise, this is a braking change that we should not accept.

ukanga:
See https://github.com/XLSForm/pyxform/pull/164#issuecomment-354723090

I will suggest you test this on the python terminal in isolation and see if the similar error occurs.
```python
>>> from pyxform.utils import unichr
>>> unichr(10000)
```

ukanga:
Addressed by #179.

shivareddyiirs:
@ukanga 
Thanks. I have another request to make new release of pyxform.

PR #165, opened 2017-12-13
ODK xml header for length over 255 characters

Archaejohn:
This change allows for ODK Aggregate to create the proper database string length with the excel column "bind::odk:length" that contains an integer of the appropriate string length, as well as fields longer than 255 characters

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -21,6 +21,7 @@ nsmap = {
     u"xmlns:ev": u"http://www.w3.org/2001/xml-events",
     u"xmlns:xsd": u"http://www.w3.org/2001/XMLSchema",
     u"xmlns:jr": u"http://openrosa.org/javarosa",
+    u"xmlns:orx": u"http://openrosa.org/xforms",
     u"xmlns:odk": u"http://www.opendatakit.org/xforms"
     }
 


lognaturel:
Awesome! The `orx` namespace could be used for other attributes so I think this should add the `odk` namespace without removing the `orx` one.

lognaturel:
How brilliant, there was a test to make sure the `orx` namespace wasn't removed! That's why the build is failing @Archaejohn.

lognaturel:
@Archaejohn also note that the settings sheet can have a `namespaces` column. Below the header you could add `odk="http://www.opendatakit.org/xforms"` to be able to use it without modifications to pyxform.

If @Archaejohn doesn't update the PR to bring back `orx` I think this can be closed.

Archaejohn:
I will update that.  The openrosa server was down and It seemed to be pointing somewhere else.

Archaejohn:
Or pointing to a non existent source.

lognaturel:
Ah yes, the openrosa server changed hands a long time ago. But "the namespace specification does not require nor suggest that the namespace URI be used to retrieve information; it is simply treated by an XML parser as a string" -- so it doesn't really matter.

Archaejohn:
Got it I closed this one because the namespace array location changed to utils.py and made the changes and created a new pull request.  Sorry about the confusion.  I have a compiled version of my original one working but I was getting warnings on the server logs but nothing fatal and everything seemed to work.  I'm guessing this fixes those warnings.

lognaturel:
👌

PR #166, opened 2017-12-15
Workarounds for PyInstaller on Windows

yanokwa:
I use pyxform inside a cross-platform app (https://github.com/opendatakit/xlsform-offline) that is packaged by PyInstaller. In the past, I maintained my own fork of pyxform, but I figured I'd send this patch upstream so I don't have to maintain my fork. The changes are...

1. `subprocess.Popen` will open a command window when run from a PyInstaller-packaged app with the `--noconsole` option. There is code at https://github.com/pyinstaller/pyinstaller/wiki/Recipe-subprocess which is based on a GPL project, so I used code that I had from a few years back that does the same thing.

2. The `_java_installed` implementation spawns all sorts of windows when in a PyInstaller-packaged app. I've changed it to a call that doesn't spawn windows.

I tested this by installing my branch and running `xls2xform.exe` on a Windows 10 machine using this form ([validate-error.xlsx](https://github.com/XLSForm/pyxform/files/1561371/validate-error.xlsx)) with a validation problem. I confirmed that validation happened when Java was in the path. I confirmed that I got a `pyxform odk validate dependency: java not found` error when Java was not in the path. I also confirmed the above behavior on a macOS machine.

### DIFF
--- a/pyxform/odk_validate/__init__.py
+++ b/pyxform/odk_validate/__init__.py
@@ -60,7 +60,7 @@ def _java_installed():
     stderr = None
     java_regex = re.compile("java version")
     try:
-        stderr = run_popen_with_timeout(["java", "-version"], 100)[3]
+        stderr = run_popen_with_timeout(["java", "-version"], 100)[3].strip().decode('utf-8')
     except:
         pass
     
--- a/pyxform/odk_validate/__init__.py
+++ b/pyxform/odk_validate/__init__.py
@@ -7,6 +7,7 @@ A python wrapper around ODK Validate
 import os
 import re
 import sys
+import subprocess
 from subprocess import Popen, PIPE
 import threading
 import signal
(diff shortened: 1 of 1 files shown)


ukanga:
@yanokwa Is ODK Validate jar supposed to only work with oracle java? Was running some tests with openjdk installed and the java installed check is failing, specifically the check `java_regex = re.compile("java version")` does not appear in openjdk installed java.

OpenJDK output
```
openjdk version "9"
OpenJDK Runtime Environment (build 9+181)
OpenJDK 64-Bit Server VM (build 9+181, mixed mode)
```

Oracle Java output
```
java version "1.8.0_152"
Java(TM) SE Runtime Environment (build 1.8.0_152-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.152-b16, mixed mode)
```
Can you resolve this or should I go ahead and revert this PR? I apologize, my test on this was with Oracle java when I was merging the PR.

yanokwa:
Ugh! It should work with all the Javas. I'll send in a PR.

Issue #163, opened 2017-12-07
add 2nd XForm validator option


MartijnR:
I'm in the process of building an [Enketo-form-engine-based nodeJS XForm validator](https://github.com/enketo/enketo-validate). I hope you will agree this will nicely enrich the ODK ecosystem (assuming it's done fairly well). I think the command line interface will be identical to ODK Validate.

The main reason for this project is that some adopters use customized Enketo engines with custom XPath functions (and they'll be using a custom version of the new validator). 

This issue is to discuss adding a 2nd validator option to pyxform with a configuration item to swap from ODK Validate to the new validator. The default would remain ODK Validate.

Please let me know if you any ideas, comments, concerns!

lognaturel:
Could warning rather than throwing an error on missing function address the problem at least in the short term? https://github.com/opendatakit/validate/issues/35 It seems this could be a simpler approach though it does have the downside of pushing the validation of actual function names to the client.

MartijnR:
Thanks for bringing that up. That development would help some people with being able to at least use custom functions.

However, one difficulty is with training users that some 'function not supported' warnings are okay and others are not. They also would miss out on parameter checks for those function calls.

I think, generally it is better to output an error for the use of unsupported XPath functions (if the validator kept up-to-date with the spec).

MartijnR:
Here's a web interface around the new validator (and also ODK validate): https://validate.enketo.org/. Comments welcome here: https://github.com/enketo/enketo-validate/issues

PR #167, opened 2017-12-15
Enketo validate #163

lindsay-stevens:
PR for #163. More or less the same as the ODK Validate workflow but using a JS validator which @MartijnR is working on.

I've left a stub JavaScript file called "enketo_validate.js" which is called with NodeJS and the stdout/stderr result piped back to pyxform, sufficient for the integration test to pass. The actual validation app might be much more complex but I suppose at some level there'd be a single entry point, possibly with a different name / location.

This new validator is off by default, but can be activated with a command line switch "--enketo_validate".


MartijnR:
Thanks! I look forward to testing it. I think in a few days, version 0.0.1 of the new validator will be ready.

MartijnR:
Version 1.0.0 has launched! I've sent you a PR to the same branch, to discuss whether that's the right way to add it.

I wonder if we should use the following approach for the validation logic instead:

1. Either ODK Validate (default) or Enketo Validate, but not both
2. `--skip_validate` would skip both options (and only do internal pyxform validation)

MartijnR:
The `replace_function` logic in `odk_validate/__init.py@_cleanup_errors`  to change `/data/path/to/node` into `${node}` would also be useful for other validators as well. Initially I thought of including an `--xlsform` flag in enketo-validate, but think it might be better to do this here.

yanokwa:
The use of `popen` in this `pyxform/validators/enketo_validate/__init__.py` will cause problems for forks who bundle pyxform with Pyinstaller. I have a PR at https://github.com/XLSForm/pyxform/pull/166/files that fixes this for ODK Validate. You might want to integrate some of those changes.

lindsay-stevens:
This problem is also faced in #167, where we're faced with potentially vendorising a JS project (enketo-validate) into pyxform.

What I suggest is to create a utility / CLI tool that a user can invoke which will download the relevant validator dependencies if they don't already exist.

If everyone is comfortable with it, this download process could be automatic, as part of the `setup.py` install. On the one hand it might be annoying if you were someone who had no interest in using the validators, but on the other hand not making it automatic would mean a two-step install process for those (the majority?) that are interested in validated XForms.

The ongoing maintenance then would be to update the fetch URL for the validators. This could also be avoided if both odk_validate and enketo-validate publish their latest compiled releases in a way that has a "current" URL that points to the latest version. 

For example if both projects uploaded their compiled packages as GitHub release attachments, then the download link can be reliably parsed from the GitHub API, e.g. https://api.github.com/repos/opendatakit/validate/releases/latest returns a JS document where the attachment URLs are populated in the "assets" array. Attaching artifacts to GitHub releases can be automated with Travis & AppVeyor based on git tags too, so that the whole process automagical.

I think this would be in scope for #167 so please let me know if this would be a supported strategy and I'll have a go at implementing it.

lindsay-stevens:
Hi @MartijnR and @yanokwa - thanks for reviewing. I've incorporated the subprocess changes and error cleanup suggestions. 

For the validator selection - I agree there should be a way to run one, both, or neither of the validators. It's possible at the moment to make all choices, but the flag names feel a bit confusing:

- `xls2xform.py myform --skip_validate`: no validators
- `xls2xform.py myform`: ODK only
- `xls2xform.py myform --skip_validate --enketo_validate`: Enketo only
- `xls2xform.py myform --enketo_validate`: both validators

Maybe it should look like this? Note that it would mean changing the default behaviour to "no validators":

- `xls2xform.py myform`: no validators
- `xls2xform.py myform --odk_validate`: ODK only
- `xls2xform.py myform --enketo_validate`: Enketo only
- `xls2xform.py myform --odk_validate --enketo_validate`: both validators

Let me know what you think and I'll give it a shot.

lognaturel:
I think it would be ideal for flag combinations that work today to maintain their meaning. How about 
**backwards-compatible**:
`xls2xform.py myform --skip_validate`: no validators
`xls2xform.py myform`: ODK only

**new**:
`xls2xform.py myform --enketo_validate`: Enketo only
`xls2xform.py myform --odk_validate`: ODK only
`xls2xform.py myform --enketo_validate --odk_validate`: both
`xls2xform.py myform --enketo_validate --odk_validate --skip_validate`: no validators

It makes internal logic a little less clean but it makes this into a non-controversial change since external interfaces don't change.

MartijnR:
The compromise @lognaturel proposed sounds good to me.

In the future, we can perhaps create the perfectly logical ui that @lindsay-stevens proposed for a major (backwards-compatibility-breaking) release.

Thanks!

pyxform/validators/odk_validate/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class ODKValidateError(Exception):
+    pass
+
+
+def _java_installed():
+    stderr = run_popen_with_timeout(["java", "-version"], 100)[3]
+    stderr = stderr.strip().decode('utf-8')
+    return "java version" in stderr
```
ukanga: I have raised a question on this https://github.com/XLSForm/pyxform/pull/166#issuecomment-354714933, maybe the fix should be handled in this pull request.

pyxform/survey.py
```diff
@@ -513,7 +514,8 @@ def insert_output_values(self, text):
             return result, not result == xml_text
         return text, False
 
-    def print_xform_to_file(self, path=None, validate=True, pretty_print=True, warnings=None):
+    def print_xform_to_file(self, path=None, validate=True, enketo=False,
```
ukanga: I think we do need the enketo option also available with the `survey.to_xml()` function in this same class since applications like https://github.com/onaio/onadata make use of it to get the XML string rather than write to file.

pyxform/xls2xform.py
```diff
@@ -12,7 +12,7 @@
 """
 
 
-def xls2xform_convert(xlsform_path, xform_path, validate=True, pretty_print=True):
+def xls2xform_convert(xlsform_path, xform_path, validate=True, enketo=False, pretty_print=True):
```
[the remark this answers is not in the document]
lindsay-stevens: No worries - added in ea5f93b

pyxform/survey.py
lindsay-stevens: No worries - added in ea5f93b

pyxform/validators/odk_validate/__init__.py
lindsay-stevens: No worries - added in b803142

lindsay-stevens:
Thanks @lognaturel I added this CLI in 35f85df

pyxform/validators/enketo_validate/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class EnketoValidateError(Exception):
+    pass
+
+
+def _node_installed():
+    stdout = run_popen_with_timeout(["node", "--help"], 100)[2]
+    stdout = stdout.strip().decode('utf-8')
+    return "Usage: node" in stdout
```
ukanga: Should this fail if you do not have `nodejs` installed? I suppose `nodejs` is now a hard dependency for `pyxform` same way `java` is at the moment? 
```python
======================================================================
ERROR: test_todo (pyxform.tests.enketo_validate_tests.EnketoValidateTests)
  File "/..../pyxform/pyxform/tests/enketo_validate_tests.py", line 17, in test_todo
    ["Enketo Validate Warnings:\nHello world\n"], check_xform(""))
  File "/..../pyxform/pyxform/validators/enketo_validate/__init__.py", line 30, in check_xform
    if not _node_installed():
  File "/..../pyxform/pyxform/validators/enketo_validate/__init__.py", line 15, in _node_installed
    stdout = run_popen_with_timeout(["node", "--help"], 100)[2]
  File "/..../pyxform/pyxform/validators/util.py", line 37, in run_popen_with_timeout
    startupinfo=startup_info)
  File "/usr/lib64/python3.6/subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "/usr/lib64/python3.6/subprocess.py", line 1344, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'node': 'node'

```

lindsay-stevens: Node would only be a hard dependency for devs / users that run the tests - but to avoid that issue I added a conditional skip in 952a940

pyxform/validators/odk_validate/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        ["java", "-jar", ODK_VALIDATE_JAR, path_to_xform], 100)
+    warnings = []
+    stderr = decode_stream(stderr)
+
+    if timeout:
+        return ["XForm took to long to completely validate."]
+    else:
+        if returncode > 0:  # Error invalid
+            raise ODKValidateError(
+                'ODK Validate Errors:\n' + ErrorCleaner.odk_validate(stderr))
```
ukanga: I think this is cleaning up way more errors than it should. While testing pip installing this I was unable to discern what error was being returned, removing the ErrorCleaner allowed me to see the actual error.
````python
xls2xform --odk_validate ~/Documents/moods.xlsx /tmp/mooods.xml                                     (pyxform) remotes/lindsay-stevens/enketo-validate ✱
  File "/..../.virtualenvs/pyxform/bin/xls2xform", line 11, in <module>
    load_entry_point('pyxform==0.10.0', 'console_scripts', 'xls2xform')()
  File "/..../.virtualenvs/pyxform/lib/python3.6/site-packages/pyxform/xls2xform.py", line 137, in main_cli
    enketo=args.enketo_validate)
  File "/..../.virtualenvs/pyxform/lib/python3.6/site-packages/pyxform/xls2xform.py", line 27, in xls2xform_convert
    warnings=warnings, enketo=enketo)
  File "/..../.virtualenvs/pyxform/lib/python3.6/site-packages/pyxform/survey.py", line 538, in print_xform_to_file
    warnings.extend(odk_validate.check_xform(path))
  File "/..../.virtualenvs/pyxform/lib/python3.6/site-packages/pyxform/validators/odk_validate/__init__.py", line 53, in check_xform
    'ODK Validate Errors:\n' + stderr)
Error: Unable to access jarfile /..../.virtualenvs/pyxform/lib/python3.6/site-packages/pyxform/validators/odk_validate/ODK_Validate.jar
```

lindsay-stevens: Would you be able to share the test file / elaborate please? I copied the existing cleaner functions and the tests for them are passing so I'm not sure where the issue is.

ukanga: Before the fix in the commit d600391c2a48cf97cc116b71948f48870ee88a39, I did 
```bash
$ pip install .

$ xls2xform ~/Documents/moods.xlsx /tmp/moods.xml
```
Which resulted in the exception:
```python                                 
  File "/..../.virtualenvs/pyxform/bin/xls2xform", line 11, in <module>
    load_entry_point('pyxform==0.10.0', 'console_scripts', 'xls2xform')()
  File "/..../.virtualenvs/pyxform/lib/python3.6/site-packages/pyxform/xls2xform.py", line 137, in main_cli
    enketo=args.enketo_validate)
  File "/..../.virtualenvs/pyxform/lib/python3.6/site-packages/pyxform/xls2xform.py", line 27, in xls2xform_convert
    warnings=warnings, enketo=enketo)
  File "/..../.virtualenvs/pyxform/lib/python3.6/site-packages/pyxform/survey.py", line 538, in print_xform_to_file
    warnings.extend(odk_validate.check_xform(path))
  File "/..../.virtualenvs/pyxform/lib/python3.6/site-packages/pyxform/validators/odk_validate/__init__.py", line 53, in check_xform
    'ODK Validate Errors:\n' + ErrorCleaner.odk_validate(stderr))



````

Notice, compared to the code block in https://github.com/XLSForm/pyxform/pull/167#discussion_r159414789 the code block above is missing the error message below. As such I was not able to tell what was wrong until I removed the clean `ErrorCleaner.odk_validate()` method to see what was in `stderr` which is the message below.

```python
Error: Unable to access jarfile /..../.virtualenvs/pyxform/lib/python3.6/site-packages/pyxform/validators/odk_validate/ODK_Validate.jar
```

lindsay-stevens: Sorry, my bad - added a fix for silencing single line errors in a817103.

Also, when I added a test for the jarfile message it was being changed by the token substitution (e.g. /path/myitem -> ${myitem}) so I've explicitly skipped errors that start with "Error: Unable to access jarfile". It's not great but seemed like a better choice than trying (and I did try) to find a regex that could reliably tell the difference between an XPath and a file path, and not introduce regressions.

lindsay-stevens:
just checking - is this ok to merge?

lognaturel:
@yanokwa I think you may have had some concerns about some of the compatibility code for Windows? Were these addressed? Or should they be addressed separately?

yanokwa:
@lognaturel My concerns have been addressed! The code I was looking for was moved to `pyxform/validators/util.py` and if I find problems with it, I'll send in a PR.

lindsay-stevens:
Great - thanks all!

Issue #107, opened 2017-04-04
Add type=xml-external


MartijnR:
We have the convenience types `select_one_from_file name.xml` and `select_multiple_from file name.xml` and a convenience function `pulldata` that all 3 automagically add an `<instance>` to the model.

This issue, for discussion, is to explicitly only add an instance, e.g.

type         | name |
----------|-------|
xml-external  | mydata |

produces:

```xml
<instance id="mydata" src="jr://file/mydata.xml" />`
```

This would cater to complex calculations that query external data without restrictions, for example:

type | name | calculation
-----|-------|-----------
integer | rooms | 
calculate | c         | count(instance('mydata')/house[rooms = current()/../rooms ])

Note that this might make `select_one_from_file` and `select_multiple_from_file` redundant (which is totally fine) , as you could feasibly use type=xml-external and type=select_one to produce the same XForm output.

Updated: type now named `xml-external`

MartijnR:
I changed the above type to `xml-data` (and the idea is that in the future `csv-data` may be added). 

Is anybody opposed to, or excited about, a PR that adds this feature?

lognaturel:
This is nice! There are a couple of things that come to mind that perhaps could be handled separately but that may affect the `type` name.

1) Generating the XML from the choices sheet. I understand you are probably imagining servers would generate the files but I think it is useful to be able to generate them in the form-building process as well. I'm imagining this would be just like for `select`s but instead of writing an internal instance an external instance would be generated. For example, if I have a file reference called `mydata` and a `list_name` `mydata` in the choices sheet then `mydata.xml` would be generated accordingly. If `mydata` isn't in the choices sheet then the behavior is as in #169. One slight complication I see is that there might need to be a new mechanism introduced to specify which columns to include. `select`s use the `choice_filter`.

2) Generating and using internal instances in addition to external ones. There are small, static data sets that are useful to do XPath queries on. In that case, it's more convenient for them to be in the form file rather than as separate attachments. It's currently possible to fake this with a dummy `select` and a `choice_filter` that forces extra columns to be included in the internal secondary instance but it would be great for this to be a first-class citizen (see what I mean at https://docs.google.com/spreadsheets/d/1mQGVWaJA4QQp3jtddMdcRSDjc1O-j_bB79GUnlj5k10/edit#gid=0). This would be most convenient in combination with 1) above. 

[...]

PR #169, opened 2017-12-19
Feature/external instance #107

lindsay-stevens:
PR for #107. More or less the same as pulldata and select_one_from_file etc. in terms of adding a supplementary instance to the form model, but this just does the instance part without any effects elsewhere on the form (no controls, bindings, etc.).

Includes some tidy up of the other functions that generate instances to make it possible to validate across all of them e.g. to check that duplicates aren't being added. Also I think this may qualify as a bug fix but for example if a user has a form using the same select from file e.g. `select_one_from_file cities.csv` with item1 and item2, then 2 identical copies of the instance would be added - now only one is added.

Includes some other minor cleanups:

- Survey's `__repr__` calling `unicode(self)` caused a recursion crash, now uses `__unicode__` instead
- Exception that was acting like a NotImplementedError is now one
- PyxformTestCase didn't show why the compilation failed but now it does

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -450,7 +450,7 @@ class Survey(Section):
         return '<?xml version="1.0"?>\n' + inline_output
 
     def __repr__(self):
-        return unicode(self)
+        return self.__unicode__()
 
     def __unicode__(self):
         return "<pyxform.survey.Survey instance at %s>" % hex(id(self))
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -261,6 +261,15 @@ class Survey(Section):
           list_name in the choices sheet. This is done to allow users to re-use
           external sources without duplicate instances being generated in the
           XForm document. However, it does require careful in form design.
+
+        There are two other things currently supported by pyxform that involve
+        external files and are not explicitly handled here, but may be relevant
+        to future efforts to harmonise / simplify external data workflows:
+
+        - `search` appearance/function: works a lot like pulldata but the csv
+          isn't made explicit in the form.
+        - `select_one_external`: implicitly relies on a `itemsets.csv` file and
+          uses XPath-like expressions for querying.
         """
         instances = []
         for i in self.iter_descendants():
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -139,7 +139,6 @@ class Survey(Section):
             type=u"choice",
             context=u"survey",
             name=list_name,
-            unique_id=list_name,
             instance=node(
                 "instance",
                 node("root", *instance_element_list),
@@ -157,7 +156,6 @@ class Survey(Section):
                     n=element[u"parent"][u"name"]
                 ),
                 name=element[u"name"],
-                unique_id=(element[u"name"],),
                 instance=element.xml_instance()
             )
 
@@ -205,7 +203,6 @@ class Survey(Section):
                             n=element[u"parent"][u"name"]
                         ),
                         name=file_id,
-                        unique_id=(file_id, uri),
                         instance=node(
                             "instance",
                             id=file_id,
@@ -227,7 +224,6 @@ class Survey(Section):
                     n=element[u"parent"][u"name"]
                 ),
                 name=file_id,
-                unique_id=(file_id, uri),
                 instance=node(
                     "instance",
                     node("root",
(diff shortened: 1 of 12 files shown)


pyxform/tests_v1/test_xmldata.py
```diff
@@ -0,0 +1,191 @@
...
+          <item>
+            <name>_</name>
+            <label>_</label>
+          </item>
+        </root>
+      </instance>
+""",
+                '<instance id="fruits" src="jr://file-csv/fruits.csv"/>'
+            ]
+        )
```
MartijnR: An idea for an additional test:
1. xml-data with name "fruits"
2. a calculate with pulldata("fruits", etc)

Then check that the instance is only added once (and no error).

MartijnR:
Judging by the tests this is very thorough! 👍  

I'm not able to technically review this, but wanted to mention that I am very excited about this additional functionality, as we'll be able to go far beyond `pulldata()` to query external data. I think it's a **really big milestone**, especially when form-builders (such as KoBoForm/Ona) start leveraging this feature in a way that is easy for users.

pyxform/tests_v1/test_xmldata.py
lindsay-stevens: Thanks for reviewing, I have added a test for this named "test_no_duplicate_with_pulldata"

pyxform/survey.py
```diff
@@ -112,64 +113,180 @@ def xml(self):
...
+          considered invalid if there is a duplicate name. This differs from
+          other item types which allow duplicates if not in the same group.
+        - for all instance sources, if the same instance name is encountered,
+          only the first instance definition will be output, even if the
+          instance definitions are different (i.e. external XML, external CSV,
+          or select_* placeholder instances). The "first instance" is
+          determined by the item position in the survey sheet, then by the
+          list_name in the choices sheet. This is done to allow users to re-use
+          external sources without duplicate instances being generated in the
+          XForm document. However, it does require careful in form design.
```
lognaturel: This is a very useful comment, thank you. There are two other things currently supported by pyxform that involve external files:
- `search` appearance/function -- works a lot like pulldata but the csv isn't made explicit in the form.
- `select_one_external` -- implicitly relies on a `itemsets.csv` file and uses XPath-like expressions for querying. 

I know this code and comment are about writing instances out to the form XML but perhaps a quick mention of them makes sense so they stay on our radar as features that should be harmonized and simplified.

lindsay-stevens: Thanks - added in 0290fc1

pyxform/tests_v1/test_xmldata.py
```diff
@@ -0,0 +1,292 @@
...
+                      <item>
+                        <itextId>static_instance-cities-0</itextId>
+                        <name>1</name>
+                      </item>
+                    </root>
+                  </instance>
+            """
+        self.assertNotIn(unwanted_extra_cities, xml)
+
+    def test_no_duplicate_with_pulldata(self):
```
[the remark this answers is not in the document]
lindsay-stevens: Thanks - addressed in 8950324. I think the silent-ignore behaviour was probably going to cause a lot confusion, so an error is raised instead, since the clash has to be fixed by the user either way.

lindsay-stevens:
Thanks @ukanga!

@lognaturel and @MartijnR is it OK with you guys for merge?

lognaturel:
@lindsay-stevens I posted on the ODK forum yesterday to see if we could get another brain or two on it. @dcbriccetti tried the PR out and says he got output like 
```
<instance id="towns" src="jr://file/towns.xml">
  <root>
    <item>
      <name>_</name>
      <label>_</label>
    </item>
  </root>
</instance>
```
Which doesn't look right.

lindsay-stevens:
@lognaturel @dcbriccetti thanks for reviewing - this kind of output is to allow external selects to re-use an external instance previously declared using `xml-external` or a `pulldata` calculate, as shown here:

- [test_can__reuse_xml__external_then_selects](https://github.com/XLSForm/pyxform/pull/169/files#diff-b0caaff0263295cb261ff2179eb412ccR336)
- [test_can__reuse_csv__pulldata_then_selects](https://github.com/XLSForm/pyxform/pull/169/files#diff-b0caaff0263295cb261ff2179eb412ccR284)

The problem is that ODK Validate doesn't like external instances for selects unless they have a root/item/name|label path to be found, as per #87. I think in enketo / collect if the external data only requires the `src` attribute it'll ignore this dummy content, but if that's not the case I could update the validations to reject re-use by selects? Let me know.

lognaturel:
Ahh, yes, the templating was ringing a vague bell but I couldn't remember what it was about. 

Maybe we can go with this for now, fix it in JavaRosa/Validate and then remove the dummy content in a later PR. @dcbriccetti does that sound like a good path forward?

MartijnR:
Looks ready to me!

Yes, that dummy content is purely there to fool ODK Validate. Sorry, I should have posted an issue for this in the ODK Validate repo. (In Enketo any content of an external instance is removed, after which it is populated from the `src` reference.)

dcbriccetti:
I see that Validate complains with the `file-csv` host type, but does it complain for the `file` type? I’m thinking not. It doesn’t matter. We can change JavaRosa to ignore any children of the instance tag for `jr://file/...`.

MartijnR:
I think the ODK Validate issue occurs if an XPath predicate is used in an `<itemset>` nodeset attribute value that points to an external instance (without content). 

The content of the instance isn't defined, so as long as the instance element itself exists any reference to a node inside this instance should pass (if it's valid syntax).

lindsay-stevens:
Great - thanks all!

PR #170, opened 2018-01-02
Add OpenJDK support

yanokwa:
Fixes the issue raised at https://github.com/XLSForm/pyxform/pull/166#issuecomment-354714933.

Tested the regex by putting the expression and test string at https://pythex.org. Also tested it by running this and getting three yeses.

```python
import re

test_1 = "java version 1.8.0_152"
test_2 = "openjdk version 9"
test_3 = "openjdk version 9 java version 1.8.0_152"

java_regex = re.compile("(java|openjdk) version")
if java_regex.match(test_1):
    print test_1, 'yes'
if java_regex.match(test_2):
    print test_2, 'yes'
if java_regex.match(test_3):
    print test_3, 'yes'
```

### DIFF
--- a/pyxform/odk_validate/__init__.py
+++ b/pyxform/odk_validate/__init__.py
@@ -58,7 +58,7 @@ def run_popen_with_timeout(command, timeout):
 
 def _java_installed():
     stderr = None
-    java_regex = re.compile("java version")
+    java_regex = re.compile("(java|openjdk) version")
     try:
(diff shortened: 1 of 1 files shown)


lognaturel:
Is there an automated test that can be added so there's a record of the string formats that are expected?

yanokwa:
Replaced by b8031426e025f035e5278b0c3397b7ff03d808f0

Issue #64, opened 2016-05-03
Avoid relying on a compiled JAR file from external PRs


dorey:
So far, external contributions of changes to the ODK validate Jar files have been helpful, but it would be much healthier for the repository to have a different way for the JAR to be compiled (perhaps locally after installation)

dorey:
I've been assured that we can compile JAR files during the `setup.py` installation step so that we can avoid committing binaries into the repo.

lognaturel:
JavaRosa is now published to the maven central repository and included in Java-based tools by gradle. As described in #113 and opendatakit/javarosa#58, it would be great if the dependency could be on JR rather than Validate. Even better if it could be pulled from maven central!

lindsay-stevens:
This problem is also faced in #167, where we're faced with potentially vendorising a JS project (enketo-validate) into pyxform.

What I suggest is to create a utility / CLI tool that a user can invoke which will download the relevant validator dependencies if they don't already exist.

If everyone is comfortable with it, this download process could be automatic, as part of the `setup.py` install. On the one hand it might be annoying if you were someone who had no interest in using the validators, but on the other hand not making it automatic would mean a two-step install process for those (the majority?) that are interested in validated XForms.

The ongoing maintenance then would be to update the fetch URL for the validators. This could also be avoided if both odk_validate and enketo-validate publish their latest compiled releases in a way that has a "current" URL that points to the latest version. 

For example if both projects uploaded their compiled packages as GitHub release attachments, then the download link can be reliably parsed from the GitHub API, e.g. https://api.github.com/repos/opendatakit/validate/releases/latest returns a JS document where the attachment URLs are populated in the "assets" array. Attaching artifacts to GitHub releases can be automated with Travis & AppVeyor based on git tags too, so that the whole process automagical.

I think this would be in scope for #167 so please let me know if this would be a supported strategy and I'll have a go at implementing it.

MartijnR:
Thanks. I like this, and combining it with #167 seems fine to me (and I guess, upgrade to ODK Validate 1.8.0 in the same go if it takes the latest). We should leverage your involvement to the max @lindsay-stevens! 

Does that sound okay (@yanokwa @ln, others)?

(For some reason https://api.github.com/repos/enketo/enketo-validate/releases/latest is still empty even though there have been 3 releases. Will try to figure out why. https://github.com/enketo/enketo-validate/issues/24)

lognaturel:
[...]

PR #171, opened 2018-01-08
external validator installer / updater

lindsay-stevens:
As per #163 and #64, this PR adds a tool for updating validators from GitHub latest release asset file.

In the case of #163 (Enketo-validate) this would be the only delivery mechanism. Maybe at some point the ODK_Validate.jar could be removed from the repo as well - but this also relies on the jar being uploaded to each release. I toyed with grabbing ODK Validate from CircleCI builds but it's less controlled to do that I think.

There's still some more left to do (edit: done!):

- [x] post-update check with XForm to make sure the update succeeded
- [x] roll-back to previous version on fail in case it failed (technically, upgrade stops if check fails)
- [x] integrate with main CLI (xls2xform) e.g. warn if enketo-validate is still just the stub file.
- [x] unit tests

Rather than being part of the `xls2xform`, a new console script is added such that these two are equivalent. It seemed like more trouble than it was worth to integrate them in a backwards compatible way, and this updater tool is a separate chunk of functionality anyway.

```
python pyxform/validators/update.py enketo list
pyxform_validator_update enketo list
```

If enketo-validate hasn't been installed, the following error would be raised upon trying to use it:

```python
EnvironmentError(
            "Enketo-validate dependency not found. "
            "Please use the updater tool to install the latest version.")
```

The full branches of this tool's menus look like this:

```
pyxform_validator_update enketo list
pyxform_validator_update enketo update [file_name] --force
pyxform_validator_update enketo check

pyxform_validator_update odk list
pyxform_validator_update odk update [file_name] --force
pyxform_validator_update odk check
```

Example output:

```
> pyxform_validator_update enketo list

Installed release:

- Tag name = 1.0.3
- Tag URL = https://github.com/enketo/enketo-validate/releases/tag/1.0.3

Latest release:

- Tag name = 1.0.3
- Tag URL = https://github.com/enketo/enketo-validate/releases/tag/1.0.3

Files available:

- linux.zip
- macos.zip
- windows.zip
```

```
> pyxform_validator_update enketo update
usage: updater.py enketo update [-h] [--force] file_name
updater.py enketo update: error: the following arguments are required: file_name
```

```
> pyxform_validator_update enketo update linux.zip

Update failed!

The current version appears to be the latest.

To update anyway, use the '--force' flag.

Installed release:

- Tag name = 1.0.3
- Tag URL = https://github.com/enketo/enketo-validate/releases/tag/1.0.3
```

```
> pyxform_validator_update enketo update windows.zip --force

Update success!

Install check of the latest release succeeded.

Latest release:

- Tag name = 1.0.3
- Tag URL = https://github.com/enketo/enketo-validate/releases/tag/1.0.3
```

```
> pyxform_validator_update enketo check

Check success!

The installed release appears to work.

Installed release:

- Tag name = 1.0.3
- Tag URL = https://github.com/enketo/enketo-validate/releases/tag/1.0.3
```


lindsay-stevens:
Would someone mind reviewing / approving this for me?

pyxform/validators/util.py
```diff
@@ -63,3 +77,68 @@ def decode_stream(stream):
...
+    A logging handler capturing all (raw and formatted) logging output.
+    Largely the same as doing "from unittest.case import _CapturingHandler",
+    but copied here because it is a simple class, and the original has an
+    underscore prefix so it might change unexpectedly.
+
+    Usage:
+    Here's how to attach it to a logger, get the logs, and then detach it.
+
+    my_logger = logging.getLogger(name="logger_name")
+    capture_handler = _CapturingHandler(logger=my_logger)
```
ukanga: Should the underscore be there  `_CapturingHandler(logger=my_logger)`?

pyxform/validators/updater.py
```diff
@@ -0,0 +1,602 @@
...
+        zip_info = open_zip_file.infolist()
+        zip_jobs = []
+        for zip_item in zip_info:
+            for file_target in bin_paths:
+                if fnmatch.fnmatch(zip_item.filename, file_target[0]):
+                    file_out_path = os.path.join(out_path, file_target[1])
+                    zip_jobs.append((zip_item, file_out_path))
+        if len(bin_paths) != len(zip_jobs):
+            raise PyXFormError(
+                "Expected {e} zip job files, found: {c}"
```
ukanga: I seem to be encountering `Expected 3 zip job files, found: 6` when I run  `enketo update`.

pyxform/validators/util.py
ukanga: Should the underscore be here  `capture_handler = _CapturingHandler(logger=my_logger)`? I believe you had renamed the class to not use the underscore.

lognaturel:
I haven't had a chance to look at this yet but intend to in the next 24hrs.

pyxform/validators/updater.py
lindsay-stevens: Thanks - addressed in 8b7d2fa, somehow duplicate files were in the zip so added handling for that.

pyxform/validators/util.py
lindsay-stevens: Thanks - addressed in 8b7d2fa, the class was customised a fair bit so I updated the whole docstring.

lognaturel:
I'm getting a strange permission denied error when running the verification with the "spec" file after updating Enketo Validate for macos. All tests pass. Is this just me?
```
File "pyxform/validators/updater.py", line 499, in _install_check
    return enketo_validate.install_ok(bin_file_path=extracted)
  File "pyxform/pyxform/validators/enketo_validate/__init__.py", line 36, in install_ok
    path_to_xform=XFORM_SPEC_PATH, bin_file_path=bin_file_path)
  File "pyxform/pyxform/validators/enketo_validate/__init__.py", line 28, in _call_validator
    return run_popen_with_timeout([bin_file_path, path_to_xform], 100)
  File "pyxform/pyxform/validators/util.py", line 51, in run_popen_with_timeout
    startupinfo=startup_info)
```

For the ODK Validate updater, it looks like it's running through the Enketo Validate mapping form zip file to binary paths (_get_bin_paths). I get "Did not find a supported main binary for file". I think what should happen is that if there is a jar in `bin_new`, `bin_new` should replace `bin`. The Validate release jar usually has a version number in the file name. Probably the easiest way to accommodate that in `odk_validate/__init__.py` is to grab the first jar in the directory.

lognaturel:
This is a small thing but I was surprised to find that when I did a `list` I would get the contents of `latest.json`. The 30mins delay before requery would make a ton of sense if updates were checked for each time xls2xform were run but since listing is now manual only I wonder whether it's necessary?

lindsay-stevens:
@lognaturel thanks for checking, and uploading the release! I added some fixes in 1fe531f:

- Turns out `is_zipfile` thinks jar files are zip files so the install function uses the file extension as well
- For both ODK and Enketo, the new bin file is given the name of the bin file expected elsewhere e.g. when using xls2xform
- I'm not sure why you'd be getting a permission denied error. My guess is that the tests still had a lock open on the test XML file, since it's a test fixture file that gets re-written during tests. So I added a check that'll wait up to 5 seconds for any such lock to clear before the install_ok check proceeds.

The `latest.json` caching is done to avoid GitHub's API limit of 60 calls per hour per IP, if unauthenticated. I used unauth'd calls because it avoids having to manage or collect GitHub creds which could get complicated for those with 2FA. So the choice was between mitigating a couple of not-so-great outcomes:

1. Someone at a large institution (or busy business, or a busy dev) blocks themselves out for an hour if calls aren't cached
2. Someone does a `list` within 30 minutes of a new release and gets stale info

I could add a `--force` flag to `list` to get around point 2 if preferred. The other way to clear the 30 minute check it is to delete the `.last_check` file, saved under `validators/enketo_validate` or `validators/odk_validate`.

lognaturel:
Thanks for the fixes and explanations @lindsay-stevens! The ODK Validate update is now working great for me. 

I'm still getting the same permissions error for Enketo Validate. This is even more puzzling since it looks like the two variants are running the same check. @MartijnR does it work for you?

> Someone at a large institution (or busy business, or a busy dev) blocks themselves out for an hour if calls aren't cached

For this to happen they'd have to explicitly list and/or update a bunch of times, right? That doesn't seem super likely but it's not a big deal to leave things as-is.

lindsay-stevens:
> For this to happen they'd have to explicitly list and/or update a bunch of times, right? That doesn't seem super likely but it's not a big deal to leave things as-is.

Yep I think these scenarios would be uncommon. Happy to leave as is.

Hopefully Martijn can help, since I don't have a mac I can debug on. It looks like CircleCI can do mac builds for open source projects if asked nicely https://circleci.com/pricing/#faq-section-os-x but that could take a while to get right.

Does the same error happen if you try to validate:

- As part of a form conversion: `xls2xform --enketo_validate pyxform/tests/example_xls/xlsform_spec_test.xlsx`?
- Independently of pyxform: `pyxform/validators/enketo_validate/bin/validate pyxform/test/test_expected_output/xlsform_spec_test.xml`?

MartijnR:
I have the same permissions issue on a mac. I've confirmed that when downloading the enketo-validate binary downloaded from [github](https://github.com/enketo/enketo-validate/releases/tag/1.0.3) works as expected. 

After running the update to install enketo-validate I see the `bin_new` folder (no `bin`) with the enketo-validate files as expected.

When running `xls2xform --enketo_validate pyxform/tests/example_xls/xlsform_spec_test.xlsx`, I get `"Enketo validate dependency not found"` (since there is no `bin`).

When running `pyxform/validators/enketo_validate/bin_new/validate pyxform/test/test_expected_output/xlsform_spec_test.xml`, I get `"Permission denied"`

After `chmod a+x pyxform/validators/enketo_validate/bin_new/validate`, enketo-validate becomes functional by itself (bypassing pyxform).

After renaming `bin_new` to `bin`, everything works inside pyxform.

lognaturel:
Same experience as @MartijnR. For some reason I thought the issue was with the XML file but no, it's just that the executable doesn't get execute permissions. Once that's given everything works as expected.

lindsay-stevens:
Hopefully it's fixed in 764588e - the equivalent of `ug+x` is done on the new bin file (`ODK_Validate.jar` or `validate`). Could you please give it a go? Also if it should definitely be `a+x` then please let me know, I thought I'd be a tiny bit paranoid and exclude `o+x`.

To explain the `bin_new` vs. `bin` thing:

- the newly downloaded file goes into `bin_new`
- if the `install_ok` check passes (which it wasn't on mac) then the existing `bin` directory (if any) is removed and `bin` is renamed to `bin_new`.
- if the check fails then `bin_new` is left there to help debugging
   - if `bin_new` still exists on on a subsequent run then it is removed and replaced

lindsay-stevens:
Great! Is it OK by you @ukanga?

ukanga:
The install and commands work okay. However, something that could be a concern is that the default ODK_Validate.jar committed in this repository is not deployed. Which means, for deployments that use this as a library will need to run `update` command for any validation to work. I see having to make changes in deployment scripts and documentation to remember to do this. Is this intentional? I understand the issue where Enketo validate is concerned but for `ODK_Validate.jar` it does break backward compatibility unless the intention is to get out of `v0.9.x` to `1.0` for the pyxform release.

lognaturel:
@ukanga I believe `ODK_Validate.jar` has just moved to a `bin/` subfolder. I verified that things worked from a clean install and I think this is ok on that front.

ukanga:
Maybe it is just my environment or how I am approaching this, can you test with docker? With:
```Dockerfile
FROM python

RUN apt-get update \
  && apt-get install -y openjdk-7-jre-headless \
  && rm -rf /var/lib/apt/lists/*

RUN  echo "git+git://github.com/lindsay-stevens/pyxform.git@fancy-updater" > /tmp/requirements.txt \
  && pip install -r /tmp/requirements.txt

CMD xls2xform /usr/local/lib/python3.6/site-packages/pyxform/tests/example_xls/tutorial.xls /tmp/tutorial.xml && cat /tmp/tutorial.xml

```
I get:
```python
  File "/usr/local/bin/xls2xform", line 11, in <module>
    load_entry_point('pyxform==0.10.0', 'console_scripts', 'xls2xform')()
  File "/usr/local/lib/python3.6/site-packages/pyxform/xls2xform.py", line 137, in main_cli
    enketo=args.enketo_validate)
  File "/usr/local/lib/python3.6/site-packages/pyxform/xls2xform.py", line 27, in xls2xform_convert
    warnings=warnings, enketo=enketo)
  File "/usr/local/lib/python3.6/site-packages/pyxform/survey.py", line 691, in print_xform_to_file
    warnings.extend(odk_validate.check_xform(path))
  File "/usr/local/lib/python3.6/site-packages/pyxform/validators/odk_validate/__init__.py", line 76, in check_xform
    'ODK Validate Errors:\n' + ErrorCleaner.odk_validate(stderr))
Error: Unable to access jarfile /usr/local/lib/python3.6/site-packages/pyxform/validators/odk_validate/bin/ODK_Validate.jar
```
Running an image that pip installs the pip version works as expected, it will print the converted XForm XML.
```Dockerfile
FROM python

RUN apt-get update \
  && apt-get install -y openjdk-7-jre-headless \
  && rm -rf /var/lib/apt/lists/*

RUN pip install pyxform

CMD xls2xform /usr/local/lib/python3.6/site-packages/pyxform/tests/example_xls/tutorial.xls /tmp/tutorial.xml && cat /tmp/tutorial.xml
```

lindsay-stevens:
Yep I'd moved the jarfile down one folder and updated the path that points to it.

I tested it with docker and got the same error result. Embarrassingly, I made the same kind of mistake with my last PR - forgetting to update the package data info in `setup.py`. I guess we'd all (including travis/appveyor) been testing with `git clone` rather than `pip install`. I noticed that `strings.ini` file was missing as well, so presumably running `nosetests` after pip install had been broken for a while.

Anyway, I tried again after f8b7d0a and the docker image prints an XForm on run, as does `nosetests`. Thanks for checking this @ukanga - could you please verify it's working for you?

MartijnR:
Great. Thanks @lindsay-stevens, @lognaturel, @ukanga

lindsay-stevens:
Awesome, thanks all.

PR #178, opened 2018-02-28
Support auto_send and auto_delete settings columns

### LINKED ISSUE #126: Add setting for auto-send and auto-delete
`orx:auto-send` and `orx:auto-delete` were added as attributes to the `submission` element in opendatakit/xforms-spec#100. Both can be set to either `"true"` or `"false`" and can combine. For example:
`<submission orx:auto-delete="true" orx:auto-send="false">`

It seems that to fit existing XLSForms idioms, each would get its own settings column.

lognaturel:
Closes #126


lognaturel:
Thanks, @ukanga!

PR #179, opened 2018-03-06
check python version to solve unichr not defined problem

### LINKED ISSUE #161: Python 3,unichr not defined in util.py error

shivareddyiirs:
Fix #161 
If unicode does not throw exception , then check the python version so that this problem can be solved.

### DIFF
--- a/pyxform/utils.py
+++ b/pyxform/utils.py
@@ -14,9 +14,9 @@ except NameError:
     unichr = chr
 else:
     if sys.version_info >(3,5,0):
-    unicode = str
-    basestring = str
-    unichr = chr
+        unicode = str
+        basestring = str
+        unichr = chr
     else:
         unicode = unicode
         basestring = basestring
--- a/pyxform/utils.py
+++ b/pyxform/utils.py
@@ -5,6 +5,11 @@ import json
 import copy
 import unicodecsv as csv
 import xlrd
+import sys
+if sys.version_info >(3,5,0):
+    unicode = str
+    basestring = str
+    unichr = chr
 
 try:
     unicode("str")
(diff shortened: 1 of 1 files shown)


shivareddyiirs:
@ukanga 
Thanks for your comment. if this pull request is merged, it would be great help for  develpers using pyxform in python 3 but having similar issue #161.

ukanga:
I think QGIS 3 might be implementing the `unicode()` function in python 3 or it gets injected somehow. On python 3 every version I have tested, if `unicode()` results in the `NameError` exception, so will the `unichr()` function. This may be the reason, despite us testing in python 2.7 to 3.6 on windows we still don't get this issue.
```python

try:
    unicode("str")
except NameError:
    # Python 3 only should go through here
    unicode = str
    basestring = str
    unichr = chr
else:
    # python 2 only should go through here
    unicode = unicode
    basestring = basestring
    # if for whatever reason python 3 is here 
    # it means unichr was not give the same treatment as unicode and basestring above
    # probably do another try except
    try:
        unichr = unichr
    except NameError:
        unichr = chr
```

Could you validate that this only happens when using QGIS?

I am going to try and replicate the #161 exception on my end once I get a hold of a windows VM and install qgis on it.

shivareddyiirs:
Yes, it only happen in QGIS.
Your idea of using another try except also  seems good and simple.

Thanks & Regards
Shiva Reddy K.
Scientist/Engineer 'SC'
Indian Institute of Remote Sensing,
Indian Space Research Organisation
Department of Space
4-Kalidas Road
Dehradun
mobile: 8791806093

Issue #174, opened 2018-02-06
Deal with duplicate column names with different cases


moshthepitt:
pyxform seems to allow the creating of forms with the same column/field name but with different cases.

e,g. you could have two columns with the names:

- Sports
- sports

This is potentially a problem because data exports to tools like SPSS may fail.

lognaturel:
Preventing that sounds like a very good idea to me!

PR #184, opened 2018-03-19
Disallow Duplicate Column Names

moshthepitt:
Ensure that duplicate column names fail validation, even if they are in different cases.

Fix: #174


pyxform/tests_v1/test_sheet_columns.py
```diff
@@ -87,7 +87,15 @@ def test_duplicate_columns_diff_cases(self):
 
 
 class InvalidChoiceSheetColumnsTests(PyxformTestCase):
+    """
+    Invalid choice cheet column tests
```
lognaturel: cheet -> sheet

PR #185, opened 2018-03-28
Warning for languages that don't have valid IANA subtag

### LINKED ISSUE #143: Show warning when language doesn't include language code
In XLSForm/xlsform.github.io#73, @MartijnR describes an agreed-upon format for describing languages in a machine-interpretable way -- `<free text> (<language code>)`.

Show a warning when that format isn't respected. I see two options: just seeing whether the string ends in something like `\([a-zA-Z-]{2,}\)` or actually validate against [IANA subtags](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry). I think it would be nice to do a full validation but perhaps that's overreaching.

The warning could say something like `"The language specified doesn't include a machine-readable code as recommended by <link to XLSForm section about the suggested format>."`

lognaturel:
Closes #143 

I don't think there's a straightforward way to check for warnings in the `tests_v1` setup so I only verified this manually - with no languages specified, with multiple valid languages specified, with multiple languages specified one of which has no code, with multiple languages specified one of which has an invalid code.


lognaturel:
Added some tests and fixed the default lang issue. 

@MartijnR would be great if you could take a quick glance at the tests and make sure this does what you expect and that you're ok with the way the warning is framed.

MartijnR:
Looks great me. 👍 

Thanks for adding this. Really cool to get forms out there where we can sync UI and form languages.

Issue #38, opened 2015-06-18
Itemset from previous answers - select one from ${node}


MartijnR:
Triggered by [this question](https://groups.google.com/forum/#!topic/formhub-users/K6hcV3Bh77Y)

It exposes the so-far-untapped XForm capability to create a choice list from previous answers (in a repeat).

It could have a wider implication by avoiding the more difficult indexed-repeat() or a combination of concat() and selected-at(). 

After discussion with @dorey we came up with the following initial syntax suggestion:

| type | name | label | choice_filter |
| --- | --- | --- | --- |
| begin repeat | rep |  |  |
| text | name | Enter name |  |
| end repeat |  |  |  |
| **select one from ${name}** | choice | Choose | string-length(./name) > 0 |

``` xml
<select1 ref="/data/choice">
   <label>Choose name</label>
   <itemset nodeset="/data/rep[string-length(./name) > 0]">
       <label ref="name"/>
       <value ref="name"/>
   </itemset>
</select1>
```

The `ref` attribute is the nodename of ${name} and the `nodeset` attribute is the parent of  ${name} with an optional choice filter.

MartijnR:
_select multiple_ support might be difficult because we cannot allow spaces in the answers...

dorey:
I like the idea of adding the ability to reference previous repeats. After thinking about the syntax a little more, I'm wondering if we could come up with a better way to specify it in the xlsform. Main reasons--

* <code>select one from ${name}</code> could require new regex changes for parsing the type field. Even if we went with <code>select_one ${name}</code> there's a high probability that parsers in pyxform and various xlsform form-builders and parsers (<img src="http://vignette2.wikia.nocookie.net/banjokazooie/images/5/51/Whistling.gif/revision/latest?cb=20090408112203" />) would need their regex fixed. And though we could do it on individual cases, it'd be cleaner if we could settle on a way of extending functionality that doesn't involve changing the regex.
* I vaguely remember people wanting to move away from <code>{ "type": "select_one listname" }</code> towards something like <code>{ "type": "select_one", "list_name": "listname" }</code>. Which would make parsing and validation of the form easier and more precise. (Note: I may have imagined this conversation)

How about:
<table>
<tr>
<th>type
</th>
</tr>
<tr>
<td>select_one ${prev_name}
</td>
</tr>
</table>
* still might need regex changes

or

<table>
<tr>
<th>type
</th>
<th>list name
</th>
</tr>
<tr>
<td>select_one
</td>
[...]

PR #187, opened 2018-04-03
Use relative references for repeats only.

### LINKED ISSUE #4: Add relative referencing
The latest Javarosa version for ODK Collect now has 

`current()/../../ref` syntax. To do relative referencing of fields in an XLSForm. One could theoretically replace all the ${fieldname} expansions with this syntax and then we would have reliable referencing of fields within the current repeat group.  This is not documented anywhere (yet).

ukanga:
This introduces relative referencing for `${variables}` in repeat sections.

Not sure if my interpretation is accurate, I found a super long old thread in [ODK forum](https://forum.opendatakit.org/t/current-in-xpath-predicate/4982/45), not sure if it is relevant still.

I picked the option of dealing with https://opendatakit.github.io/xforms-spec/#a-big-deviation-with-xforms hoping it does address it.

Fix #4


MartijnR:
OMG, this is exciting! 👍 Maybe we can take out that deviation from the spec some time.

Will check it out!

lognaturel:
As the thread @ukanga links to notes, Javarosa doesn't interpret `current()` correctly in choice filters. This is the variant of @MartijnR's form above that has the expected behavior in Javarosa (note the wrong `name =  current()/crop` and `name = current()/../crop`):
```xml
<select1 ref="/repeat-relative-current-1/rep/sel_a">
  <label> Verify <output value=" current()/../crop "/> </label><itemset nodeset="instance('crop_list')/root/item[name =  current()/crop ]">
    <value ref="name"/>
    <label ref="jr:itext(itextId)"/>
  </itemset>
</select1>
<group ref="/repeat-relative-current-1/rep/group">
  <select1 ref="/repeat-relative-current-1/rep/group/sel_b">
    <label> Verify <output value=" current()/../../crop "/> </label><itemset nodeset="instance('crop_list')/root/item[name = current()/../crop]">
      <value ref="name"/>
      <label ref="jr:itext(itextId)"/>
    </itemset>
  </select1>
</group>
```

Perhaps there can be some flags or model version information added in somehow but I don't think it can be the default behavior for pyxform right now, at least not for choice filters. The problem is that even if we patch JR today (which doesn't sound trivial), all users running older versions of Collect (and there are many), will be unable to use new forms with choice filters. That's what the very long thread was about.

MartijnR:
Ah, I figured that **crazy serious bug** was fixed long time ago. This means that today, I can still manually create a `choice_filter` with `current()` in XLSForm or ODK Build and have it work in Enketo but not in ODK Collect. That should be fixed in any case, right (regardless of this PR)?  I understand you need a way to determine whether the XForm may be using a bug-compliant usage of current(). Tagging XForms for increased correctness is one way of doing that. However, I wonder if doing a simple analysis to detect whether the XForm relies on the bug as follows would be sufficient to switch to `current-bug-compliant` mode in JavaRosa: 

If expression matches `/current\(\)\/[^.]+/` switch to `current-bug-compliant` mode.

This relies on the syntax where `current()/` is not followed by a `.` (or `..`). It may catch close to 100% of all the (very few) handcoded forms that use current() currently. I would much prefer such a solution to tagging XForms for (likely) not relying on bugs.

MartijnR:
Test 1:

| survey |                      |       |        |                |
| ------ | -------------------- | ----- | ------ | -------------- |
|        | type                 | name  | label  | choice_filter  |
|        | begin repeat         | rep   |        |                |
|        | select_one crop_list | crop  | Select |                |
|        | begin group          | group |        |                |
|        | select_one crop_list | sel_b | Verify | name = ${crop} |
|        | end group            |       |        |                |
|        | end repeat           |       |        |                |

expected output: `[ name = current()/../../crop ]`

Test 2: 

| survey |                      |       |        |                |
| ------ | -------------------- | ----- | ------ | -------------- |
|        | type                 | name  | label  | calculate      |
|        | begin repeat         | rep   |        |                |
|        | select_one crop_list | crop  | Select |                |
|        | text                 | a     |        | name = ${crop} |
|        | begin group          | group |        |                |
|        | text                 | b     | Verify | name = ${crop} |
|        | end group            |       |        |                |
|        | end repeat           |       |        |                |

expected output for a: `calculate="../crop"`
expected output for b: `calculate="../../crop"`

lognaturel:
These tests look great. Would be good to get to a high-level game plan also. Can we do relative paths everywhere other than choice filters for now? Do we wait on the overall change?

ukanga:
Thank you guys for the comments and tests, I will add them in the next iteration on this PR.

ukanga:
Example of XForm XML with relative paths as generated by this PR. The first two are the examples @MartijnR provided [above](https://github.com/XLSForm/pyxform/pull/187#issuecomment-378755627).
- [Choice filter with relative paths](https://gist.githubusercontent.com/ukanga/64fa7684bcdbc740a0346c13264b8285/raw/62837ae7f920452225b0abdfe4654d49ba3240d5/choice_filter_w_relative_paths.xml)
-  [Calculate with relative paths](https://gist.githubusercontent.com/ukanga/18d0457b9e26316fa8e64218075ea607/raw/09f1e3ee29bb5b2d77e9b2311f8b19a6c580f9ed/calculate_w_relative_paths.xml)
- [For loop with relative paths](https://gist.githubusercontent.com/ukanga/da9c0f8d378470e12b69e4a7b2880feb/raw/8c66cebb5cab528d4a39ee16191034c8ac543559/for_loop_w_relative_paths.xml)
- [A more complex repeat structure with relative paths XML](https://gist.githubusercontent.com/ukanga/6f175ef659b82bbbb77383df60457209/raw/0d8c1825e51b4005286a5936b51a7f5e4a817698/relative_paths.xml).

lognaturel:
> If you have the chance to test the above XML in ODK Collect or Enketo, let me know what's not working as expected.

Everything works as expected in Collect other than the choice filters. 🎉 I think we will have a fix for JavaRosa soon. I /think/ this is actually ok to merge though I'd like a sanity check on my thinking. The JavaRosa problem is that `current()` doesn't refer to the right thing in selects. As far as I can tell, this only affects two things: choice filters and the itemset nodeset. It's not very likely users are using relative paths to set the itemset nodeset. One would use that in a scenario like collecting household info in which the enumerator is asked to select the owner of each pet from a dynamically generated list of children:

```
repeat people
  repeat children
  repeat pets
    select itemset nodeset=current()../../children
```

It's cool and powerful but there's no XLSForm support for itemsets from arbitrary nodes (https://github.com/XLSForm/pyxform/issues/38) and it's pretty esoteric. I've only shown users how to do it with absolute paths but have never seen it come up with relative paths before.

The other case is choice filters. XLSForms with choice filters in repeats fail today because an absolute path doesn't make sense there. Currently in Collect we get an error about needing `indexed-repeat`. With this change we'd get no error but also no choices shown. Maybe a little harder for users to troubleshoot but nothing too bad. And as soon as they got patched Collect it would Just Work™️.

The change we're going to make in JR would only affect users who are trying to use `current()` already. This PR doesn't affect them (maybe a good test to add @ukanga? Demonstrate that if a path is already relative this doesn't change anything?).

If someone else who uses Collect and understands the JR issue can run through this reasoning and validate it, I'd be comfortable merging.

lognaturel:
> You could decide to strip the current()/ part if the expression is not inside a predicate, e.g. calculate="../crop" instead of calculate="current()/../crop"

I would be in favor of this. `current()` is a bit of a strange bird so the more it can be avoided the better.

MartijnR:
Thanks. I can confirm the forms work in Enketo too.  (after correcting calculation in calculate_w_relative_paths.xml to  `calculate="current()/../crop" ` calculate="current()/../../crop ", but that's not a pyxform issue.)

I agree best to avoid using current() when it's not required.

ggalmazor:
Hi! Just wanted to drop by and say that @lognaturel's reasoning looks good to me. I also agree that I would prefer having `../` instead of `current()` wherever possible.

ukanga:
@MartijnR @lognaturel not sure what was pending on this but I see most of what was on your comments was addressed. Any blocker to getting this merged?

pyxform/tests_v1/test_repeat.py
```diff
@@ -0,0 +1,157 @@
...
+                |         |                      |       |        |                |
+                | choices |                      |       |        |                |
+                |         | list name            | name  | label  |                |
+                |         | crop_list            | maize | Maize  |                |
+                |         | crop_list            | beans | Beans  |                |
+                |         | crop_list            | kale  | Kale   |                |
+            """,  # noqa pylint: disable=line-too-long
+            xml__contains=[
+                """<itemset nodeset="instance('crop_list')/root/item[name =  ../crop ]">""",  # noqa pylint: disable=line-too-long
+                """<itemset nodeset="instance('crop_list')/root/item[name =  ../../crop ]">""",  # noqa pylint: disable=line-too-long
```
MartijnR: We're using a predicate here, and we are changing the context to `instance('crop_list')`, so we would need current() here: `[name = current()/../crop]` and `[name = current()/../../crop]`

MartijnR:
Thanks! Once current() is back in choice-filters, I'll run the test forms in Enketo for a final/proper check.

ukanga:
@MartijnR 3c31570829e08a237656087966347bdcd21f8abd should address it.

yanokwa:
I got some XForms consulting time with @lognaturel and this PR looks like what she wanted. That is pyxform generates relative paths with `.` and `..` everywhere except in predicates like choice filters. In those cases, `current()` should be used. My only minor concern is that are there any other places besides choice filters where we could use predicates?

MartijnR:
I confirm that this works great! Thanks!

I noticed that the calculations in Test 2 that I provided (`name = ${crop}`) make no sense (will always result in `false`), but that doesn't matter as the test still serves its purpose and the expression is valid.

PR #195, opened 2018-04-18
Flags are not set due to misnamed variable

yanokwa:
I noticed at https://github.com/opendatakit/xlsform-offline/issues/10 that I was getting popups. I think I've found the issue!

@ukanga If it's not too much trouble, I'd love a hotfix release.


lindsay-stevens:
Hi Yaw,

Looks good to me. I'm guessing this is my bad, so sorry for that.

Thumbs up for #196 (pip fix) as well.


Cheers,
Lindsay

PR #198, opened 2018-05-10
Set mediatype for file to application/*

### LINKED ISSUE #194: Use application/* or add gpx and xml media types to file type
The file type currently results in the following XForms output:
```
<upload mediatype="text/plain,application/pdf,application/vnd.ms-excel,application/msword,text/richtext,application/vnd.openxmlformats-officedocument.wordprocessingml.document,application/vnd.openxmlformats-officedocument.spreadsheetml.sheet,application/zip,application/x-zip,application/x-zip-compressed" ref="...">
```

I see that `application/*` was avoided with the original implementation because of some Chrome issue? https://github.com/XLSForm/pyxform/issues/33#issuecomment-100297975 @MartijnR or @ukanga, perhaps one of you remembers more.

I think it would be ideal if `application/*` could now be used as the default for `file`. If not, then can we add `application/gpx+xml` and `application/xml`?

We could also use the `parameters` column to give the form author more control.

lognaturel:
Closes #194 

One reason to have a list was to avoid binary types. However, this can be bypassed so protection is weak. Users can use ~body::mediatype~ `body::accept` to set explicit allowed types.

This replaces the file type test with a more modern variant.


MartijnR:
Code looks good to me. Thanks.

PR #199, opened 2018-05-10
Add support for randomizing selects

### LINKED ISSUE #183: Randomize order of items in select_one and select_multiple
The [`randomize(nodeset n, number? seed)`](https://opendatakit.github.io/xforms-spec/#fn:randomize) function was added to the XForms spec recently. One powerful use is to randomize the items in instances that describe choices for selects either in a reproducible order (with a seed) or not (no seed):

```xml
<select1 ref="/form/state">
    <itemset nodeset="randomize(instance('choices')/states/state)">
        <label ref="name"/>
        <value ref="value"/>
    </itemset>
</select1>
```

I propose using the `parameters` column to enable and configure this feature with two new parameters for `select_one` and `select_multiple`:
* `randomize`: `true|false` (or `yes|no`) - if not specified, defaults to false. If explicitly set to true, the choices for the select will be generated in a secondary instance whether or not a `choice_filter` is specified. The `randomize` function will be called on the `nodeset` specified in the `itemset`.

* `seed`: `number` - if not specified, no seed argument is passed to the `randomize` function. If specified, it is passed in as the second argument to the `randomize` function. If `seed` is specified, the `randomize` parameter is assumed to be true and ignored.

Examples:

|         type          |    name     |       label       |             parameters              |          calculation           |
|-----------------------|-------------|-------------------|-------------------------------------|--------------------------------|
| select_multiple fruit | fruits      | Select some fruit | randomize=true                      |                                |
| calculate             | random_seed |                   |                                     | once(decimal-date-time(now())) |
| select_one fruit      | fruit       | Select a fruit    | randomize=true, seed=${random_seed} |                                |

lognaturel:
Closes #183 

Would be good to first get a sanity check that tests are exhaustive.

I'm not happy that there's so much repeated code between pulling out parameters for this feature and doing it for range. But I thought it might be a bit easier to pick the right functions to introduce once we have a third feature that leverages the `parameters` column. Can take another pass through it if anyone feels too offended, though.

We will need to update Validate before this correctly validates.

### DIFF
--- a/pyxform/tests_v1/test_randomize_itemsets.py
+++ b/pyxform/tests_v1/test_randomize_itemsets.py
@@ -148,9 +148,28 @@ class RandomizeItemsetsTest(PyxformTestCase):
             name="data",
             errored="true",
             md="""
+            | survey |                    |         |       |                             |
+            |        | type               | name    | label | parameters                  |
+            |        | select_one choices | select  | Select| randomize=true, seed=ukanga |
+            | choices|                    |         |       |                             |
+            |        | list_name          | name    | label |                             |
+            |        | choices            | a       | opt_a |                             |
+            |        | choices            | b       | opt_b |                             |             
+
+            """,
+            error__contains=[
+                "seed value must be a number."
+                ]
+        )
+
+    def test_randomized_select_one_seed_without_randomize(self):
+        self.assertPyxformXform(
+            name="data",
+            errored="true",
+            md="""
             | survey |                    |         |       |                  |
             |        | type               | name    | label | parameters       |
-            |        | select_one choices | select  | Select| seed=ukanga      |
+            |        | select_one choices | select  | Select| seed=42          |
             | choices|                    |         |       |                  |
             |        | list_name          | name    | label |                  |
             |        | choices            | a       | opt_a |                  |
@@ -158,7 +177,8 @@ class RandomizeItemsetsTest(PyxformTestCase):
 
             """,
             error__contains=[
-                "seed value must be a number."
+                "Parameters must include randomize=true to use a seed."
                 ]
         )
+
 
--- a/pyxform/tests_v1/test_randomize_itemsets.py
+++ b/pyxform/tests_v1/test_randomize_itemsets.py
@@ -38,7 +38,7 @@ class RandomizeItemsetsTest(PyxformTestCase):
         )
 
     def test_randomized_seeded_filtered_select_one(self):
-         self.assertPyxformXform(
+        self.assertPyxformXform(
             name="data",
             md="""
             | survey |                    |         |       |                         |               |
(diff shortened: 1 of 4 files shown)


pyxform/tests_v1/test_randomize_itemsets.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+
+    def test_randomized_select_one_bad_seed(self):
+        self.assertPyxformXform(
+            name="data",
+            errored="true",
+            md="""
+            | survey |                    |         |       |                  |
+            |        | type               | name    | label | parameters       |
+            |        | select_one choices | select  | Select| seed=ukanga      |
```
ukanga: Does seed imply an implicit randomize? Does it mean that if we had `seed=42` then we should have a `randomize()` output? Or do we need to raise the complaint that `randomize` has not been set?

pyxform/tests_v1/test_randomize_itemsets.py
```diff
@@ -0,0 +1,164 @@
...
+            |        | choices            | b       | opt_b |                         |             
+
+            """,
+            xml__contains=[
+                "<itemset nodeset=\"randomize(instance(\'choices\')/root/item, 42)\">"
+                ]
+        )
+
+    def test_randomized_seeded_filtered_select_one(self):
+         self.assertPyxformXform(
```
ukanga: When testing how a variable `${a}` would result in I noticed that there are indent issues on this function, you could look into that.

pyxform/tests_v1/test_randomize_itemsets.py
lognaturel: Great question. Right now `seed` is quietly ignored if randomize isn't explicitly set to true. I thought this was ok initially but you're right it would be better to warn. I'm also ok with the presence of a seed implying randomize is true. Anyone have strong feelings here?

KeynesYouDigIt: explicit is better than implicit, but seed randomization isnt something I want to think about very often as a user or dev. Seed should have a default when random is true and ignored when its not. Being pretty new to XML though im not sure if thats easy. But if it is thats how id set it up.

lognaturel: @KeynesYouDigIt As you can see from the tests, the seed is optional. It exists to allow form designers to recreate the order shown to enumerators if that's important in their statistical analysis. You can read more about the function at https://opendatakit.github.io/xforms-spec/#fn:randomize.

MartijnR: I have a tiny preference for explicit I think (i.e. seed without randomize shows warning).

lognaturel: Thanks everyone! Will fix that and indentation error ASAP.

lognaturel:
I think I caught everything, please take a look!

PR #201, opened 2018-05-19
add python versions support policy to readme

KeynesYouDigIt:


README.rst
```diff
@@ -29,6 +29,8 @@ The ``xls2xform`` command can then be used::
 
     xls2xform path_to_XLSForm output_path
 
+The platform is tested and running in Python 2.7 and Python 3
```
lognaturel: I'm having trouble understanding this sentence! Maybe something like "`pyxform` can be run with either Python 2 or Python 3. Continuous integration runs tests on both Python generations to ensure continued compatibility."

PR #204, opened 2018-05-24
Support guidance hints

### LINKED ISSUE #142: Add additional guidance hints
This feature has recently been added to the [ODK XForms Spec](https://opendatakit.github.io/xforms-spec/#languages). Rationale for the feature is described [here](https://github.com/opendatakit/xforms-spec/issues/132).

To add this feature to pyxform, could we do something like this perhaps?

input:

| type | name | label | hint | hint_print | relevant
|-----|-------|------|------|---|-----|
| int | a | label | normal hint | This question is skipped if the answer to ... is less than or equal to 5. | ${num} > 5

output:
```xml
<text id="/data/a:hint">
    <value>normal hint</value>
    <value form="print">This ... or equal to 5.</value>
</text>
```

lognaturel:
Closes #142

I considered calling `xml_guidance_hint` directly from `question` but ultimately found it cleaner to include all hint variants in `xml_label_and_hints`.

### DIFF
--- a/pyxform/constants.py
+++ b/pyxform/constants.py
@@ -28,6 +28,7 @@ AUTO_DELETE = u"auto_delete"
 DEFAULT_LANGUAGE = u"default_language"
 LABEL = u"label"
 HINT = u"hint"
+GUIDANCE_HINT = u"guidance_hint"
 STYLE = u"style"
 ATTRIBUTE = u"attribute"
 
--- a/pyxform/json_form_schema.json
+++ b/pyxform/json_form_schema.json
@@ -127,6 +127,16 @@
 							"description" : "A key value pair where the key is a language, and the value is the content in that language."
 						}
 					},
+					"guidance_hint" :
+					{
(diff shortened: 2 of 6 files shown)


MartijnR:
I'm confused about the XForm syntax using guidance hint with itext. Please give me a day or so to get some time to review. I had something different when implementing form="print" previously for a client: https://enketo-aggregate.appspot.com/www/formXml?readable=true&formId=printhint2 but am no longer sure if that made sense.

lognaturel:
Sounds good, @martijnr! I matched hint and label exactly. I don’t have access to the server you linked to but would be interested in seeing the XML you had generated.

MartijnR:
Ah sorry, I have updated it now.

lognaturel:
No, it's me. What I did doesn't at all match what we'd agreed on at https://opendatakit.github.io/xforms-spec/#languages, I was going too quickly and forgot to review the spec.

PR #205, opened 2018-05-25
Add max-pixels parameter for images

### LINKED ISSUE #137: Add orx:max-pixels attribute for images
As described in opendatakit/xforms-spec#79.

The output should be of the structure
`<bind ref="/path/to/node" orx:max-pixels="1024" type="binary"/>`

I'm thinking something like 
```
| survey |        |          |       |                     |
|        | type   |   name   | label | parameters          |
|        | image  | my_image |...    | max-pixels=1024  |
```

lognaturel:
Closes #137

Let's hope I didn't do anything too crazy this time!

a1e24ed pulls out code common to types that can have values in the parameters column as we had discussed in a prior PR.

Instead of passing on the `max-pixels` value in `parameters` I went straight to adding it to the `bind` dictionary.

One thing that confuses me is that in the happy case test the order of the bind attributes isn't what I expected: https://github.com/XLSForm/pyxform/compare/master...lognaturel:pyxform-137?expand=1#diff-86152448d7cce73f07b4cd294f930633R14. In fact, when I set debug=True, `orx:max-pixel` comes before `type` to give `<bind nodeset="/data/my_image" orx:max-pixels="640" type="binary"/>`. I don't think this is a symptom of a deeper problem but would be good to get a double check.

The very astute will notice an unrelated change at https://github.com/XLSForm/pyxform/compare/master...lognaturel:pyxform-137?expand=1#diff-b29aeb282b118614b8333cdfc19f700dL859. I noticed that it wasn't used so I snuck that in as part of the extraction.

### DIFF
--- a/pyxform/tests_v1/test_max_pixels.py
+++ b/pyxform/tests_v1/test_max_pixels.py
@@ -1,0 +1,44 @@
+from pyxform.tests_v1.pyxform_test_case import PyxformTestCase
+
+class MaxPixelsTest(PyxformTestCase):
+    def test_integer_max_pixels(self):
+        self.assertPyxformXform(
+            name="data",
+            md="""
+            | survey |        |          |       |                |
+            |        | type   | name     | label | parameters     |
+            |        | image  | my_image | Image | max-pixels=640 |
+            """,
+            xml__contains=[
+                'xmlns:orx="http://openrosa.org/xforms"',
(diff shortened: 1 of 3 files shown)


MartijnR:
Thanks! I looked at the new tests (only) in order to not overstep my boundaries. They look great to me.

PR #209, opened 2018-06-14
Add support for rank questions

### LINKED ISSUE #202: add type=rank
Very very similar implementation to `select_one` (and select_multiple). I think the only difference is  `<odk:rank>` instead of `<select1>` in the body.

type | name | 
-----|-------|
rank mylist | a         |

output:

```xml
<!-- primary instance -->
<data>
     <a/>
</data>
<!-- body -->
<orx:rank ref="/data/a">
     <label>label</label>
     <hint/>
     <item>
           <label>A</label>
           <value>a</value>
     </item>
     <item>
           <label>B</label>
           <value>b</value>
     </item>
<orx:rank>
```
| type | name | choice_filter |
|------|------|--------------|
| rank something | b | color='blue' |

```xml
<!-- primary instance -->
<data>
     <b/>
</data>
<!-- body -->
<orx:rank ref="/data/b">
     <label>label</label>
     <hint/>
     <itemset nodeset="instance('something')/root/item[color='blue']">
           <label ref="./label" />
           <value ref="./value" />
     </itemset>
<orx:rank>
```

See also: https://github.com/opendatakit/xforms-spec/pull/180

lognaturel:
Closes #202 

I tested a simple case, one with a choice filter and one with translations. As always, I appreciate feedback on whether this seems like sufficient verification.

### DIFF
--- a/pyxform/aliases.py
+++ b/pyxform/aliases.py
@@ -28,6 +28,7 @@ select = {
     u"select_multiple_from_file": constants.SELECT_ALL_THAT_APPLY,
     u"select one from file": constants.SELECT_ONE,
     u"select multiple from file": constants.SELECT_ALL_THAT_APPLY,
+    u"rank": constants.RANK,
 }
 cascading = {
     u'cascading select': constants.CASCADING_SELECT,
--- a/pyxform/builder.py
+++ b/pyxform/builder.py
@@ -45,7 +45,7 @@ class SurveyElementBuilder(object):
         u"trigger": TriggerQuestion,
         u"select": MultipleChoiceQuestion,
         u"select1": MultipleChoiceQuestion,
-        u"rank": MultipleChoiceQuestion,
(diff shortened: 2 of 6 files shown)


MartijnR:
This is exciting. Hope we can publish a new version soon after merging.

lognaturel:
Whoops, thanks, @MartijnR! I looked at other namespaced tags/attributes and it looks like the namespacing is just done by prepending a namespace tag at the beginning of the tag/attribute name so that's what I did. If there's a preferred approach let me know.

MartijnR:
Yes, I think that should work fine (as long as the xmlns:odk namespace declaration is already part of the XForm output). 👍

ukanga:
@MartijnR are the changes you requested to your satisfaction? You may need to approve this one.

MartijnR:
Looks perfect. Thanks!

PR #210, opened 2018-07-03
[WIP] Always output a first instance for repeats

lognaturel:
- [ ] remove count test? https://github.com/XLSForm/pyxform/pull/210#discussion_r199865425
- [ ] figure out what the nested case should actually look like https://github.com/XLSForm/pyxform/pull/210#discussion_r199863454
- [ ] fix tests that now fail 

I'm a little unclear on what should happen with nested repeats. The test at `test_nested_repeat_first_instance_defaults` shows what I expect to work though it is a little odd. It doesn't work in Collect and I haven't dug deeper to figure out why yet. @MartijnR, is Enketo happy with that?

(<i>Sidenote: is `instance` the most overloaded word ever or what?!</i>)

### DIFF
--- a/pyxform/tests_v1/test_repeats.py
+++ b/pyxform/tests_v1/test_repeats.py
@@ -1,0 +1,119 @@
+from pyxform.tests_v1.pyxform_test_case import PyxformTestCase
+
+# Verify that a repeat results in both a single repeat instance and a
+# repeat template. This means clients always show one repeat and
+# can optionally have a template.
+class RepeatTests(PyxformTestCase):
+    def test_repeat_first_instance(self):
+        self.assertPyxformXform(
+            md="""
+            | Survey |              |         |               |
+            |        | type         | name    | label         |
+            |        | begin repeat | repeat  | My Repeat     |
+            |        | integer      | age     | the age       |
+            |        | end repeat   | repeat  |               |
+            """,
+            xml__contains=[
+                """<repeat>
+            <age/>
+          </repeat>
+          <repeat jr:template="">
+            <age/>
+          </repeat>
+          """],
+        )
+
+    def test_repeat_first_instance_default(self):
+        self.assertPyxformXform(
+            md="""
+            | Survey |              |         |               |         |
+            |        | type         | name    | label         | default |
+            |        | begin repeat | repeat  | My Repeat     |         |
+            |        | integer      | age     | the age       | 23      |
+            |        | end repeat   | repeat  |               |         |
+            """,
+            xml__contains=[
+                """<repeat>
+            <age>23</age>
+          </repeat>
+          <repeat jr:template="">
+            <age>23</age>
+          </repeat>
+          """],
+        )
+
+    def test_repeat_first_instance_default_count(self):
+        self.assertPyxformXform(
+            debug=True,
+            md="""
+            | Survey |              |         |               |         |              |
+            |        | type         | name    | label         | default | repeat_count |
+            |        | begin repeat | repeat  | My Repeat     |         | 3            |
+            |        | integer      | age     | the age       | 23      |              |
+            |        | end repeat   | repeat  |               |         |              |
+            """,
+            xml__contains=[
+                """<repeat>
+            <age>23</age>
+          </repeat>
+          <repeat </age>
+          </repeat>
+          """],
+        )
+
+    def test_repeat_first_instance_group_default(self):
+        self.assertPyxformXform(
+            md="""
+            | Survey |              |         |               |         |
+            |        | type         | name    | label         | default |
+            |        | begin repeat | repeat  | My Repeat     |         |
+            |        | begin group  | group   | My group      |         |
+            |        | integer      | age     | the age       | 23      |
+            |        | end group    | group   |               |         |
+            |        | end repeat   | repeat  |               |         |
+            """,
+            xml__contains=[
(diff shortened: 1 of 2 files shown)


pyxform/tests_v1/test_repeats.py
```diff
@@ -0,0 +1,119 @@
...
+            |        | integer      | age     | the age       | 23      |
+            |        | end repeat   | repeat2 |               |         |
+            |        | end repeat   | repeat  |               |         |
+            """,
+            xml__contains=[
+                """<repeat>
+            <repeat2>
+              <age>23</age>
+            </repeat2>
+            <repeat2 jr:template="">
```
MartijnR: I think this template shouldn't be here.

pyxform/tests_v1/test_repeats.py
```diff
@@ -0,0 +1,119 @@
...
+          </repeat>
+          <repeat jr:template="">
+            <repeat2>
+              <age>23</age>
+            </repeat2>
+            <repeat2 jr:template="">
+              <age>23</age>
+            </repeat2>
+          </repeat>
+          """],
```
MartijnR: This nested template seems the correct way of doing this. If `<repeat>` didn't have a template it would be placed where you've put it above (above the previous comment).

pyxform/tests_v1/test_repeats.py
```diff
@@ -0,0 +1,119 @@
...
+            |        | integer      | age     | the age       | 23      |              |
+            |        | end repeat   | repeat  |               |         |              |
+            """,
+            xml__contains=[
+                """<repeat>
+            <age>23</age>
+          </repeat>
+          <repeat </age>
+          </repeat>
+          """],
```
MartijnR: There is an XML syntax error here. I think only one `<repeat>` would have to be in the model but am not sure if it matters.

MartijnR: Oh I see that syntax error is probably meant to be the template.

MartijnR:
I think Enketo may also not be able to deal with the first nested repeat instance in `test_nested_repeat_first_instance_defaults`. It would be a bug. I'll check it out. Thanks!

pyxform/tests_v1/test_repeats.py
lognaturel: Ohhh, I butchered it to see the output and never fixed it, sorry.

lognaturel:
@MartijnR I saw you're on holiday so let's do a release now and take our time to get this one just right for the next one!

MartijnR:
Sounds good. Thanks. I'm actually back already though! My holiday was only for a measly 2 days :( (though enough to be great fun).

pyxform/tests_v1/test_repeats.py
lognaturel: Interesting. To be clear, you're thinking that only the template instances of repeats should contain template instances of repeats? So with triple nesting, something like:

```
r1
  r2
    r3

r1 template
  r2 template
    r3 template
```

That does make sense I think. But I have no idea how to implement it...

pyxform/tests_v1/test_repeats.py
lognaturel: Wait, now I'm confused! :D

pyxform/tests_v1/test_repeats.py
lognaturel: You know what, the count doesn't affect the structure so I don't think this test is super meaningful. I meant to actually remove it. What do you think?

lognaturel:
> My holiday was only for a measly 2 days :( 

Aw! Here I was imagining you far off on a boat with a drink with a little umbrella in it... glad it was fun despite being short!

Looks like the next step is agreeing on what nested repeats should actually look like. I think you're probably right that all templates should be output followed by all non-templates. That will require a much larger code change, I think, unfortunately.

pyxform/tests_v1/test_repeats.py
MartijnR: Sorry! I was referring a very hypothetical case were `<repeat>` does not have a template but `<repeat2>` does have a template (which is not possible to create with XLSForm). In that case there would have to be at least 1 `<repeat>` instance, and I think the template for `<repeat2>` should be placed inside of this instance. I mentally extract `<repeat>` and consider it to a form by itself (with repeat as root). I think that approach covers all theoretical use cases. Whatever we come up with I think there should be only 2 templates in the above form, 1 for `<repeat>` and 1 for `<repeat2>`.

MartijnR:
I checked `test_nested_repeat_first_instance_defaults` in Enketo with [this complete form](test_nested_repeat_first_instance_defaults), and that works.

> That will require a much larger code change, I think, unfortunately.

What about the following approach:

1. let the template output remain as it was in pyxform (so for each `begin_repeat` there is one template and nested repeats have nested templates), 
2. copy each high-level repeat XML template node (ie. each template that does not have a template ancestor), strip the `jr:template` attribute from it and from its (nested) descendants and place this copy after (or before) its original?

lognaturel:
> copy each high-level repeat XML template node (ie. each template that does not have a template ancestor), strip the jr:template attribute from it and from its (nested) descendants and place this copy after (or before) its original?

I'm not seeing a great way to do that so I will close for now and hopefully have time to come back to it in the not too distant future!

PR #211, opened 2018-07-04
Release version 0.11.3

ukanga:


lognaturel:
No need to change for this but I'd be interested to know how you do versioning. With the number of new features in this release I would have expected this to be 0.12.0 rather than being a point release!

PR #214, opened 2018-07-17
Add support for prefix, delimiter and tag for SMS

lognaturel:
See https://opendatakit.github.io/xforms-spec/#compact-record-representation-(for-sms) for the XForms spec.

Turns out that `tag` is already an alias for `name` in pyxform columns so I used `short_tag` instead. I would love to deprecate some of the aliases at some point but that's a separate conversation.

One thing that's not totally explicit in the XForms spec is whether `odk:tag` only applies to questions. I believe that to be the case because it would be a mess otherwise (we don't have any notion of closing a tag) but @jd-alexander, @yanokwa can one of you please confirm?

**EDIT:** actually, I did write "Questions that have a tag attribute..." but now I can't remember whether we explicitly discussed this.

### DIFF
--- a/pyxform/aliases.py
+++ b/pyxform/aliases.py
@@ -43,6 +43,10 @@ settings_header = {
     u"sms_allow_media": constants.SMS_ALLOW_MEDIA,
     u"sms_date_format": constants.SMS_DATE_FORMAT,
     u"sms_datetime_format": constants.SMS_DATETIME_FORMAT,
+
+    u"prefix": constants.COMPACT_PREFIX,
+    u"delimiter": constants.COMPACT_DELIMITER,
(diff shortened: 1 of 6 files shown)


yanokwa:
I was operating under the assumption that tags are for questions. I don't know how it'd work for groups because the point of a tag is to have a short and mostly unique (although the spec doesn't say this, but maybe it should) way to identify a question.

lognaturel:
Thanks, @yanokwa! I changed the naming and force pushed. The only user-facing change is that `short_tag` has been changed to `compact_tag`. I left `prefix` and `delimiter` undecorated because I think the more we can match XForms the better.

PR #223, opened 2018-09-03
Make sure that the json scheme is a valid json

HaidarZ:


lognaturel:
Good catch, @HaidarZ! Could you please match the indentation of the other braces?

It would be helpful to know what you're using the json representation for. Currently, I don't believe any known downstream tools rely on it so there may be other parts in which it's incomplete or poorly-structured.

HaidarZ:
> Good catch, @HaidarZ! Could you please match the indentation of the other braces?
> 
> It would be helpful to know what you're using the json representation for. Currently, I don't believe any known downstream tools rely on it so there may be other parts in which it's incomplete or poorly-structured.

With respect to the tool , I used the integrated auto formatter found inside jetbrains pycharm.
I've checked all the file , it's now properly formatted.

PR #240, opened 2018-11-24
Issue 61 Table list supports additional appearance mods

### LINKED ISSUE #61: appearance=table-list breaks if additional appearance is added
Appearance `table-list` is properly detected, resulting correctly in inputs with `appearance="label"` and `appearance="list-nolabel"`

However, when adding another appearance, e.g. `table-list something`, it appears to not be detected. The output does not add `appearance="label"` and `appearance="list-nolabel"` to inputs in this group.

[test form](https://docs.google.com/spreadsheets/d/1iCU1xcu_76xJ0FLSRg4OEzRp_ccMImFaAWqQw_ko8GI/edit#gid=0)

Background: just something I happened to see when testing the boundaries of a new 'random' feature for select and select1 questions. 'appearance="random" is not actually planned to be supported in tables  ~~at least not initially~~ ever. 

### Update:

**a `randomize()` XPath function was introduced instead which is much better. This issue still stands though**

KeynesYouDigIt:
Fixes #61 

I found that where `table-list` is parsed was looking for an exact string match, and changed it to loot for `table-list` in the first item when `split()`ing the appearance string. 

This will only work when `table-list` is the first appearance attribute --- but it seems like that would normally be the case? Or should we support something like `minimal table-list` ? if so let me know (easy fix of using `in` instead of `==`)

Notes -- 
while nosetests are passing, there are massive differences in test_output that I don't understand.

I will work on getting rid of these, but if anyone has a quick idea on how to fix let me know


KeynesYouDigIt:
Yep, sure enough the big diff from test_output is from running tests


```
$ nosetests
..............................................................................................................................usage: nosetests (...)
.....................................................................................................................



$ git status
On branch table_list_supports_additional_appearance_mods
Your branch is up to date with 'origin/table_list_supports_additional_appearance_mods'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   pyxform/tests/test_output/ODKValidateWarnings.xml
	modified:   pyxform/tests/test_output/cascades_old.xml
	modified:   pyxform/tests/test_output/cascading_selec(...)
```

KeynesYouDigIt:
@ukanga or @yanokwa  or @lognaturel ?

ukanga:
> This will only work when table-list is the first appearance attribute --- but it seems like that would normally be the case? Or should we support something like minimal table-list ? if so let me know (easy fix of using in instead of ==)

I am not yet sure what use case we may have yet to have more than table-list but we could get there eventually. Can you make the change to support the check of `table-list` in the list of attributes instead?

ukanga:
On the XML differences - this seems to be happening when you are under python 3 - likely some rearrangement how the dictionary is ordered in python3 differs in python2 which was the original environment the tests were written. It should not prevent this specific change. Also, we may have included test output files in the codebase in some instances by mistake. Some cleanup will be required eventually.

ukanga:
@KeynesYouDigIt ping.

KeynesYouDigIt:
Hey sorry for the radio silence, im mostly limited to weekends on ODK for now -> patch pushed :) @ukanga 

For future reference, is branching from master or a different branch best practice?

martynscn:
@ukanga, I am in dire need of implementing something that is of the form of
a minimal table-list. Can you work on the variant you proposed?

Thanks a lot

ukanga:
> For future reference, is branching from master or a different branch best practice?

Branching from master is best practice.

pyxform/xls2json.py
```diff
@@ -653,7 +653,7 @@ def workbook_to_json(
                 # Code to deal with table_list appearance flags
                 # (for groups of selects)
                 ctrl_ap = new_json_dict.get(u"control", {}).get(u"appearance")
-                if ctrl_ap == constants.TABLE_LIST:
+                if ctrl_ap and constants.TABLE_LIST in ctrl_ap.split():
                     table_list = True
                     new_json_dict[u"control"][u"appearance"] = u"field-list"
```
ukanga: I think this line should also change. If we had other appearance attributes like `minimal table-list`, we only end up with `appearance="field-list"` when we should expect `appearance="field-list minimal"`.

KeynesYouDigIt: Hey back on this today, will let you know when I can patch it up

KeynesYouDigIt: Something like this as the test MD?
```MD = '''
| survey |                    |            |       |                   |                                                                                                                                                                                           |
|        | type               | name       | label |appearance         |
|        | begin_group        | tablelist1 | Table |table-list minimal |
|        | select_one yes_no  | options1a  | Q1    |                             
|        | select_one yes_no  | options1b  | Q2    |                                                                                                                                                                                         |
'''  # nopep8 ```

KeynesYouDigIt: I have what I _THINK_ I want in the tests, whats a good way to test this formally? use the enketo site and actually create the form? IM still not to familiar with how it should actually look

KeynesYouDigIt: Also, how do I add a "choices" tab in the Markdown?

KeynesYouDigIt: Nevermind think I found it

KeynesYouDigIt:
Passing CI, LMK what else it needs!

KeynesYouDigIt:
@ukanga poke.

ukanga:
Looks good.

PR #261, opened 2019-02-11
Silence warnings to prevent breaking the build

yanokwa:
This might look like a bad idea since errors won't break the build, but we will only lose errors that break installing wheel and the requirements. The former is fine, but the latter might mean we don't have the correct version of the requirements if the Windows images already have those packages installed. Seems like a pretty minor risk.


ukanga:
I think I will go with @bmarika changes on this.

PR #267, opened 2019-02-19
Ensure parameters are typed properly

yanokwa:
* Wrap parameters in`int()` to make the comparisons valid
* Use large and realistic numbers in the test


yanokwa:
Sorry about letting this one slip through, @ukanga. I'd appreciate hot fix!

Issue #269, opened 2019-03-04
Allows mismatched brackets (bug)


joeflack4:
## Description
Allows conversion of form without erroring from occurrences of mismatched brackets (variable curly brace notation).

## Additional information
For example, the following text was in a logic field at the time of conversion.

`${current_fp_method. -> ${current_fp_method}.`

This resulted in a confusing runtime error.

bmarika:
Hi @joeflack4, 
I would like to work on this issue. Is it possible to share a snippet of the excel file that you used where the data is not confidential? This will help me in replicating the issue on my end. 

If is it not possible to share the data, you share the `entire error log.`

PR #272, opened 2019-03-13
Ensure brackets are balanced when used to reference cells.

bmarika:
closes https://github.com/XLSForm/pyxform/issues/269


pyxform/xls2json_backends.py
```diff
@@ -220,6 +221,10 @@ def slugify(s):
                 u"%s_header" % sheet.name
             ] = xls_to_dict_normal_sheet(sheet)
 
+    if not balanced_brackets(json.dumps(result), '{', '}'):
+         raise PyXFormError(
```
ukanga: I am well aware that the codebase is not fully pep8 complaint, though I am of the opinion we should be reducing the violations rather than increasing them.
```sh
pyxform/xls2json_backends.py:225:10: E111 indentation is not a multiple of four
```

pyxform/utils.py
```diff
@@ -199,4 +199,24 @@ def get_languages_with_bad_tags(languages):
...
+    Ensure  brackets are properly matched.
+
+    When one is working with skip logic, the excel file needs to to be
+    validated to ensure that the curly braces are balances.
+    Otherwise, one could run into ODKValidateErrors after the file is
+    accepted by PyxForm.
+    """
+    opening_brackets = []
+    closing_brackets = []
+    for char in string:
```
ukanga: Would this be an expensive operation especially for large forms? Wondering if there is a different way to go through the whole JSON string a little bit faster. Perhaps:
- finding all occurrences of `{` and `}`
- using regex

Some benchmarks might help to identify if there are any bottlenecks.

ukanga: I think also we may need to worry about only brackets that should affect what is meant to be a variable, i.e.:
```
 ... | This is a ${variable} reference | ...
```
verses
```
... | I got a random { in a label | ...
```
In the latter case, I don't think there should be a matching closing `}`.

pyxform/xls2json_backends.py
```diff
@@ -220,6 +221,10 @@ def slugify(s):
...
+            "Please ensure that all the { } are matched properly"
```
ukanga: Is it possible to be more specific about which brackets we need to close? Some XLSForm files can be quite big and it may be difficult to narrow down which row is the culprit.

pyxform/tests/bug_tests.py
```diff
@@ -239,5 +239,14 @@ def test_xlsm_files_are_allowed(self):
         self.assertIsInstance(result, dict)
 
 
+class TestMismatchedBrackets(unittest.TestCase):
```
ukanga: Can you move this into the new style markdown based tests? See examples in https://github.com/XLSForm/pyxform/tree/master/pyxform/tests_v1. We have been working slowly at migrating from using Excel files in new tests unless it is necessary. I think this is an example that does not require us to add a 6.5kb file in the codebase.

pyxform/tests/bug_tests.py
```diff
@@ -239,5 +239,14 @@ def test_xlsm_files_are_allowed(self):
...
+    """Test that a spreadsheet with mismatched brackets is validated"""
+    def test_curly_brackets_validation_in_excels(self):
+        filename = "brackets_mismatch.xls"
```
ukanga: I think the relevant clause is faulty, besides the mismatched bracket, in the XLSForm file.

pyxform/utils.py
lognaturel: Agreed with @ukanga. Additionally, counting the braces is not sufficient -- they need to be properly balanced. For example, I believe something like `{}}}{{` would be considered valid by this check because the count of opening and closing braces match. A stack-based solution is typical for this kind of verification.

lognaturel: I've had this in the back of my head, apparently, and wanted to share a couple more thoughts I've had. This check should happen after the `${}` variables have already been expanded. That way, all you need to look for is something like `\${.* ` (`${` followed by anything followed by a space). Then you can output a warning for any matches. You do NOT want to check for all `{` and `}` being balanced because there could be a legitimate reason to use them in a context other than a variable to be expanded.

I would make it a warning because maybe someone did want to use a `${` without closing it somewhere. I don't feel strongly about that, though, and could be convinced it should be an error.

PR #287, opened 2019-04-15
issue 157 - warn users when any language is missing a tranlsation and blanking fields

KeynesYouDigIt:
I was having trouble detecting media columns and the like early, and I realized the media column wasn't really the problem here, it was the unlabelled column.

So I dug for where the `-` was being literally added to the XML as an answer value and added the warning there.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -191,7 +191,7 @@ class Survey(Section):
 
         return node(
             "h:html",
-            node("h:head", node("h:title", self.title), self.xml_model()),
+            node("h:head", node("h:title", self.title), self.xml_model(warnings)),
             node("h:body", *self.xml_control(), **body_kwargs),
             **nsmap
         )
@@ -562,6 +562,7 @@ class Survey(Section):
         When translations are not provided "-" will be used.
         This disables any of the default_language fallback functionality.
         """
+
         if warnings is None:
             warnings = []
 
@@ -571,19 +572,54 @@ class Survey(Section):
                 paths[path] = paths.get(path, set()).union(content.keys())
 
         for lang, translation in self._translations.items():
+            this_path_has_warning = False
             for path, content_types in paths.items():
                 if path not in self._translations[lang]:
                     self._translations[lang][path] = {}
+                    # missing question/question thingy thats like a hint
+                    question_and_column = path.split(":")
+                    missing_warning = self._generate_missing_translation_warning(
+                        lang, question_and_column[0], question_and_column[1]
+                    )
+                    warnings.append(missing_warning)
+                    # no need to warn about content types missing translations since the
+                    # whole path has a warning now.
+                    this_path_has_warning = True
+
                 for content_type in content_types:
                     if content_type not in self._translations[lang][path]:
-                        if lang == 'default':
-                            warnings.append('\tDefault language not set,' +
-                                            ' with no default translations for ' + content_type
-                                            + ' Please consider setting a `default_language` in your'
-                                            + ' settings tab to insure questions and options appear'
-                                            + ' as expected.'
-                                            )
-                        self._translations[lang][path][content_type] = "-"
+                        self._translations[lang][path][content_type] = u"-"
+                        # missing question thingy thats like media
+                        if not this_path_has_warning:
+                            # the path has a translation but the content type is missing one.
+                            missing_warning = self._generate_missing_translation_warning(
+                                lang, path.split(":")[0], content_type
+                            )
+                            warnings.append(missing_warning)
+
+    def _generate_missing_translation_warning(self, lang, question_name, column_name):
+        found_on_msg = (
+            "\n--\n Question missing translation: "
+            + question_name
+            + "\n Column missing: "
+            + column_name
+        )
+
+        if lang == "default":
+            return (
+                "\tDefault language not set," + " with missing default translations."
+                " Please consider setting a `default_language` in your"
+                + " settings tab to insure questions and options appear"
+                + " as expected."
+                + found_on_msg
+            )
+
+        return (
+            "\tMissing field translations found for "
+            + lang
+            + " field may not appear as expected."
+            + found_on_msg
+        )
 
     def _setup_media(self):
         """
@@ -734,7 +770,7 @@ class Survey(Section):
         # space to text
         # TODO: check out pyxml
         # http://ronrothman.com/public/leftbraned/xml-dom-minidom-toprettyxml-and-silly-whitespace/
-        xml_with_linebreaks = self.xml().toprettyxml(indent="  ")
+        xml_with_linebreaks = self.xml(warnings).toprettyxml(indent="  ")
         text_re = re.compile(r"(>)\n\s*(\s[^<>\s].*?)\n\s*(\s</)", re.DOTALL)
         output_re = re.compile(r"\n.*(<output.*>)\n(\s\s)*")
         pretty_xml = text_re.sub(
@@ -847,8 +883,6 @@ class Survey(Section):
         Print the xForm to a file and optionally validate it as well by
         throwing exceptions and adding warnings to the warnings array.
         """
-        print('print xform to file')
-        print('i has warnigns!')
         if warnings is None:
             warnings = []
         if not path:
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -562,7 +562,6 @@ class Survey(Section):
         When translations are not provided "-" will be used.
         This disables any of the default_language fallback functionality.
         """
-
         if warnings is None:
             warnings = []
 
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -178,7 +178,7 @@ class Survey(Section):
 
         return NSMAP
 
-    def xml(self):
+    def xml(self, warnings=None):
         """
         calls necessary preparation methods, then returns the xml.
         """
@@ -432,13 +432,13 @@ class Survey(Section):
                 yield i.instance
             seen[i.name] = i
 
-    def xml_model(self):
+    def xml_model(self, warnings=None):
         """
         Generate the xform <model> element
         """
         self._setup_translations()
         self._setup_media()
-        self._add_empty_translations()
+        self._add_empty_translations(warnings)
 
         model_children = []
         if self._translations:
@@ -555,13 +555,16 @@ class Survey(Section):
                             choice_value,
                         )
 
-    def _add_empty_translations(self):
+    def _add_empty_translations(self, warnings=None):
         """
         Adds translations so that every itext element has the same elements \
         accross every language.
         When translations are not provided "-" will be used.
         This disables any of the default_language fallback functionality.
         """
+        if warnings is None:
+            warnings = []
+
         paths = {}
         for lang, translation in self._translations.items():
             for path, content in translation.items():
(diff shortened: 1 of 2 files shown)


pyxform/tests_v1/test_language_warnings.py
```diff
@@ -47,24 +49,51 @@ def test_label_with_unknown_subtag_should_warn(self):
...
         self.assertTrue(len(warnings) == 1)
-        self.assertTrue('do not contain valid machine-readable codes: English (schm). Learn more' in warnings[0])
-        os.unlink(tmp.name)      
+        self.assertTrue(
+            'do not contain valid machine-readable codes: English (schm). Learn more' in warnings[0])
+        os.unlink(tmp.name)
+
+    def test_no_default_language_should_warn_with_unlabeled_media_tags(self):
+        # form should test media tag w NO default language set.
+        survey = self.md_to_pyxform_survey("""
```
lognaturel: Awesome that you are starting with a test! 👍 

I believe that this form is fine and should not result in a warning (a good test to have!) because none of the columns for user-facing strings specify a language. The issue is when only a subset of the columns do. That is, if you had `label::French (fr)` and `media::image`, there'd be a problem because the former specifies a language and the later doesn't. Similarly, `label::French (fr)` and `media::image::French (fr)` is fine but `label` and `media::image::French (fr)` is a problem.

KeynesYouDigIt: Yep thats exactly the problem I hit when I ran that test. Fixed now should be good to go! :shipit:

KeynesYouDigIt:
This one should be good to go, any idea why codecov drops so much?

KeynesYouDigIt:
@lognaturel should be good to go?

pyxform/tests_v1/test_language_warnings.py
```diff
@@ -47,24 +49,50 @@ def test_label_with_unknown_subtag_should_warn(self):
...
+            | survey  |                 |         |                     |                  |             |
+            |         | type            | name    | label               | choice_filter    | media::image|
+            |         | select_one opts | opt     | My opt              | fake = 1         | opt1.jpg    |
+            | choices |                 |         |                     |                  |             |
+            |         | list_name       | name    | label               | fake             |             |
+            |         | opts            | opt1    | Opt1                | 1                |             |
+            |         | opts            | opt2    | Opt2                | 1                |             |
+            | settings|                 |         |                     |                  |             |
+            |         | form_title      | form_id | public_key          | default_lanuage  |             |
+            |         | opts_form       | abc1    |                     | English (en)     |             |
```
lognaturel: This change doesn't make sense -- there is no column that uses the `English (en)` name. Also, this test was for another feature so please leave it as-is and add add additional tests.

pyxform/tests_v1/test_language_warnings.py
```diff
@@ -47,24 +49,50 @@ def test_label_with_unknown_subtag_should_warn(self):
...
 
         self.assertTrue(len(warnings) == 1)
-        self.assertTrue('do not contain valid machine-readable codes: English (schm). Learn more' in warnings[0])
-        os.unlink(tmp.name)      
+        self.assertTrue(
+            'do not contain valid machine-readable codes: English (schm). Learn more' in warnings[0])
+        os.unlink(tmp.name)
+
+    def test_no_default_language_should_warn_with_unlabeled_media_tags(self):
+        # form should test media tag w NO default language set.
```
lognaturel: This test should very narrowly identify the issue. For example, a label with `English (en)` set as the language and `media::image` with no language *and no other columns*. It would be good to also have a test with something like `label::English (en)` and `media::image::French (fr)`. There should be a warning any time that a language doesn't have complete translations.

lognaturel:
I haven't had a chance to look at this in detail yet but I made a couple of comments on the tests. In particular, the coverage is not complete.

Changing several core methods to take in a `warnings` parameter doesn't look like it should be necessary but I haven't yet dug in very deep. I think it should be possible to address just with local state.

pyxform/tests_v1/test_language_warnings.py
KeynesYouDigIt: eek good point, not sure where my head was at

pyxform/tests_v1/test_language_warnings.py
KeynesYouDigIt: I don't follow that second case im afraid - there should be a warning if there is a language outputting `-` instead of a word at any time? or when there is a third unlabelled column there should be a warning?

pyxform/tests_v1/test_language_warnings.py
KeynesYouDigIt: I think I was using this to figure something else out, sorry!

pyxform/tests_v1/test_language_warnings.py
KeynesYouDigIt: I think I follow, tests added lmk if they do the trick!

KeynesYouDigIt: Also going to try and cover untranslated choices and other stuff, not jsut media tags.

KeynesYouDigIt: So with the current functionality, each language known to miss at least one field will get at least one warning. This prevents warnings from being overly verbose when there are multiple translations missing.... but is that what we want? or should I extend warnings to add a warning for each question/field missing a translation? Slack me if that question is confusing and I will try to jump on.

KeynesYouDigIt:
I really think I wasn't updating the tests with my improved understanding of the problem as time went on here, gotta slowdown and do that. @lognaturel check out the new tests and let me know if they are closer to what you were thinking

KeynesYouDigIt:
After some thought I redisigned and added more tests. I'm curious about a particular case though -- 

```python
            | survey  |                 |         |                     |                |
            |              | type             | name    | label::English (en)      | hint           |
            |              | integer         | nums    | How many nums?      | nums is numbers|
```
Doesnt seem to generate a warning since theres no attempt to translate anything into `default` when running this. Why does a `media` tag generate this mysterious `default` translation but  a hint doesn't?

KeynesYouDigIt:
@lognaturel or @ukanga got a chance to review?

KeynesYouDigIt:
> Changing several core methods to take in a warnings parameter doesn't look like it should be necessary but I haven't yet dug in very deep. I think it should be possible to address just with local state.

I could add a sort of self.warnings attribute if this seems much more readable to everybody

KeynesYouDigIt:
bump @lognaturel / @ukanga

KeynesYouDigIt:
bit of rebasing craziness today but should be shippable now!

KeynesYouDigIt:
Anybody around to take another look at this? @ukanga / @lognaturel ?

ukanga:
Would it be ideal to be able to point out which columns a user should check to apply the recommended changes? It is hard to be able to tell which columns a user should, especially in situations where there may be a high number of columns. The warnings:
```
Warnings:
        Missing field translations found for French (fr) field may not appear as expected
        Missing field translations found for English (en) field may not appear as expected
        Missing field translations found for English field may not appear as expected
        Missing field translations found for French field may not appear as expected
        The following language declarations do not contain valid machine-readable codes: English, French. Learn more: http://xlsform.org#multiple-language-support
```
This warnings are generated with the column headers as below:

type | name | label::English (en) | label::French (fr) | hint::English | hint::French
-- | -- | -- | -- | -- | --

There is less searching in my opinion if the message included the columns that are likely affected such as:
```
Warnings:
        Missing field translations found for French (fr) field may not appear as expected (hint)
        Missing field translations found for English (en) field may not appear as expected (hint)
        Missing field translations found for English field may not appear as expected (label)
        Missing field translations found for French field may not appear as expected (label)
        The following language declarations do not contain valid machine-readable codes: English, French. Learn more: http://xlsform.org#multiple-language-support
```

ukanga:
@KeynesYouDigIt  I apologise for the delay. See my comment above.

vincebuscarello:
hi @ukanga - I thought about that and then figured it was out of scope, let me give this a look and see if I can pull it off!

KeynesYouDigIt:
@ukanga not sure why its failing but https://github.com/XLSForm/pyxform/pull/287/commits/7a666d66b48c0269d37e0925977d46e176877aac should take care of it!!

KeynesYouDigIt:
@ukanga  / @lognaturel code should be good, I am afraid I dont understand the circle errors at all. Is something wrong with the form validator CI is using?

pyxform/survey.py
```diff
@@ -538,25 +538,71 @@ def _setup_choice_translations(name, choice_value, itext_id):
...
+                    # whole path has a warning now.
+                    this_path_has_warning = True
+
                 for content_type in content_types:
                     if content_type not in self._translations[lang][path]:
-                        self._translations[lang][path][content_type] = "-"
+                        self._translations[lang][path][content_type] = u"-"
+                        # missing question thingy thats like media
+                        if not this_path_has_warning:
+                            # the path has a translation but the content type is missing one.
```
KeynesYouDigIt: I dont love this `path_has_warning` pattern at all. is there a list of `content_types` we support? couldnt find it in the spec for some reason, but id be happy to jsut make sure the content_type actually needs translation. That said if y'all are OK with this pattern then it should be fine to merge.

KeynesYouDigIt:
@yanokwa / @lognaturel  / @ukanga should be set to go?

lognaturel:
The tests and the output look great to me! 👏 

I would have added a pass over the translations in in `print_xform_to_file` to generate the warnings then instead of hooking into `_add_empty_translations` and having to pass `warnings` around. What do you think? Or we can see if @ukanga has thoughts on that.

yanokwa:
@KeynesYouDigIt It'd be a nice thing for you to try (and learn) to do an interactive rebase and reword and squash and force push so you have a clean history. Seems like there'd be 2 commits? One to fix the build, one to add the feature.

ukanga:
In relation to `warnings`, that is a pattern that had been established of which on `git blame` I see that I continued in that tradition. I think it can do with a refactor but it might touch on a lot more than translation. Perhaps we address that in a different pull request. The starting point is likely adding the `warnings` as a field in the class `SurveyElement`.

ukanga:
@KeynesYouDigIt I am sorry this has taken long for me to review, can you please see Yaw's advice https://github.com/XLSForm/pyxform/pull/287#issuecomment-509300820? I will be ready to merge then. The alternative is I squash merge the PR for the sake of cleaner commit history.

KeynesYouDigIt:
Yep! squashing and rebasing now, then ill flip warnings to be a class field. So it will be 3 not 2 commits, grab me right away if thats not mergeable.

lognaturel:
Awesome, @KeynesYouDigIt!

> ill flip warnings to be a class fields

If you haven't done this already, I'd say hold off. I think the scope can be tightened rather than broadened. But as @ukanga says, that could be a later refactor pass through.

ukanga:
I will be okay with one commit if it does not include the `.circleci` changes, you got about 2 commits in this PR.

ukanga:
I created an issue for the refactor at https://github.com/XLSForm/pyxform/issues/330.

ukanga:
The `OpenJDK-8-JRE-headless` install issues have to do with the package being [discontinued](https://github.com/XLSForm/pyxform/issues/332). We have resolved this failure in the current master branch, please rebase and drop the commits affecting `.circleci/config.yml`.

PR #288, opened 2019-04-24
Big-image support

### LINKED ISSUE #34: big-image and media::big-image column heading for itext tag is not available in XLSForm
There is a "big-image" itext value tag type that is not supported by XLSForm.

Usage section below is copied from this page: https://opendatakit.org/help/form-design/guidelines/

---

Itext value tag types

The possible choices for the form attribute of the itext <value> tag are:

```
"long",
"short",
"image",
"audio",
"video", and
"big-image".
```

By default, itext "image" values are not clickable. However, if you also include a "big-image", the image displayed by "image" will be clickable and will display a pannable, zoomable view of the file specified by "big-image". You can return to your form after opening a "big-image" by hitting the phone's "back" button. Specifying "big-image" alone has no effect, you must always include "image".

lincmba:
Include support for `big-image` column header
Only three media types were supported initially, including `big-image` support ensures support for all media types specified here https://opendatakit.github.io/xforms-spec/#supported-media-types
fixes #34

### DIFF
--- a/pyxform/question_type_dictionary.py
+++ b/pyxform/question_type_dictionary.py
@@ -371,4 +371,13 @@ QUESTION_TYPE_DICT = {
     "xml-external": {
         # Only effect is to add an external instance.
     },
+    "big-image": {
+        "control": {
+            "tag": "upload",
+            "mediatype": "big-image/*"
+        },
+        "bind": {
+            "type": "binary"
+        }
+    }
 }
--- a/pyxform/tests_v1/test_xmldata.py
+++ b/pyxform/tests_v1/test_xmldata.py
@@ -367,6 +367,39 @@ class ExternalInstanceTests(PyxformTestCase):
         xml = survey._to_pretty_xml()
         self.assertEqual(1, xml.count(expected))
 
+    def test_big_image_column(self):
+        """test Support for big-image itext value tag type """
+        md = """
+            | survey  |                |                 |                |       |                  |
+            |         | type           | name            | label          | image | media::big-image |
+            |         | select_one a_b | big-image-test  | big-image-test | y.jpg | y.jpg            |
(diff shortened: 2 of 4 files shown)


pyxform/tests_v1/test_sheet_columns.py
```diff
@@ -53,6 +53,21 @@ def test_column_case(self):
...
+        self.assertPyxformXform(
+            name="data",
+            md="""
+                    | survey |            |              |        |
+                    |        | type       | name         | label  |
+                    |        | image      | my_image     | Image  |
+                    |        | big-image  | my_big_image | BImage |
+                    """,
+            errored=False,
+        )
```
lognaturel: Please make sure the test actually validates the XML output!

pyxform/tests_v1/test_xmldata.py
```diff
@@ -367,6 +367,39 @@ def test_can__reuse_xml__external_then_selects(self):
...
+            """  # noqa
+        self.assertPyxformXform(
+            name="data",
+            md=md,
+            errored=False,
+            itext__contains=[
+                '<translation default="true()" lang="default">',
+                '<text id="/data/big-image-test/b:label">',
+                '<value form="image">jr://images/b.jpg</value>',
+                '<value form="big-image">jr://big-image/d.jpg</value>',
```
lognaturel: The scheme should be `jr://image/`, not `jr://big-image`.

pyxform/question_type_dictionary.py
```diff
@@ -371,4 +371,13 @@ def generate_new_dict():
     "xml-external": {
         # Only effect is to add an external instance.
     },
+    "big-image": {
```
lognaturel: There is no `big-image` upload type, it's only a media type that can be specified for question labels.

PR #291, opened 2019-04-30
Strip white spaces from headers

### LINKED ISSUE #13: Strip spaces from headers
When headers `name` or `type` are given with preceding or following spaces (ie, as `name`, `type` [1]), pyxform is not striping them. This causes a lot of confusing among newbies, especially, and in general. In Excel, it is incredibly hard to tell if there is a trailing space or not.
The same may also be true of the different sheets (`survey` and `choices`). While we are at it, the headers + sheet names should also be auto-lowercased.

Note, this may be a relatively easy fix for someone new to pyxform. As such, I added a "get-started-issues" label to this issue. We should add language to a "How to Contribute" section on pyxform somewhere to a list of issues tagged with this label.

[1] - Oh, you can't see the difference? That is because its pretty impossible to, by just looking at things. Or even github markdown strips spaces. One of the two :-P

lincmba:
Strip white spaces from headers.
Fixes #13

Signed-off-by: Lincoln Simba <lincolncmba@gmail.com>

### DIFF
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -4,10 +4,11 @@ A Python script to convert excel files into JSON.
 from __future__ import print_function, unicode_literals
 
 import codecs
-import json
 import os
 import re
 import sys
+
+import json
 
 from pyxform import constants, aliases
 from pyxform.errors import PyXFormError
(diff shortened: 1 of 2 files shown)


lincmba:
Sheet names are already being lowercased by this bit of code 
`workbook_dict = {x.lower(): y for x,y in workbook_dict.items()}` in `workbook_to_json()` function.

Issue #123, opened 2017-06-08
External choices with integers in names/values are converted to decimals


yanokwa:
If you have an external choice with the name/value of "1", pyxform converts this value to "1.0" in the itemsets.csv. This does not happen with regular choices.

Input
* [numeric.xlsx](https://github.com/XLSForm/pyxform/files/1059737/numeric.xlsx)

Output
* [numeric.xml.txt](https://github.com/XLSForm/pyxform/files/1059739/numeric.xml.txt)
* [itemsets.csv.txt](https://github.com/XLSForm/pyxform/files/1059740/itemsets.csv.txt)

phil-wilson:
@MartijnR Note this breaks cascading selects and/or using pulldata() functions, as the integer being converted to decimal does not match the original value in choices sheet or other csv source.

MartijnR:
Definitely a bug and a serious one.

yanokwa:
@phil-wilson @MartijnR Either of you able to send in a PR?

MartijnR:
No, sorry, I cannot unfortunately.

phil-wilson:
@spatialdude

PR #297, opened 2019-05-10
check that External choices with integer values are not converted to decimals during file conversion

WinnyTroy:


WinnyTroy:
Addresses #123

PR #304, opened 2019-05-21
Only convert yes/no to true()/false() on whitelisted attributes

### LINKED ISSUE #226: XLSForm "yes" --> XForm "true()", even for custom bind::'s
## Description
I have a custom `bind` I'm using for a testing program/library I'm building. In my XLSForm, the column is named "bind::xtest-linearAssert".

In that column, I write down things I want to assert; things I expect can be validly entered into a form. Then I convert the XLSForm to XML and run my program, which uses JavaRosa for parsing, on it.

However, I just noticed that when I enter "yes" into the custom column, it shows up in the XML as "true()". I believe that this is an accidental side effect of other "yes" --> "true()" conversions, e.g. the `required` field.

Just FYI, the assertions I'm running where I'd like to enter "yes" are for simple "yes/no" questions, so this will be a common test case.

## Expected behavior
Plain text "yes" in an XLSForm does not get converted to "true()" when in a custom bind field/column.

## Actual behavior
It is converted.

## Additional information
Here's the form I'm using:  [XformTest1.xlsx.zip](https://github.com/XLSForm/pyxform/files/2381503/XformTest1.xlsx.zip)

Here's an example of a bind where the "yes" was converted to "true()":
`<bind nodeset="/XFormTest1/date_group/system_date_check" required="true()" type="select1" xtest-linearAssert="true()"/>`

yanokwa:
Fixes #226

The whitelist comes from the XForms 1.0 attributes that we list at https://opendatakit.github.io/xforms-spec/#bind-attributes. 

Note that I didn't include `saveIncomplete`. That's because Collect doesn't actually check the value. It just checks the presence of the attribute. I'll be updating the spec.

I also don't include attributes like `jr:requiredMsg`, and `odk:location-priority`. That's because they don't accept boolean values.

### DIFF
--- a/pyxform/survey_element.py
+++ b/pyxform/survey_element.py
@@ -132,6 +132,14 @@ class SurveyElement(dict):
         "FALSE": "false()",
     }
 
+    CONVERTIBLE_BIND_ATTRIBUTES = [
+        "readonly",
+        "required",
+        "relevant",
+        "constraint",
+        "calculate",
+    ]
+
     # Supported media types for attaching to questions
     SUPPORTED_MEDIA = ["image", "audio", "video"]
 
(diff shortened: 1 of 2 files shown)


yanokwa:
Failing build seems to be because of a CI issue. 
```
E: Failed to fetch http://deb.debian.org/debian/pool/main/c/ca-certificates-java/ca-certificates-java_20170929~deb9u1_all.deb  404  Not Found
```

yanokwa:
Test has been added, @ukanga. Thanks for asking for it!

PR #305, opened 2019-05-22
Add support for track-changes

yanokwa:
Builds on https://github.com/XLSForm/pyxform/pull/304.

Implemented this PR per the spec at https://opendatakit.github.io/xforms-spec/#audit-attributes.
* Added a test for blank parameters because it seemed like a good idea.
* Added a test for expected and unexpected values for track-changes
* Added test to show track-changes and location work together
* Instead of checking for the number of params for location, we now check if you have the correct params.
* Reformatted code to match current style


yanokwa:
Rebased off master, then added changes from #304, then reformatted with black.

Issue #251, opened 2019-01-10
Add external instance if pulldata() is used in constraint, required, readonly, relevant expressions


MartijnR:
Similar to https://github.com/XLSForm/pyxform/issues/43, but extending this feature to look at all columns that contain XPath expressions. Currently only the calculate column is considered.

### Test for constraint:

| type | name | label | constraint
| -- | -- | -- | -- | 
| text | Part_ID | Participant ID |  pulldata('ID', 'ParticipantID', 'ParticipantIDvalue', .)

### Test for readonly

type | name | label | readonly
-- | -- | -- | -- |
text | Part_ID | Participant ID |  pulldata('ID', 'ParticipantID', 'ParticipantIDvalue', ${id})

### Test for required

type | name | label | required |
-- | -- | -- | -- |
text | Part_ID | Participant ID |  pulldata('ID', 'ParticipantID', 'ParticipantIDvalue', ${id})

### Test for relevant

type | name | label | relevant
-- | -- | -- | -- | 
text | Part_ID | Participant ID |  pulldata('ID', 'ParticipantID', 'ParticipantIDvalue', ${id})

Expected output for all 4 test, contains:

```xml
<instance id="ID" src="jr://file-csv/ID.csv"/>
```
### Test for multiple

type | name | label | constraint | relevant | required
-- | -- | -- | -- | -- | -- 
text | Part_ID | Participant ID |  pulldata('ID', 'ParticipantID', 'ParticipantIDvalue', .) |  pulldata('ID', 'ParticipantID', 'ParticipantIDvalue', ${id}) |  pulldata('ID', 'ParticipantID', 'ParticipantIDvalue', ${id})

Expected output contains **only a single occurrence** of :

```xml
<instance id="ID" src="jr://file-csv/ID.csv"/>
```

MartijnR:
Note that the tests in the commit for #43 uses an XML file. I believe that is no longer the recommend practice and markdown tests should be used (so they can be copied from this issue).

p-netm:
closed by #308

PR #308, opened 2019-05-23
Add external instance if pulldata() is used in constraint, required, readonly, relevant expressions

p-netm:
Add external instance if pulldata() is used in constraint, required, readonly, relevant expressions
- required
- constraint
- readonly
- relevant

- functionality of the same for: calculate was retained.


p-netm:
fixes #251

PR #310, opened 2019-05-28
Disallow duplicate column names within choices sheet

### LINKED ISSUE #23: PYXForm should fail if a choiceList has two or more choices sharing the same value.
When processing the XLSX file, it would be helpful if the conversion failed if a user has entered two rows in the choices sheet for a given choice list that have the same value. 

The use case is someone copy-and-pasting a series of answers:

listA  val1 Yes
listA  val2 No
listA  val2 Maybe

and not noticing that val2 is repeated twice (Maybe should be val3).

If we want to support the current mechanism where this is allowed, we should probably have 
a settings tab flag that allows for duplicated choice values.

WinnyTroy:
Fixes #23


pyxform/xls2json.py
```diff
@@ -477,6 +478,32 @@ def workbook_to_json(
...
                     del option[headername]
+        list_name_choices = [option.get("name") for option in options]
+        if len(list_name_choices) != len(set(list_name_choices)):
+            for k, v in Counter(list_name_choices).items():
+                try:
+                    if (
+                        v > 1
+                        and settings["allow_choice_duplicates"].capitalize() != "Yes"
+                    ):
+                        raise PyXFormError(
```
lincmba: Could we raise once with all the duplicate columns have been identified?

WinnyTroy: Sure. changed the implementation to pick all the duplicated choice list names present.

Issue #274, opened 2019-03-19
Code cleanup


ukanga:
- [x] use https://github.com/ambv/black #306
- link other code issues to this issue
- check code quality on PRs, should be improving not regressing.

lindsay-stevens:
Since 2021 there are [github actions](https://github.com/XLSForm/pyxform/tree/master/.github/workflows) run for every PR. It has a `verify` step that runs linters: black, isort, pycodestyle, and flake8; and a `test` step that runs the test suite in the supported environment combos.

Not sure there's much more that'd be worth running on every PR. For example: coverage, mypy, vulture etc can be useful but at this stage I'd prefer to run those as part of a cleanup PR than forcing it on every contribution.

PR #314, opened 2019-06-12
Default Pretty Print to False

### LINKED ISSUE #162: Default to no pretty print to save space
According to https://forum.opendatakit.org/t/-/10712/3, not using pretty print can save more than a megabyte on large forms. 

Pretty print on
```xml
<text id="static_instance-wards-10339">
    <value>Gangare</value>
</text>
```
Pretty print off
```xml
<text id="static_instance-wards-10339"><value>Gangare</value></text>
``` 
I propose we make pretty print off the default behavior because I don't think a lot of people are editing XML forms after the fact.

p-netm:
if the cli flag ** --pretty_print ** is not given then xform file is printed as ugly, explicitly providing the flag converts .xlsx file to pretty-printed .xml. This as discussed [here](https://forum.opendatakit.org/t/-/10712/3) has the advantage that the xform sizes have relatively less size when ugly_printed as opposed to when pretty-printed.

fixes #162


ukanga:
@p-netm I would recommend you avoid the renaming of the tests for right now unless the reason is compeling, It would be better to reimplement the tests in `pyxform/tests_v1/` and remove them from here. It is my intention to do some cleaning on tests that are in `pyxform/tests/`, any pointers will be welcome and you can make some suggestions on #274.

Could you please rebase your work with the latest master for another review?

p-netm:
> I would recommend you avoid the renaming of the tests for right now unless the reason is compeling

Yeah, i think pytest recognizing these files as test files while they exist is compelling enough reason. for as long as the tests exists testing with pytest previously would result in false positives that would only be caught in the CI/CD step because nosetests does pick up on them.

pyxform/xls2xform.py
```diff
@@ -97,9 +97,9 @@ def _create_parser():
         help="Run the Enketo Validate XForm external validator.",
     )
     parser.add_argument(
-        "--no_pretty_print",
-        action="store_false",
-        default=True,
+        "--pretty_print",
```
ukanga: @p-netm I think this change can be made with the existing API without changing it. I would recommend you maintain existing API and make the default not pretty print by adjusting only `True/False` of affected variables. I think this change should only apply to this file, the change `survey.py` should still remain, any code that is using `pyxform` as a library can make the choice of whether it wants a pretty printed version or not.

p-netm:
> @p-netm I would suggest the renaming of tests should be left out of this pull request entirely, it has nothing to do with the issue being addressed in this PR.

ok,... ill see if i can revert those changes.

pyxform/xls2xform.py
p-netm: so just to confirm: the argument ```--no_pretty_print``` should be retained and the only change i need to make is the argument's default boolean value. Also i should revert the changes i made to the survey.py file
@ukanga Is this correct?

ukanga: Yes, that's correct.

p-netm: I agree the change to ```suvey.py``` is trivial and I only did that for consistency in default values.

however for the command line args; I think that this may inconvenience say someone converting an xlsfile from the command line and wants to pretty print. So for such a case, *how would one specify the argument to pretty print?*, given that the file will by default be converted to ugly xml

ukanga: I would expect when you add the variable false to it, but I suppose that does not seem as friendly. In that respect then I will accept the change for the parameter to `--pretty-print` with your changes.

p-netm:
Changes requested in review will be merged form here PR #324

PR #318, opened 2019-06-18
This fixes #317 - Report control name

### LINKED ISSUE #317: Report the name of a not closed control
Currently xls2json.py reports a not closed control with the text: 

Unmatched begin statement: [control_type]

It is not easy for the user to know which control is the one having problems particularly with big surveys. It would be good to report the control name

qlands:
This changes the stack from an array of tuples to a array of dicts {control_type, control_name,parent_children}


yanokwa:
@qlands Thanks for your PR. Could you please include a test?

pyxform/tests/bug_tests.py
```diff
@@ -36,6 +36,25 @@ def runTest(self):
             survey.print_xform_to_file(output_path, warnings=warnings)
 
 
+class NotClosedGroup(unittest.TestCase):
```
ukanga: Can you please use the new way we right tests? See examples in [pyxform/tests_v1](https://github.com/XLSForm/pyxform/tree/dd584b85032fcd74ba55f0ab3ecaf22e26e7d7a8/pyxform/tests_v1) . This test could probably be added in [pyxform/tests_v1/test_groups.py](https://github.com/XLSForm/pyxform/blob/master/pyxform/tests_v1/test_groups.py).

ukanga: Also, it would be nice to confirm the exception raised is a `PyXFormError` and has the expected message. An example is at https://github.com/XLSForm/pyxform/blob/master/pyxform/tests_v1/test_xmldata.py#L26.

qlands: OK. Shall I remove the old testing code in bug_tests.py?

PR #322, opened 2019-07-01
Warn users where repeated or complex logic is found.

### LINKED ISSUE #102: Check for large or repeated relevance and warn of performance problems
Sometimes users build forms with repeated or long complex logic that will cause stack overflows downstream. The form that triggered this request had this 1300 character relevance repeated 2x and the inverse repeated 5x.

> ${B1_1StoveType}=23 or ${B1_1StoveType}=9 or ${B1_1StoveType}=10 or ${B1_1StoveType}=11 or ${B1_1StoveType}=12 or ${B1_1StoveType}=13 or ${B1_1StoveType}=14 or ${B1_1StoveType}=15 or ${B1_1StoveType}=16 or ${B1_1StoveType}=17 or ${B1_1StoveType}=18 or ${B1_1StoveType}=19 or ${B1_1StoveType}=20 or ${B1_1StoveType}=21 or ${B1_1StoveType}=22 or ${B1_2StoveType}=23 or ${B1_2StoveType}=9 or ${B1_2StoveType}=10 or ${B1_2StoveType}=11 or ${B1_2StoveType}=12 or ${B1_2StoveType}=13 or ${B1_2StoveType}=14 or ${B1_2StoveType}=15 or ${B1_2StoveType}=16 or ${B1_2StoveType}=17 or ${B1_2StoveType}=18 or ${B1_2StoveType}=19 or ${B1_2StoveType}=20 or ${B1_2StoveType}=21 or ${B1_2StoveType}=22 or ${B1_3StoveType}=23 or ${B1_3StoveType}=9 or ${B1_3StoveType}=10 or ${B1_3StoveType}=11 or ${B1_3StoveType}=12 or ${B1_3StoveType}=13 or ${B1_3StoveType}=14 or ${B1_3StoveType}=15 or ${B1_3StoveType}=16 or ${B1_3StoveType}=17 or ${B1_3StoveType}=18 or ${B1_3StoveType}=19 or ${B1_3StoveType}=20 or ${B1_3StoveType}=21 or ${B1_3StoveType}=22 or ${B1_4StoveType}=23 or ${B1_4StoveType}=9 or ${B1_4StoveType}=10 or ${B1_4StoveType}=11 or ${B1_4StoveType}=12 or ${B1_4StoveType}=13 or ${B1_4StoveType}=14 or ${B1_4StoveType}=15 or ${B1_4StoveType}=16 or ${B1_4StoveType}=17 or ${B1_4StoveType}=18 or ${B1_4StoveType}=19 or ${B1_4StoveType}=20 or ${B1_4StoveType}=21 or ${B1_4StoveType}=22

These kinds of relevancies create a ton of objects in JavaRosa's form parsing and on Android 4.4, where garbage collection is terrible, you get a stack overflow.

It'd be nice if pyxforms warned users about long logical expressions or suggested storing repeated logic in a calculate and referring to that calculate.

p-netm:
closes #102 

- Adds 2 utils functions:
  -  a function that checks for repeated relevance expressions
  - a function that checks for complex relevance expressions

Complex_expression checker function implements a naive algorithm based on the number of variables in the expression that are operated on by `or|and` logical operators.

### DIFF
--- a/pyxform/utils.py
+++ b/pyxform/utils.py
@@ -272,46 +272,45 @@ def default_is_dynamic(element_default, element_type=None):
             else:
                 expression.append(expression_element)
     return contains_dynamic
-def expression_is_repeated(expression, index, expression_hash_map):
+
+def expression_is_repeated(expression, expression_hash_map, row_number):
     """
-    Checks If a logical expression is complex or repeated
+    Checks If a logical expression repeated
 
     expression (str) - text representing expression to be tested
-    index (int) - attempt at giving the location of the error location
-    expression_hash_map (dict) - store hashes for error messages as keys and index
-        as values
-    return (list) the warning message in a list or an empty list if no warning
+    expression_hash_map (dict) - store hashes of the expressions as keys
+        and their indexes as values on the work sheet
+    row_number (int) - location of the row in the sheet
+    return (bool) True if expression is already in surveySheet
     """
-    warnings_list = []
-    actual_row = (
-        index + 1
-    )  # this is assuming that order of rows is maintained to this point
-    this_expression_hash = md5(expression.encode()).hexdigest()
-    if this_expression_hash in expression_hash_map.keys():
-        # get list of locations where expression has been seen before
-        row_indices = expression_hash_map[this_expression_hash]
-        row_indices.append(actual_row)
-        warning_message = """%s: Duplicate relevancies detected.
-        In future, its best to store repeated logic in calculate 
-        and referring to that calculate.""" % (
-            ", ".join(map(str, row_indices[1:]))
+    if expression in expression_hash_map.keys():
+        return (
+            row_format_string % row_number
+            + " Duplicate expression detected. In future, "
+            + "it is best to store repeated logic in calculate"
+            + " and referring to that calculate."
         )
-        warnings_list.append(warning_message)
     else:
-        expression_hash_map[this_expression_hash] = [actual_row]
+        expression_hash_map[expression] = row_number
+        return ''
 
-    return warnings_list
 
-
-def expression_is_complex(expression, index):
+def expression_is_complex(expression, row_number):
     """
-    A heuristic that checks whether a logical relevance expression is complex
+    A heuristic that checks whether a logical relevance expression is complex.
+    using the number of logical combinations, which is imperatively a function
+    of the total number of operands whose operators are logical in nature i.e. and & or
 
     expression (str) - text representing expression to be tested
-    index (int) - 0-based index that points to where the expression is
-    return (str) - a warning message one is necessary
+    row_number (int) - location of the row in the sheet
+    return (string) - the warning message, blank string if no warning
     """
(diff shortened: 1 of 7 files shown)


pyxform/utils.py
```diff
@@ -270,3 +275,46 @@ def default_is_dynamic(element_default, element_type=None):
...
+    expression (str) - text representing expression to be tested
+    row_number (int) - location of the row in the sheet
+    return (string) - the warning message, blank string if no warning
+    """
+    regex = r"or|and"
+    num_of_variables = re.split(regex, expression)
+    if len(num_of_variables) > COMPLEX_RELEVANT_VARIABLES_MAX_THRESHOLD:
+        return (
+            row_format_string % row_number
+            + " Possible complex or long logical expression detected, "
```
lognaturel: How about "Long relevance expression detected. This may slow down form opening or filling. See http://xlsform.org#performance to learn more." At that page (which I can write), we can describe techniques to make sure forms run as fast as possible.

pyxform/utils.py
```diff
@@ -270,3 +275,46 @@ def default_is_dynamic(element_default, element_type=None):
...
+        and their indexes as values on the work sheet
+    row_number (int) - location of the row in the sheet
+    return (bool) True if expression is already in surveySheet
+    """
+    if expression in expression_hash_map.keys():
+        return (
+            row_format_string % row_number
+            + " Duplicate expression detected. In future, "
+            + "it is best to store repeated logic in calculate"
+            + " and referring to that calculate."
```
lognaturel: How about "Duplicate expression detected. Consider introducing a calculate to speed up form opening or filling. See http://xlsform.org#performance to learn more."

lognaturel: I think that this check is actually pretty problematic because it will complain about expressions that are references to calculates! That is, let's say you have the expression `${foo} > 2` a hundred times in your form. You'll get 100 warnings. Then you go and introduce the calculate `bar` and replace your expression with `${bar}`. Now you'll get 100 warnings about `${bar}` being repeated.

Additionally, not all expressions are created equal. `${foo} > 2` is not a big deal computationally and introducing a calculate to capture that expression has a cost because it makes the XLSForm longer and harder to read. Also, not all duplicates are the same. The same expression 3 times is not the same thing as the same expression 100 times.

I tried this with https://github.com/SwissTPH/WHO_VA_2016/blob/master/WHOVA2016_v1_5_2_XLS_form_for_ODK.xlsx and got 161 warnings. This warning would drown out any other useful information with large forms.

My current thinking is that we should not warn for every single duplicate expression. An alternative would be something like showing a single warning that points to the XLSForm documentation if there are any duplicate relevant expressions.

lognaturel:
I think if we want to consider something of this nature we should take a step back and discuss a game plan. I think the complex check is likely useful and the duplicate one is likely best to avoid. In fact, JavaRosa/Collect already optimizes away those duplicates so they would not affect Collect performance.

PR #323, opened 2019-07-01
capture and clean all the data in the external choices sheet

WinnyTroy:


lognaturel:
This looks like a fix to #326, right? Would be great to have issues to track known issues so it's easier to help users who run into them! 😊

ukanga:
Yes, it does fix #326. It slipped my mind. I will organise another release this weekend.

lognaturel:
You're the best, @ukanga. 💜

PR #340, opened 2019-07-23
Spelling fix

yanokwa:
I confirmed that (case-insensitive)`sepatator` is no where else in the source code.


yanokwa:
Build error seems just to be with codeconv and not this change.

PR #341, opened 2019-07-24
Add support for start-geopoint

### LINKED ISSUE #336: Add geopoint-start to generate setgeopoint action triggered by the odk-instance-first-load event
| type          	| name        	|  
|---------------	|-------------	|
| geopoint-start 	| my-location 	|   

ODK XForms
```xml
<bind nodeset="/data/my-location" type="geopoint" />
<odk:setgeopoint ref="/data/my-location" event="odk-instance-first-load" />
```

See forum discussion at https://forum.opendatakit.org/t/xlsform-spec-proposal-auto-location-type-for-background-location-capture-with-setlocation-action/20831

There's a chance that the type name will change but the implementation will be the same.

yanokwa:
Closes #336

The template of the question is in `question_type_dictionary.py`. I define a new attribute, `action` that will later be used to get all actions in the form. 

Note that in the ODK XForms spec, [actions](https://opendatakit.github.io/xforms-spec/#actions) can happen after the bind (what I'd like to call `model actions`) or inside a control (what I'd like to call `control actions`). For now, I name functions/variables generically and expect what we can parameterize these later (e.g., add a `type` attribute whose value can either be `model` or `control`)

In `survey_element.py`, I took the code we use for binds and make minor changes. The differences here are that I check for the existence of an `action`, drop the `name` attribute from what we pass on, and use `ref` instead of `nodeset`. I'm not sure that `self.get_xpath()` is the correct method to use for the ref, but it does seem to work.

In `survey.py`, we take the JSON that we've built and output it.

### DIFF
--- a/pyxform/builder.py
+++ b/pyxform/builder.py
@@ -51,14 +51,15 @@ class SurveyElementBuilder(object):
     # we use this CLASSES dict to create questions from dictionaries
     QUESTION_CLASSES = {
         "": Question,
+        "action": Question,
         "input": InputQuestion,
-        "trigger": TriggerQuestion,
-        "select": MultipleChoiceQuestion,
-        "select1": MultipleChoiceQuestion,
         "odk:rank": MultipleChoiceQuestion,
-        "upload": UploadQuestion,
         "osm": OsmUploadQuestion,
         "range": RangeQuestion,
+        "select": MultipleChoiceQuestion,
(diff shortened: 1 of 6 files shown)


yanokwa:
This PR implements the spec, but it won't pass validation until https://github.com/XLSForm/pyxform/pull/342 is finished.

PR #342, opened 2019-07-24
Add Validate v1.13.0 with JavaRosa v2.15.0

yanokwa:


yanokwa:
I can confirm that the problem here is with ODK Validate, not the conversion. 

If I change `run_odk_validate` to `False` in `pyxform/tests_v1/test_support_external_instances.py` and run `nosetests pyxform.tests_v1.test_support_external_instances`, the tests pass.

PR #344, opened 2019-07-30
Make it possible to turn a select_one into a select_multiple without …

yanokwa:
This PR is response to the discussion at https://github.com/opendatakit/javarosa/issues/263#issuecomment-369019465.

Note that this is will prevent existing forms from converting, so it's a breaking change. If it helps, ODK Build also does not allow spaces for select choices.


ukanga:
And considering some platforms like onadata allow form replacements - it might be ideal to figure out how to accommodate this change for existing forms - as a completely not allowing the replacement would be out of the question. I am wondering if it should be a warning instead?

yanokwa:
I'm not sure I understand what form replacements have to do with this. Can you describe the scenario you are worried about in greater detail?

ukanga:
I have a form with select one choices, and some of the choices have spaces in them already; for example, the choice `"mushroom rocks"`. If I happen to modify the form such that I do not change the select one, and it's choices, with this check, I will not be allowed to replace or create a new version of the form unless I change the choice from `"mushroom rocks"` to perhaps `"mushroom_rocks"` which has no spaces.  If I had already hundreds of submissions with the choice `"mushroom rocks"`,  I will now have data that has both `"mushroom rocks"` and `"mushroom_rocks"` which means the same thing but complicates already existing workflows when you do data analysis. You are also very likely to treat the two choices differently in the analysis if there is no transformation. I consider this a breaking change on other platforms in the analysis of already existing data other than pyxform alone during transformation.

The reason we have validation for select multiples choices not have spaces is because the data is stored as a space separated list. Perhaps we may want to reconsider the approach on how we allow choices with spaces to be represented; it could give us a different take on a solution to this issue.

yanokwa:
Reworking how we deal with spaces in multiple choices a big ecosystem change. I don't think the benefits outweigh the costs.

I think for this narrow issue, a warning is the best approach. But, the warning message doesn't seem particularly compelling. "Warning: Using a space in a select_one choice list makes it slightly harder use the same list for a select_multi."

@MartijnR Hate to throw you under the 🚌, but this was your idea. Do you still think it's compelling? What wording would you use for the prose?

MartijnR:
@ukanga makes a good point about form replacements and data analysis consequences, that I never thought of. I cannot think of a good way of dealing with this and disallowing spaces in select_single.

I think a warning would have to be strong to avoid creating such forms, as a first step, so that we can quite safely (?) disallow spaces in a few years. Something like "Warning: a select_one option name contains a space. We strongly advice to not use spaces. They will be disallowed in the future." 

I think it's worth making this warning step if the long-term goal is still to ban spaces. It will be nice and logical to not have different list requirements for select_one/select_multiple (and `count-selected(select1)` would eventually work correctly).

lognaturel:
I like @MartijnR's approach and can't think of anything better that can be done currently. I don't love "We strongly advice to not use spaces" because it leads to questions like "who is 'we'" and "why." How about also listing up to 3 options that have spaces to make it easier to address? I'm thinking something like 
```
Warning: a select_one option name contains a space (option 1, option 2, option 3, ...). To simplify analysis and make select ones convertible to select multiples, spaces in option names will be disallowed in the future.
```

yanokwa:
Had to delete my repo to fix some local git problems. 😭

Closing in favor of this new PR: https://github.com/XLSForm/pyxform/pull/364.

PR #345, opened 2019-07-30
Ignore pyxform_validator_update files that are user-specific

yanokwa:
Unlike `installed.json` which is repo-specific, these files are user-specific and so should be probably be ignored. One alternative might be that we move these to a `/tmp`, but `.gitgnore` seemed like the narrowest solution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ env
 # ignore pypi manifest
 MANIFEST
 pyxform/tests/test_output/
+
+# ignore pyxform_validator_update files
+pyxform/validators/odk_validate/.last_check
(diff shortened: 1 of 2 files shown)


lindsay-stevens:
@yanokwa FWIW from way back in #171 I figured the /bin dir could be gitignored too, so that pyxform didn't need to be updated for every ODK validate release.

yanokwa:
@lindsay-stevens We should do that eventually, but feels easier for users to have it bundled for now since they are fairly coupled. 

Relatedly, we might want to deprecated Validate and put that verification into JR. https://github.com/opendatakit/javarosa/issues/463#issuecomment-512997521

PR #348, opened 2019-08-07
Use pipenv to manage dependencies.

KeynesYouDigIt:
Pipenv is the recommended dependency manager, and is something we could use to manage custom commands like `$ pyxform_validator_update odk update ODK-Validate-v.....1.jar`


KeynesYouDigIt:
Still needs readme update, clean install test

KeynesYouDigIt:
@ukanga should the codecov failing actually be a problem for merging this? doesnt seem like testable code.

Issue #203, opened 2018-05-22
Why aren't instances generated for all selects?


lognaturel:
From `survey.py`:
```
Note that per commit message 0578242 and in xls2json.py R539, an
instance is only output for select items defined in the choices sheet
when the item has a choice_filter, and it is that way for backwards
compatibility.
```

From https://github.com/XLSForm/pyxform/commit/0578242:
```
Switching back to old way of handling or_other selects, and no longer supporting them 
in filtered selects. Also, no longer using itemsets for selects that don't have 
a choice_filter. The reason for those changes is backwards compatibility.
```

Does anyone know which tool(s) this ensures compatibility with?

The commit was from 2012, is it possible it's not needed anymore? I'm asking because I like consistency and having less code to think about. I'm also thinking about additions/variations on instances and selects that we're introducing (e.g. #202, #176) and it would be simpler to only have one case with those.

CC @ukanga @MartijnR

MartijnR:
The advantage that the non-itemset selects have is that they can safely be considered to be static by a (crude) data collection client, and therefore don't risk adding a task to the evaluation cascade thereby improving form loading and form traversal performance.

However, I can see the appeal for simplifying code and a slightly intelligent data collection client would be able to discard itemset nodesets that do not have an XPath predicate (choice_filter) from the evaluation cascade. 

So I'd have no problem to consistently use itemsets.

lognaturel:
Thanks, @MartijnR! Sorry, I'm not exactly sure what you mean by "evaluation cascade" in this context. I think of cascading in the context of recomputing dependent calculations using the [XForms DAG](https://www.w3.org/TR/2009/REC-xforms-20091020/#rpm-processing-recalc-mddg) when certain values change in the form. Do you mean that in the case of nodesets defined without an XPath predicate they may still be included in the DAG? I suppose in that case they wouldn't be dependent on anything, though, right? I think I'm missing something!

MartijnR:
Yes, I meant all logic dependencies calculations (CommCare calls it a [cascade](https://dimagi.github.io/xform-spec/#cascade)). 

Indeed an itemset **without a predicate refererring to a node inside the primary instance** wouldn't be dependent on anything so _shouldn't_ affect performance (and doesn't in Enketo). However, a less optimized client would just evaluate everything every time (like the first version of Enketo in 2012). 

So I'm fine with this proposed change.

lognaturel:
Ok, that makes sense, thank you!

[...]

PR #354, opened 2019-08-30
Report the choice name in the JSON

qlands:
The current JSON output does not contain the list name for "normal" select_one or select_multiple variables. It is useful to have it specially if one wants to construct a data dictionary.

I used the same key ("query") as is already used by "select one external" types.


lognaturel:
The `query` concept is introduced by the select one external feature and is not actually part of [the form spec](https://opendatakit.github.io/xforms-spec/0 so I don't think it makes sense as an attribute name here.

I think `list_name` would be more appropriate.

I don't work on any tools that use the JSON representation. @jnm, you do, right? I think it's generally in an odd state and would be worth some cleanup if folks are depending on it.

qlands:
Hi,

We use the JSON output extensively in our applications through [ODK Tools](https://github.com/qlands/odktools) which convert an ODK Excel Survey file into a MySQL database. The list name is used to create a child look-up table with the options that then is linked to the parent data-table. The list_name is important because we use it to name the look-up table.

I don't know who is the maintainer of the JSON part of pyxform but I would be happy to contribute if it does not have a patron.

I will upload the change to the PR soon.

lognaturel:
Thanks for that extra context, @qlands. I understand why in your case you'd want to represent the choice list once even though in the corresponding XForm it would be copied for each question that uses it. What I might do in that case is use a hash of the `children` to name the table. Could that possibly work for you?

In https://github.com/XLSForm/pyxform/issues/203, I ask why static choice lists are ever generated. Another possible resolution to your issue would be to always generate secondary instances for choice lists since those do have names (accessible through `itemset`).

Because there is no concept corresponding to `list_name` in the XForm representation, I'm still not very comfortable with this but I'd like @jnm or @ukanga to chime in.

qlands:
I could generate a hash as name but such name will not correspond to anything in the XLS form. 

Secondary instances generate "itemset" or "query" however I cannot force users to a particular format for select. I understand that list_name is not part of XForm however I'm not asking to change the XForm representation just the JSON by adding the list name. 

On other note: I ran black on my files....where can I check which file went without black?

qlands:
Sorry people... I messes up with this PR. I will open other one cleaner without so many reverts!!!

qlands:
We can continue the discussion in #357 . Thanks.

Issue #270, opened 2019-03-12
Add warnings to or_other users about Google Sheets compatibility


yanokwa:
Google Sheets submissions don't allow underscores in the column name. So if someone uses or_other, it won't work. We should warn on form conversion. 

https://forum.opendatakit.org/t/add-an-open-ended-other-choice-to-a-list-of-options/11820

PR #356, opened 2019-09-05
Google submissions do allow underscores

### LINKED ISSUE #353: Google Sheets warnings is overwhelming when there are many fields with underscores
If you have a long form with underscores in the column, pyform generates a warning. Because the warning lists all the columns with the problem, can overwhelm the other warnings we get.

yanokwa:
Closes #353 and reverts https://github.com/XLSForm/pyxform/pull/299 because ODK Collect v1.23.3 allows submissions with underscores as shown here:
Screen Shot 2019-09-05 at 2 41 49 PM

For further proof, https://github.com/opendatakit/collect/pull/1851/commits/3cca371037e21178b910ee0fadaa49cbf99cdf15 shows that the code in Collect that does this check was removed.

In the future, we really should test these PRs end to end before merging.

### DIFF
--- a/pyxform/tests/xls2json_tests.py
+++ b/pyxform/tests/xls2json_tests.py
@@ -203,19 +203,6 @@ class BasicXls2JsonApiTests(TestCase):
         ]
         self.assertEqual(choice_filter_survey.to_json_dict()["children"], expected_dict)
 
-    def test_underscore_warnings(self):
-        """Raise warnings incase there are underscores in column names"""
-        warnings = []
-        parse_file_to_json(utils.path_to_text_fixture("hidden.xls"), warnings=warnings)
-        self.assertGreater(len(warnings), 0)
-        warning = (
-            "Google Sheets submissions don't allow underscores in the "
-            "column name. If you intend to use Google Sheets "
-            "submissions, replace underscores with hyphens in the "
-            "following names: hidden_test"
-        )
-        self.assertIn(warning, warnings)
-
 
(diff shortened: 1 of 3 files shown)


lognaturel:
To do a little post mortem here:
- in Feb 2018, a user posted on the forum with a screenshot from Collect showing an error message about underscores in column headers with Google Sheets
- some time before Mar 2018, Google Sheets started supporting underscores in column headers
- in Mar 2018, @yanokwa filed https://github.com/XLSForm/pyxform/issues/270 to warn about underscores in column headers
- also in Mar 2018, https://github.com/opendatakit/collect/pull/1851 was merged. This PR was to add repeat support in the Google Sheets integration but it also snuck in opendatakit/collect@3cca371 which removes the error message the user had posted about
- in May 2019, @lincmba submitted https://github.com/XLSForm/pyxform/pull/299 to warn about underscores in field names

This is a tough one because integrations are involved. We obviously we have no control over Google Sheets and the folks who work on Collect and on pyxform are not all the same.

Some things we can improve:
- file issues for problems found while solving another issue. If @grzesiek2010 or one of the reviewers had filed an issue for opendatakit/collect@3cca371, we might have better remembered that the behavior changed.
- do a little more root cause analysis before filing an issue like https://github.com/XLSForm/pyxform/issues/270. Maybe @yanokwa could have noticed then that the limitation was no longer valid.

At the end of the day, there are a lot of moving pieces here so not sure we could have done a whole lot better but I think it's helpful to think about what went wrong.

PR #290, opened 2019-04-30
Tests that hints are displayed within repeats during conversion

WinnyTroy:
This PR closes issue #73 and presumably as well #74

Issue #74, opened 2016-07-26
Hint text for groups is ignored


spatialdude:
Hint text specified for groups is ignore instead of creating a &lt;hint&gt; element.

This appears to be the same issue as #73, so resolving this issue may also resolve #73

PR #358, opened 2019-09-09
Fix spelling and update failing tests

yanokwa:
Mostly whitespace and root node name changes, but there are two commits that need close consideration. 

* 9a934535106703c3e2ffdd358670c1b30b311bda - Confirm this is the correct syntax for external instances. @lognaturel, can you sanity check this?

* d885690ecb1ab0008a0ab18f6b5034b1922e0d83 - Confirm that groups and repeats shouldn't display hints. It's not in the spec (https://opendatakit.github.io/xforms-spec/#groups and https://opendatakit.github.io/xforms-spec/#repeats) and pyxform currently doesn't output hints. Sanity check, @MartijnR?

### DIFF
--- a/pyxform/tests_v1/test_fields.py
+++ b/pyxform/tests_v1/test_fields.py
@@ -100,7 +100,7 @@ class FieldsTests(PyxformTestCase):
             """
 
         expected = """
-            <select1 ref="/pyxform_autotestname/S1">
+    <select1 ref="/pyxform_autotestname/S1">
       <label>s1</label>
       <item>
         <label>a</label>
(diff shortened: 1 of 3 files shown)


lognaturel:
> Confirm this is the correct syntax for external instances. @lognaturel, can you sanity check this?

`select_one_external` is NOT for external secondary instances, it's for "fast external itemsets" (http://xlsform.org/en/#external-selects). So yes, the odd `query` syntax is expected for that feature.

I've never heard of group-level hints. Collect has no way of displaying these.

lognaturel:
Confirmed that group-level hints have never been output. https://github.com/XLSForm/pyxform/pull/290 added the test but didn't actually do anything to output the hints.

I think that if there's a desire to address #74, the first step is to have an XForms spec discussion because there is currently no mention of group-level hints at https://opendatakit.github.io/xforms-spec/#groups.

PR #364, opened 2019-09-10
WIP: Make it possible to go from select_one to select_multiple

yanokwa:


tiritea:
To follow up from now closed https://github.com/XLSForm/pyxform/pull/344...

FYI, I feel quite strongly that we can in no way restrict spaces from select1 responses. There is a very compelling usecase - standard responses - that relies heaving on this. Specifically, presenting the user a list of standardized responses, along with an 'other' where they can alternatively enter their own text response [actually I support open selections for this, not the `or_other` XLSForm hack].

It is also worth noting that, IMHO, supporting migrating a select1 to a select-multi is a bit of an edge case... A select1 and select-multi question are syntactically and sematically completely different question types. One return a _singleton_, the other a _list_ (this is as different as a geopoint vs geotrace). Further, a select1 has a strict notion of mutual-exclusivity, whereas a select-multi is more 'any that apply' which can include some level of redundancy and/or overlap. So in my view changing from one to the other results in a quite new, and incompatible, form definition.

So fundamentally restricting the usecase which a select1 can support, just to accommodate the possibility that the form writer might have made a mistake, and might want to change their form at a some later date, doesn't make sense to me. Its going to end up being a semantically different, incompatible form in any case; if you change question types in your form then you are going to have to update everything depends upon it, in this case including choice values.


I guess I'm OK with a warning, although I do feel its unnecessary - there's nothing wrong with having spaces in select1 values. But that's as far as we should go to accommodating this edge case IMO.

lognaturel:
I think it's clear that this change can't be made. I've opened #370 to discuss the possibility of a soft warning (or not).

Issue #157, opened 2017-10-19
Sometimes users don't apply consistent language tags across all translatable columns


yanokwa:
If you have no `default_language`, but have label/hint with language (`label::French`) and media with no language (`media::image`), you'll get unexpected behavior. That is, your labels will not be in French, they will be in the default language.

This was discovered here: https://forum.opendatakit.org/t/adding-media-column-results-in-blank-screen/10237

KeynesYouDigIt:
Ill take a crack at this one unless theres a better one. PR forthcoming!

KeynesYouDigIt:
Sorry I am just getting back to this, anyone know where exactly default_language is set? I am roving around the Survey class in survey.py but cant seem to pick out how its calculated and how to set it to null for tests. Is in in excel or the underlying OS that our code looks for this?

KeynesYouDigIt:
OK - I see a default lang column in the solution excel sheet, now I have to simulate that being empty in a test.

KeynesYouDigIt:
Closing in I hope, sorry for the delays :/

lognaturel:
If @KeynesYouDigIt or someone else wants to take another look at this, I think the first step should be to define what the warning text should be and enumerate all the cases that need to be covered by tests.

One case that's easy to forget is translations in choice lists. Choice list labels can also be translated so can cause this problem as well. See a test at https://github.com/XLSForm/pyxform/compare/master...lognaturel:pyxform-355 that should produce a warning because the survey sheet has labels in the language `English (en)` but the choices sheet only has labels in the default language.

KeynesYouDigIt:
Finally crawling my butt back to this one, let me know if its not a priority anymore for some reason.

KeynesYouDigIt:
The warning messages we had before were confusing, but was the idea right? Something like

> For question {q}, we detect an untranslated {hint, media, choice-list}. It might be a good idea to set the default language.

lognaturel:
Shoot, I missed these messages, @KeynesYouDigIt. I don't think warning for every question addresses this issue. I've written more on [this forum post](https://forum.getodk.org/t/missing-default-language-warning/25732/6?u=ln)

> The ideal would be to identify all localizable columns in the form and all languages in the form and make sure that every localizable column exists in every language.

> I believe the localizable column types are label, hint, media::image, media::video, media::audio but it would be good to double-check that in the code.

[...]

PR #365, opened 2019-09-11
Revert attempt at closing #157

### LINKED ISSUE #355: Conversion fails if label has language and choice_filter is used
This is a regression with v0.15.0 and I suspect the problem is https://github.com/XLSForm/pyxform/pull/287.

Here's the form in question.
[test.xlsx](https://github.com/XLSForm/pyxform/files/3560905/test.xlsx)

**Fails**
If I have a question with the proper label and a choice_filter, it fails.

Screen Shot 2019-08-30 at 10 20 05 AM

```
  File "/usr/local/bin/xls2xform", line 10, in <module>
    sys.exit(main_cli())
  File "/usr/local/lib/python2.7/site-packages/pyxform/xls2xform.py", line 175, in main_cli
    enketo=args.enketo_validate,
  File "/usr/local/lib/python2.7/site-packages/pyxform/xls2xform.py", line 48, in xls2xform_convert
    enketo=enketo,
  File "/usr/local/lib/python2.7/site-packages/pyxform/survey.py", line 899, in print_xform_to_file
    raise error
IndexError: list index out of range
```

**Works**

If I have I change to choice-filter `choice-filter` or `label::English (en)`, then it works.

Screen Shot 2019-08-30 at 10 20 27 AM

Screen Shot 2019-08-30 at 10 23 06 AM

lognaturel:
Reverts XLSForm/pyxform#287, Fixes #355

#287 introduced a crash in the common case of having translations in the survey sheet, a choice filter, but no translations in the choices sheet. I suspect it crashes in other cases.

As described in https://github.com/XLSForm/pyxform/issues/355#issuecomment-530496439, the error messages do not provide the most helpful user experience. They display XPath paths for question names and in some cases show the wrong column name (see `long` instead of `label` in the test example).

I hate to revert, @KeynesYouDigIt, @ukanga, but I think that this feature needs to be carefully designed. A lot of users have missing translations, willingly or not, so it's important that they see helpful messages. There are also various edge cases here and I suspect the identified crash is not the only one.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -178,7 +178,7 @@ class Survey(Section):
 
         return NSMAP
 
-    def xml(self, warnings=None):
+    def xml(self):
         """
         calls necessary preparation methods, then returns the xml.
         """
@@ -191,7 +191,7 @@ class Survey(Section):
 
         return node(
             "h:html",
-            node("h:head", node("h:title", self.title), self.xml_model(warnings)),
+            node("h:head", node("h:title", self.title), self.xml_model()),
             node("h:body", *self.xml_control(), **body_kwargs),
             **nsmap
         )
@@ -432,13 +432,13 @@ class Survey(Section):
                 yield i.instance
             seen[i.name] = i
 
-    def xml_model(self, warnings=None):
+    def xml_model(self):
         """
         Generate the xform <model> element
         """
         self._setup_translations()
         self._setup_media()
-        self._add_empty_translations(warnings)
+        self._add_empty_translations()
 
         model_children = []
         if self._translations:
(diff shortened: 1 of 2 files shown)


lognaturel:
Thanks, @ukanga. 🙏

KeynesYouDigIt:
Such a bummer :( ill add a test case and fix it. Any log or error message on the crash?

lognaturel:
@KeynesYouDigIt thanks for coming back to it! Take a look at #355 for a stack trace. Like I said there, I think the first step should be to design a user experience for #157. It's really important to have error messages that actually help the user identify where the problem is to fix it.

PR #368, opened 2019-09-19
Default to 'data' when filename is invalid.

nribeka:
Addresses #130

When the xls filename start with number, and the user doesn't set the 'name' in the settings sheet, the pyxform will complain because it will try to use the filename as an xml tag.

This will default the xml tag to data when the filename is not a valid xml tag.

### DIFF
--- a/pyxform/tests_v1/pyxform_test_case.py
+++ b/pyxform/tests_v1/pyxform_test_case.py
@@ -78,7 +78,8 @@ class PyxformMarkdown(object):
         # ideally, when all these tests are working, this would be
         # refactored as well
         survey = create_survey_element_from_dict(imported_survey_json)
-        survey.name = kwargs.get("name")
+        if not kwargs.get("skip_name"):
+            survey.name = kwargs.get("name")
         survey.title = kwargs.get("title")
         survey.id_string = kwargs.get("id_string")
 
--- a/pyxform/tests_v1/test_form_name.py
+++ b/pyxform/tests_v1/test_form_name.py
@@ -26,3 +26,68 @@ class FormNameTest(PyxformTestCase):
                 '</input>',
             ],
         )
+
+    def test_default_to_data_when_no_name(self):
+        """
+        Test no form_name and skipping name field, default to 'data'.
+        """
+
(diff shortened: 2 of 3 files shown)


nribeka:
I'm still looking at how to unit tests this. I might need to move the changes to ```workbook_to_json``` and set form_name to data there. I think this would allow me to update the base unit test easier.

nribeka:
| XLS filename | The name in settings | The expected name tag in the model  |
|---|---|---|
| not valid | not being set | data |
| not valid | being set | name from the settings |
| valid | not being set  | filename |
| valid | being set | name from the settings |

lognaturel:
One of the big challenges with #152 is that it makes all of its changes in a single commit and changes the tests along with the code change. What would be important is that the changes to the existing tests are made as narrowly as possible and are made BEFORE any code changes. What I'm imagining is that the first commit would ensure that the root node name is explicitly set for every test. All the tests should pass with that commit. For the old style XLS-reading tests, the goal would be not to change the expected XML.

Then we can be convinced that a subsequent code change is safe.

lognaturel:
I also don't think 7a6839b is the expected change -- the root node name should always be `data` unless it is explicitly overwritten by the user. The filename should never be used.

nribeka:
@lognaturel,

I didn't realize we want to ignore the filename altogether.

So if we ignore the filename altogether, we can just set the default for form_name field to 'data'. This will essentially set the tag to 'data' all the time and only gets overwritten if the user specified the name field in the settings. Is this the changes that you're looking for? And you want to show that defaulting everyone to 'data' all the time will not break existing unit tests?

lognaturel:
We talked in Slack but for completeness --

> we can just set the default for form_name parameter in workbook_to_json to data

That sounds right and I think that's what #152 did.

> And you want to show that defaulting everyone to 'data' all the time will not break existing unit tests?

Right, so the first step is to make sure that every test either is happy with the root node name changing or explicitly sets the root node name to whatever its output implies. If the tests still pass with those changes, then we're in a safe place to make the change.

nribeka:
Closing this in favor of #375

PR #375, opened 2019-09-23
Updating default form_name to data instead of using filename.

### LINKED ISSUE #130: Convert files with leading numbers in their name

nribeka:
Fix #130 

Defaulting the tag name to 'data' instead of using filename. This will alleviate issue when filename is not valid for tag name.

The tag name can be changed by overriding it using the 'name' column in the setting sheet.

More conversation in https://github.com/XLSForm/pyxform/pull/368.


pyxform/tests_v1/pyxform_test_case.py
```diff
@@ -78,7 +78,8 @@ def _ss_structure_to_pyxform_survey(ss_structure, kwargs):
         # ideally, when all these tests are working, this would be
         # refactored as well
         survey = create_survey_element_from_dict(imported_survey_json)
-        survey.name = kwargs.get("name")
+        if not kwargs.get("skip_name"):
```
nribeka: I needed to add the skip here because the test will always set the form name. I will never get the ```data``` tag without this.

pyxform/tests_v1/test_form_name.py
```diff
@@ -0,0 +1,93 @@
...
+        Test no form_name and skipping name field, default to 'data'.
+        """
+
+        self.assertPyxformXform(
+            md="""
+               | survey |      |      |           |
+               |        | type | name | label     |
+               |        | text | city | City Name |
+               """,
+            skip_name=True,
```
nribeka: Skipping the name will product ```data``` (default tag name).

pyxform/tests_v1/test_form_name.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+
+        self.assertPyxformXform(
+            md="""
+               | survey |      |      |           |
+               |        | type | name | label     |
+               |        | text | city | City Name |
+               """,
+            id_string='some-id',
+            instance__contains=['<pyxform_autotestname id="some-id">'],
```
nribeka: Not skipping name will produce the default tag name coming from the parent unit test class.

pyxform/tests_v1/test_form_name.py
```diff
@@ -0,0 +1,93 @@
...
+        Test no form_name and setting name field, should use name field.
+        """
+
+        self.assertPyxformXform(
+            md="""
+               | survey |      |      |           |
+               |        | type | name | label     |
+               |        | text | city | City Name |
+               """,
+            name='some-name',
```
nribeka: Setting the name will override the default ```data```.

pyxform/xls2json.py
```diff
@@ -1304,7 +1304,7 @@ def get_filename(path):
 
 
 def parse_file_to_json(
-    path, default_name=None, default_language="default", warnings=None, file_object=None
+    path, default_name='data', default_language="default", warnings=None, file_object=None
```
nribeka: Defaulting the default_name when parsing xls file to 'data'. I'm wondering if I should remove line 1315-1316 too. I think (with my limited python skill) they're dead code.

```
    if default_name is None:
        default_name = unicode(get_filename(path))
```

The ```default_name``` will never be ```None```. The value will be either ```data``` or whatever passed from the calling function.

lognaturel: This doesn't seem Python-specific! The desired behavior is for the root node name to always be `data` unless the user explicitly specifies an alternative. So there should no longer be code to ever set the name from the file name, right?

Are you absolutely sure that both `workbook_to_json` should nave a default `form_name` of `data`? I'm less sure about where that's used.

pyxform/tests_v1/pyxform_test_case.py
lognaturel: I'm not thrilled by this approach because it means adding a code path only for testing. I'm wondering whether it might be appropriate to have an "old style" test for this particular change since the code that is changed and needs to be exercised is in `xls2json.parse_file_to_json`. @ukanga might be best position to answer that. I know he doesn't want to use those old style tests anymore for testing the actual conversion process which I agree strongly with. In this case the goal is to test a change to how the XLS file itself is treated, though.

nribeka: Yeah, the old tests I think will better suited for this case. I needed to add the skipping part because the test will set the form name when I don't pass name field in the markdown setting part.

I can add one old unit test if everyone is OK with that? I was under the impression that we are not touching the old unit tests.

pyxform/xls2json.py
nribeka: The change in ```workbook_to_json``` is related to the unit test. If I do skip form name and the parent unit test doesn't set the name, the form name becomes ```'u' None```.

Line 339 set the value:

nribeka: I think the ```workbook_to_json``` still need to default the form name to 'data'. This will ensure the base of the tag name will always be data. Leaving it to None will open the future possibility of having the tag name 'None'.

pyxform/tests_v1/pyxform_test_case.py
ukanga: I agree adding a `skip_name` is unnecessary here. I don't think we need to change this one at all. We could change all tests that refer to `pyxform_autotestname` such that the paths only has `/data/field-name-path`. Perhaps we could even default it to data i.e.
```python
survey.name = kwargs.get("name", "data")
```
We can still use the auto-generated `id_string` and `title`.

I am very okay with tests in `pyxform/tests_v1` being updated to reflect the new default. My main opposition was on changes to fixtures on the old `pyxform/tests/`. The one XML fixture here I understand why that was unavoidable.

pyxform/xls2json.py
```diff
@@ -1304,16 +1304,14 @@ def get_filename(path):
...
 ):
     """
     A wrapper for workbook_to_json
     """
     if warnings is None:
         warnings = []
     workbook_dict = parse_file_to_workbook_dict(path, file_object)
```
nribeka: Dead code path removed.

pyxform/tests_v1/pyxform_test_case.py
nribeka: Base unit test updated. Reverted some code related to the skipping part. I think this is ready for review again.

pyxform/tests_v1/test_form_name.py
```diff
@@ -0,0 +1,67 @@
...
+            """
+            | survey |      |      |           |
+            |        | type | name | label     |
+            |        | text | city | City Name |
+            """,
+            {},
+            autoname=False,
+        )
+
+        self.assertEqual(survey.name, unicode("data"))
```
lognaturel: The XML should be the same as for the test below, right? I think it'd be good to have that be explicitl

pyxform/tests_v1/test_form_name.py
lognaturel: Maybe it'd be good to do this with a settings sheet to even more closely match the code paths used by real users?

pyxform/xls2json.py
```diff
@@ -1471,7 +1469,11 @@ def _organize_renames(self):
         path = sys.argv[1]
 
     warnings = []
-    json_dict = parse_file_to_json(path, warnings=warnings)
+
+    directory, filename = os.path.split(path)
```
lognaturel: Why is this still needed?

nribeka: I removed the code.

pyxform/tests_v1/test_form_name.py
nribeka: Added explicit xml comparison. I needed to manually set the id_string and title because the parent class override them with None even though I pass it using the settings tab. This part I think have issue.

nribeka:
```
        survey = create_survey_element_from_dict(imported_survey_json)
        survey.name = kwargs.get("name", "data")
        survey.title = kwargs.get("title")
        survey.id_string = kwargs.get("id_string")
```

The code above will create survey object based on the parsed markdown string. But after the survey is created, the next line will overwrite the ```id_string``` and ```title``` with whatever value from the kwargs. So if you don't set the values in the kwargs, it will set it to ```None```.

pyxform/tests_v1/test_form_name.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        Test using data as the name of the form which will generate <data />.
+        """
+        self.assertPyxformXform(
+            md="""
+               | survey |      |      |           |
+               |        | type | name | label     |
+               |        | text | city | City Name |
+               """,
+            name='data',
```
nribeka: @lognaturel,

I can't use the settings tab here too. If I add settings tab in the markdown, it will correctly create the survey using the settings tab. But the next line will override it with the values from the kwargs. And the values in the kwargs is set in ```_autoname_inputs```.

This unit test will show this issue:
```
    def test_should_use_setting_tab(self):
        """
        Test using data as the name of the form which will generate <data />.
        """
        self.assertPyxformXform(
            md="""
               | survey   |           |      |           |
               |          | type      | name | label     |
               |          | text      | city | City Name |
               | settings |           |      |           |
               |          | id_string | name |
               |          | some-id   | data |
               """,
            debug=True,
            instance__contains=['<data id="some-id">'],
            model__contains=['<bind nodeset="/data/city" type="string"/>'],
            xml__contains=[
                '<input ref="/data/city">',
                '<label>City Name</label>',
                '</input>',
            ],
        )
```

nribeka: The above test will fail because it will generate this xml file (tag name is not data):
```
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <h:head>
    <h:title>pyxform_autotesttitle</h:title>
    <model>
      <instance>
        <pyxform_autotestname id="pyxform_autotest_id_string">
          <city/>
          <meta>
            <instanceID/>
          </meta>
        </pyxform_autotestname>
      </instance>
      <bind nodeset="/pyxform_autotestname/city" type="string"/>
      <bind jr:preload="uid" nodeset="/pyxform_autotestname/meta/instanceID" readonly="true()" type="string"/>
    </model>
  </h:head>
  <h:body>
    <input ref="/pyxform_autotestname/city">
      <label>City Name</label>
    </input>
  </h:body>
</h:html>
```

pyxform/tests_v1/test_form_name.py
nribeka: Setting sheet will not work. See my other comments.

pyxform/tests_v1/test_form_name.py
ukanga: It seems the way to test the default_name="data" is by testing `parse_file_to_json()` function directly. The markdown implementation may need to be updated to match the same but it will be fine to handle that outside of this PR.

ukanga:
@nribeka If you can fix the black issues, this should be good to merge.

PR #376, opened 2019-09-26
Updating choices for single language.

### LINKED ISSUE #285: Don't generate itext block if only one language is defined
The `itext` block can include translations for form elements (see [spec](https://opendatakit.github.io/xforms-spec/#languages)). `pyxform` always adds this level of indirection for cascading selects, even if the XLSForm only defines labels/hints for a single language. This isn't a big deal for small secondary instances but can make the XML much bigger for large ones.

Related to #32 and might be good to address while in the same brain space.

nribeka:
Fix #285.

This will inline the label when there's only 1 language for the survey instead of referencing the itext fields.


lognaturel:
The approach here is to look at the first choice in a list and see whether its label is represented as a dict or a simple value, right? This seems fine to me! I'm guessing you looked at doing it with the header row of the choices sheet and that there's no way to do it with that?

I think the conditionals at https://github.com/XLSForm/pyxform/pull/376/files#diff-642ca628a5111f8d9ff1314acb61ceadL550 would also go away.

nribeka:
Yes, this is looking at the label whether it is dictionary type or simple value.

nribeka:
I can do checking on the row header and that's the other approach that I mentioned in the ticket. What that entails:
* I need to introduce a new field in the intermediate json_dict (probably gonna call it "multi_language" of boolean type).
* Update the survey object and add new field (same probably).
* Use that field to conditionally generate the xml document.

What do you think?

lognaturel:
I don't see a big benefit to having the property exposed in the json_dict. The checks are cheap and I think they're pretty easy to understand.

I see also that on the issue you asked about modifying the expected XML. I don't think there's going to be any way around that. Hopefully there aren't too many tests affected and the diff will be easy to review.

nribeka:
Kewl then @lognaturel. Going to fix the expected xml next then. There are 5 xml files I think. Not bad.

nribeka:
Unit tests fixed and I added a few tests. Ready for some eyes to check it out.

pyxform/tests_v1/test_inline_translations.py
```diff
@@ -0,0 +1,63 @@
...
+"""
+Test setting form name to data.
+"""
+from pyxform.tests_v1.pyxform_test_case import PyxformTestCase
+
+class InlineTranslationsTest(PyxformTestCase):
+
+    def test_inline_translations(self):
+        """
+        Test using data as the name of the form which will generate <data />.
```
lognaturel: This comment isn't relevant

pyxform/tests_v1/test_inline_translations.py
```diff
@@ -0,0 +1,63 @@
...
+                             '<itextId>static_instance-states-0</itextId>',
+                             '<itextId>static_instance-states-1</itextId>',
+                             '<itextId>static_instance-states-2</itextId>'],
+            xml__contains=['<label ref="label"/>'],
+            xml__excludes=['<label ref="jr:itext(itextId)"/>'],
+        )
+
+    def test_multiple_translations(self):
+        """
+        Test using data as the name of the form which will generate <data />.
```
lognaturel: Comment should relate to the test. This one is actually a little odd because the reason that translations are generated is because the survey sheet only has a default language (`label` without a suffix) and the choices sheet only has English (en). I think that's ok but to be more explicit, consider adding a language suffix.

What happens if both have the `::English(en)` suffix? No translations are generated, right? Or is it only the case if the settings sheet explicitly sets English(en) as the default? Since this case is a little unclear, I think it's worth having a test for.

lognaturel:
It looks like the tests fail on some Python version and that black fails on others. Would be good to look into.

pyxform/tests_v1/test_inline_translations.py
nribeka: Updated the comment. It was a copy-pasta error.

pyxform/tests_v1/test_inline_translations.py
nribeka: I think since I'm checking it whether it's a dictionary or a straight string, whenever there's ```::English(en)```, it will think the survey is multi-language because it will actually generate a dictionary instead of the regular label.

Do you want me to inline this as well? This will require an update in the approach to test for multi-language or not. Instead of checking the first element to see if it's a dictionary or not, i need to count each element in the ```choices``` dictionary. Multi language will have more than 1 elements in the ```choices```.

nribeka:
Fixed for 3.5. I ended up need to sort the label and the name to make sure they're consistent on both 2.7 and 3.5.

I tried to fix the black issue earlier, but failed miserably hahaha ...

PR #377, opened 2019-09-30
Running black to fix formatting.

nribeka:
The PR should fix the black issue in the codebase.


nribeka:
I tried to run the black and it ended up not working on lower python, but working on the higher python version. So great.

PR #378, opened 2019-09-30
Disabling quiet to see errors from black.

### LINKED ISSUE #379: Circle CI failing on Python 3.6 and 3.7.
The circle CI configuration will execute black on Python 3.6 and 3.7 to ensure formatting.

```
  if ! type /home/circleci/.local/bin/black > /dev/null; then
    echo "Black is not supported in this python version :("
  else
    /home/circleci/.local/bin/black --target-version=py27 pyxform --check --quiet || (echo 'The source code could use a bit more black.' && exit 1)
  fi
```

The check will execute the ```if``` branch on Python 2.7 and Python 3.5. The check will execute the ```else``` branch on Python 3.6 and 3.7.

The issue with the check (I think) is because of the ```quiet``` flag being passed to black. I think on circle ci, the first part return false (it is not on my laptop). And that's the reason it's always executing the echo and exit 1 part.

My PR is updating the check to search for text in the output of executing black.

nribeka:
Trying out to fix the black. Disabling the quiet flag and checking the output text.

Fix #379.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,11 @@ jobs:
             if ! type /home/circleci/.local/bin/black > /dev/null; then
                 echo "Black is not supported in this python version :("
             else
-                /home/circleci/.local/bin/black --target-version=py27 pyxform --check || (echo 'The source code could use a bit more black.' && exit 1)
+                output="$(/home/circleci/.local/bin/black --target-version=py27 pyxform --check  2>&1 >/dev/null)"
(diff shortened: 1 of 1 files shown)


yanokwa:
I'm not sure what problem this fixes. Is black failing? If so, can you file an issue or clarify the need in the PR.

nribeka:
@yanokwa added issue to try to explain the problem that I am seeing.

PR #380, opened 2019-10-01
Adding repeat instance to the model for repeat

### LINKED ISSUE #182: Don't always include jr:template for repeats
Currently `jr:template` is always added to repeats but it is not generally what users want. Typically users don't provide defaults for questions in the repeat.

I think the ideal would be to only include `jr:template` if one or more of the questions in the repeat has a default. Another option would be to add something like `template=yes|no` to the `parameters` column.

nribeka:
Fix #182.

This will add instance for repeat section.


nribeka:
@lognaturel 2 failed tests. The tests is counting number of elements in the model which obviously going to fail because we're adding new elements to the model. I'm just wondering if this is what we're looking for.

pyxform/section.py
```diff
@@ -40,12 +42,30 @@ def xml_instance(self, **kwargs):
...
             elif isinstance(child, ExternalInstance):
                 continue
+            else:
+                if isinstance(child, RepeatingSection) and not append_template:
+                    append_template = not append_template
+                    repeating_template = child.generate_repeating_template()
+                result.appendChild(child.xml_instance(append_template=append_template))
+            if append_template and repeating_template:
+                append_template = not append_template
+                result.insertBefore(repeating_template, result._get_lastChild())
```
nribeka: ```result._get_lastChild()``` vs ```result.childNodes[-1]```?

pyxform/section.py
```diff
@@ -104,11 +120,11 @@ def xml_control(self):
 
     # I'm anal about matching function signatures when overriding a function,
```
lognaturel: This comment is no longer relevant, right?

nribeka: Remove it?

pyxform/tests_v1/test_repeat_template.py
```diff
@@ -0,0 +1,141 @@
...
+        repeat_b_template = '<repeat_b jr:template="">'
+        self.assertEqual(1, survey_xml.count(repeat_b_template))
+
+        section_instance = "<section>"
+        self.assertEqual(1, survey_xml.count(section_instance))
+        repeat_a_instance = "<repeat_a>"
+        self.assertEqual(1, survey_xml.count(repeat_a_instance))
+        repeat_b_instance = "<repeat_b>"
+        self.assertEqual(1, survey_xml.count(repeat_b_instance))
+
```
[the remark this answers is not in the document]
nribeka: Ah I didn't know about that. I just assume to use the equality assertion 👯‍♂

lognaturel:
I talked to @seadowg about this who pointed out that it's quite likely form designers are currently using repeats in Collect to represent 0 or more of a thing (rather than 1 or more). For example, if the desire is to collect information about children or siblings, it's quite likely that the flow of getting prompted initially about whether or not to add a repeat and potentially saying no is actually the right flow.

I don't know exactly what to do about this now. Marking as `needs discussion`.

nribeka:
This would also need input from @MartijnR.

MartijnR:
Enketo creates the first repeat automatically (and has been since 2012). In the past users have expressed a desire to have 0, but I believe this has been addressed sufficiently by introducing support for repeat-count and allowing the first repeat to be deleted manually (which initially wasn't possible). Since then, here have been no issues reported. If it is important to (be able to) start with 0, I agree we need to address that though.

craigappl:
@ukanga , note that we need to review the comments from today's session. [Link](https://github.com/XLSForm/pyxform/issues/182#issuecomment-541444640)

nribeka:
@lognaturel or @ukanga do you guys aware of way to do warning assertions in the v1? There's an old test that assert the warning, but it's the old unit test.

pyxform/section.py
lognaturel: This was about overriding `xml_instance` and the name has now changed, right? So yes, the comment should be removed. It's also looking to me like neither this method nor `generate_repeating_template` need to have a `**kwargs` parameter, is that right?

pyxform/xls2json.py
```diff
@@ -859,6 +860,20 @@ def workbook_to_json(
...
+                if (
+                    control_type == constants.REPEAT
+                    and not repeat_behavior_warning_added
+                ):
+                    warnings.append(
+                        "Repeat behavior has changed. Previously, some clients like "
+                        "ODK Collect prompted users to add the first repeat. Now, "
+                        "the user will only be prompted to add repeats after the first "
+                        "one. Representing 0 repetitions will require changing the form "
+                        "design. Read more at http://xlsform.org/en/#repeats."
```
lognaturel: You can now change this to http://xlsform.org#representing-zero-repeats

lognaturel:
I think you can see an example of verifying warnings with v1 tests in the `test_language_warnings` test.

nribeka:
@lognaturel I just checked that unit tests and the warning testing will not work on this use case because the warning is generated when we parse the xls to json, but the v1 tests doesn't use the same function. It's using custom function to convert the md to json.

pyxform/xls2json.py
nribeka: Updated the link to the above.

Issue #370, opened 2019-09-20
Warn when a choice name for a select_one has a space in it


lognaturel:
Original discussion about spaces in select ones at https://github.com/opendatakit/javarosa/issues/263#issuecomment-369019465

https://github.com/XLSForm/pyxform/pull/344 explored preventing spaces in select_one choice names but this was deemed infeasible because:
- forms with spaces in select_one choice names already exist. Forcing a change in the choice names when updating those forms with pyxform makes data analysis really difficult because there will be two different choice names for the same entity recorded at different times. It's also very hard to communicate about this change to users.
- form designers use data sets from other systems that may have spaces in the names. For select_ones, it's ok for those spaces to exist.

Some warning text has been proposed in preparation for an eventual change:
- "Warning: a select_one option name contains a space. We strongly advice to not use spaces. They will be disallowed in the future."
- "Warning: a select_one option name contains a space (option 1, option 2, option 3, ...). To simplify analysis and make select ones convertible to select multiples, spaces in option names will be disallowed in the future."

For the reasons listed above, I feel like it's unlikely such a change would ever happen. It may still be worth providing a warning such as:
- "Warning: a select_one option name contains a space (option 1, option 2, option 3, ...). Removing the space will allow you to use the same list for a select_multiple and may simplify analysis."
- "Warning: a select_one option name contains a space (option 1, option 2, option 3, ...). See xlsform.org/newsection to learn more."

CC @yanokwa @tiritea @ukanga @MartijnR @tedrick

tiritea:
>  I feel like it's unlikely such a change would ever happen

I agree (would and probably _should_ never... )

I might even tone it down further, eg

- "Note: a select_one option name contains a space (option 1, option 2, option 3, ...). Removing the space will also allow you to use this list for a select_multiple."

A 'Warning' implies you did something wrong, but you haven't. This is more like an FYI.

lognaturel:
Agreed, @tiritea. I've modified the top post to reflect that.

MartijnR:
Agreed. Thanks!

PR #381, opened 2019-10-02
Add support for choice from previous repeat answers

### LINKED ISSUE #38: Itemset from previous answers - select one from ${node}

ukanga:
Fix #38

### DIFF
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -180,9 +180,21 @@ class MultipleChoiceQuestion(Question):
                 node("label", ref=itemset_label_ref),
             ]
             result.appendChild(node("itemset", *itemset_children, nodeset=nodeset))
+        elif not self["children"] and self["list_name"]:
+            list_name = survey.insert_xpaths(self["list_name"], self).strip()
+            path = list_name.split("/")
+            depth = len(path)
+            if depth > 2:
+                name = path[-1]
+                nodeset = "/".join(
+                    path[:-2] + [path[-2] + "[string-length(./" + name + ") > 0]"]
+                )
+                itemset_children = [node("value", ref=name), node("label", ref=name)]
+                result.appendChild(node("itemset", *itemset_children, nodeset=nodeset))
         else:
-            for n in [o.xml() for o in self.children]:
-                result.appendChild(n)
+            for child in self.children:
+                result.appendChild(child.xml())
+
         return result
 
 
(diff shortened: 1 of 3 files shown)


pyxform/question.py
```diff
@@ -180,9 +180,21 @@ def xml_control(self):
...
+        elif not self["children"] and self["list_name"]:
+            list_name = survey.insert_xpaths(self["list_name"], self).strip()
+            path = list_name.split("/")
+            depth = len(path)
+            if depth > 2:
+                name = path[-1]
+                nodeset = "/".join(
+                    path[:-2] + [path[-2] + "[string-length(./" + name + ") > 0]"]
+                )
+                itemset_children = [node("value", ref=name), node("label", ref=name)]
```
lognaturel: This is related to https://github.com/XLSForm/pyxform/issues/370 because the values will be user-generated and could contain spaces. If this turns out to be a popular feature, we may want to allow users to specify `value` and `label` separately. I don't think this needs to be done now but wanted to call it out in case that changes things for anyone.

pyxform/tests_v1/test_repeat.py
```diff
@@ -266,3 +266,26 @@ def test_hints_are_present_within_groups(self):
...
+        |         | select one ${name} | choice     | Choose name    |
+        | choices |                    |            |                |
+        |         | list name          | name       | label          |
+        |         | fruits             | banana     | Banana         |
+        |         | fruits             | mango      | Mango          |
+        """
+        self.assertPyxformXform(
+            md=xlsform_md,
+            xml__contains=[
+                '<itemset nodeset="/pyxform_autotestname/rep[string-length(./name) &gt; 0]">'
```
lognaturel: There shouldn't be a predicate.

DavisRayM:
Superseded by #472

PR #382, opened 2019-10-02
Updating validator error messages.

### LINKED ISSUE #39: Bad XLSForm calculation breaks code, and why are errors not dealt with when using xls2xform.py as main?
Wonderful work so far on pyxform! I use pyxform to automate xform creation for my project. I have noticed that a bad calculate in XLSForm (a survey with a single line, e.g. `{type: 'calculate', name: 'badcalc', label: 'This calculation fails', calculation:'because it doesn't know what to do'}`) produces an error such as 

```
  File "/path/to/package/pyxform/pyxform/xls2xform.py", line 65, in <module>
    warnings = xls2xform_convert(args.path_to_XLSForm, args.output_path)
  File "/path/to/package/pyxform/pyxform/xls2xform.py", line 22, in xls2xform_convert
     survey.print_xform_to_file(xform_path, validate=True, warnings=warnings)
  File "/path/to/package/pyxform/pyxform/survey.py", line 446, in print_xform_to_file
    warnings.extend(check_xform(path))
  File "/path/to/package/pyxform/pyxform/odk_validate/__init__.py", line 112, in check_xform
     m = 'ODK Validate Errors:\n' + _cleanup_errors(stderr)
  File "/path/to/package/pyxform/pyxform/odk_validate/__init__.py", line 86, in _cleanup_errors
return u'\n'.join(k)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xcd in position 170: ordinal not in range(128)
```

This is not very informative. The problem is that the current code keeps python's place indicator of where the error is. This cannot be converted to unicode implicitly, so an error is thrown. I have made a fix to the code (locally), and now the error message says 

```
ODK Validate Errors:
org.javarosa.xform.parse.XFormParseException: Invalid calculate for the bind attached to "${badcalc}" : Couldn't understand the expression starting at this point: because͎it doesn't know... in expression because it doesn't know what to do

>> XForm is invalid. See above for the errors.

Result: Invalid
```

But now the problem is that an ODK Validate error raises an unhandled `Exception`. A change to the main method (read: inside the `if __name__ == '__main__':` suite) in `xls2xform.py` to handle these exceptions would be pretty easy.

Nafundi's XLSForm Offline handles conversion errors, but does not handle the bad calculate. I would have thought that Nafundi's changes would make it back into the code, but I suppose not.

I would be happy to contribute what I have done to the project. What is the process to contribute/submit pull requests? What about creating test cases or checking that my new code does not break the old code?

nribeka:
Fix #39 

This PR is based on closed PR #41 

This will handle errors and making sure the program is not showing ugly python stack trace when encountering errors when converting form.


pyxform/tests/xls2xform_tests.py
```diff
@@ -221,6 +225,27 @@ def test_xls2xform_convert_params_with_flags(
...
+            path_to_XLSForm=path_to_text_fixture("bad_calc.xlsx"),
+            output_path=None,
+            json=False,
+            skip_validate=True,
+            odk_validate=True,
+            enketo_validate=True,
+            pretty_print=True,
+        ),
+    )
+    def test_xls2xform_convert_throwing_odk_error(self, parser_mock_args):
```
nribeka: I needed to add this to the old tests because codecov complained that i make the test coverage dropped.

pyxform/tests/bug_tests.py
```diff
@@ -297,5 +298,14 @@ def test_xlsm_files_are_allowed(self):
         self.assertIsInstance(result, dict)
 
 
+class TestBadCalculation(unittest.TestCase):
```
nribeka: See reason below.

nribeka:
@lognaturel + @ukanga,

I think codecov is calculating the diff hit on the part where I change the spacing of the code on xls2xform too. I dunno how to increase the code coverage because that part is counted as removal.

ಠ_ಠ

pyxform/xls2xform.py
```diff
@@ -167,18 +168,26 @@ def main_cli():
...
+                xform_path=args.output_path,
+                validate=args.odk_validate,
+                pretty_print=args.pretty_print,
+                enketo=args.enketo_validate,
+            )
+            if len(warnings) > 0:
+                logger.warning("Warnings:")
+            for w in warnings:
+                logger.warning(w)
+            logger.info("Conversion complete!")
```
ukanga: Can we move this outside the try block? I don't think any of this throws any of the exceptions below. I think it is cleaner that way.

nribeka: Isn't that gonna make it print the ```Conversion complete``` all the time?

ukanga: It should be the same amount of time as it currently would. I expect that the only function throwing `EnvironmentError` and `ODKValidateError` is:
```python
warnings = xls2xform_convert(
                xlsform_path=args.path_to_XLSForm,
                xform_path=args.output_path,
                validate=args.odk_validate,
                pretty_print=args.pretty_print,
                enketo=args.enketo_validate,
            )
```
I think it is the norm to only have protected within the try except clause, hence I would propose:
```python
        try:
            warnings = xls2xform_convert(
                xlsform_path=args.path_to_XLSForm,
                xform_path=args.output_path,
                validate=args.odk_validate,
                pretty_print=args.pretty_print,
                enketo=args.enketo_validate,
            )
        except EnvironmentError as e:
            # Do not crash if 'java' not installed
            logger.error(e)
        except ODKValidateError as e:
            # Remove output file if there is an error
            os.remove(args.output_path)
            logger.error(e)
        else:
            if len(warnings) > 0:
                logger.warning("Warnings:")
            for w in warnings:
                logger.warning(w)
            logger.info("Conversion complete!")
```

nribeka: Ah I didn't know about that way of writing the try-catch. Let me update it.

nribeka:
@ukanga added the else block in the try-catch block.

nribeka:
Codecov wanting me to write more tests I guess.

lognaturel:
This looks good to me. I don't really care about the Codecov complaint and would be comfortable merging without addressing it. In fact, I might even go so far as to suggest removing the "old style" tests but I don't feel strongly about it.

I think all that's needed here is a last pass by @ukanga to see whether he likes how his feedback was incorporated.

PR #383, opened 2019-10-03
Allowing xml-internal question type.

### LINKED ISSUE #176: Add type=xml-internal and generate instance XML from choices sheet
As a form designer, I want to be able to specify static data sets that can be queried from the form. Currently this can be hacked with `selects` and `choice-filter` but it's not elegant (e.g. https://docs.google.com/spreadsheets/d/1mQGVWaJA4QQp3jtddMdcRSDjc1O-j_bB79GUnlj5k10/edit#gid=0 where we want to specify a lookup table without needing an external file). 

#107 added type `xlm-external` to specify an external XML instance that can be queried from within the primary instance.

`xml-internal` would work the same way but create an internal secondary instance. This is useful for specifying additional settings as described in #66 or including small, static data sets that are useful to do XPath queries on. 

type         | name |
----------|-------|
xml-internal  | mydata |

For this to be useful, pyxform should also generate the instance. This could be done from the existing choices sheet which has very similar behavior currently. If going this route, it might be necessary to change the rules that govern whether or not a column's values are included in an instance. Currently columns other than name/value and label are only included if there's a choice-filter that uses it. For more generic data sets `choice-filter` doesn't make sense. For example, the rule could be changed to something like if any row in this list specifies a value in the column, include it.

Alternately, another sheet with name something like `datasets` could be introduced. The benefit there is that `name` and `label` wouldn't need to be required and a different rule for including columns could easily be introduced.

Instance generation should be possible for `xml-external` as well (much like the itemsets.csv file is generated).

nribeka:
Fix #176.

This would allow question of type xml-internal.

### DIFF
--- a/pyxform/tests_v1/test_xml_internal.py
+++ b/pyxform/tests_v1/test_xml_internal.py
@@ -1,0 +1,103 @@
+# -*- coding: utf-8 -*-
+"""
+Test for the xml-internal question
+"""
+from pyxform.tests_v1.pyxform_test_case import PyxformTestCase
+
+
+class XmlInternalTest(PyxformTestCase):
+    def should_add_instances_for_xml_interna(self):
+        md = """
+            | survey  |                    |            |           |            |                   |
+            |         | type               | name       | label     | hint       |appearance         |
+            |         | xml-internal       | states     |           |            |                   |
+            |         | xml-internal       | yes_no     |           |            |                   |
+            | choices |                    |            |           |            |                   |
+            |         | list_name          | name       | label     |            |                   |
+            |         | yes_no             | yes        | Yes       |            |                   |
+            |         | states             | VA         | Virginia  |            |                   |
+            |         | states             | IN         | India     |            |                   |
+            |         | fruits             | apple      | Apple     |            |                   |
+            |         | fruits             | banana     | Banana    |            |                   |
+            """
+        survey = self.md_to_pyxform_survey(md_raw=md)
+        survey_xml = survey._to_pretty_xml()
+
+        self.assertContains(survey_xml, '<instance id="yes_no">', 1)
+        self.assertContains(survey_xml, '<instance id="states">', 1)
+        self.assertContains(survey_xml, '<instance id="fruits">', 1)
+
+        self.assertPyxformXform(
+            md=md,
+            model__contains=[
+                '<instance id="yes_no">',
+                '<instance id="states">',
+                '<instance id="fruits">',
+            ],
+        )
+
+    def should_add_single_instance_with_choice_filter(self):
+        md = """
+            | survey  |                    |            |           |               |                   |
+            |         | type               | name       | label     | choice_filter | appearance        |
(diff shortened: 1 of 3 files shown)


nribeka:
@lognaturel see also my comment in #176. Is this the expected behavior?

lognaturel:
@nribeka is this behavior on master or just on this branch? What happens if you actually include a select on the `fruits` list, is it inlined or added as an instance?

nribeka:
@lognaturel, this is in master.

As long as you add ```choice_filter``` it will add everything from the choices sheet as separate instance even though it's not being used by any question.

If you add select on fruits, then it will add it to the body and also create an instance of it.

```
      <instance id="fruits">
        <root>
          <item>
            <label>Apple</label>
            <name>apple</name>
          </item>
          <item>
            <label>Banana</label>
            <name>banana</name>
          </item>
        </root>
      </instance>
```
```
    <select1 ref="/data/fruit">
      <label>The Fruit</label>
      <item>
        <label>Apple</label>
        <value>apple</value>
      </item>
      <item>
        <label>Banana</label>
        <value>banana</value>
      </item>
    </select1>
```

nribeka:
I think it's coming from this line: https://github.com/XLSForm/pyxform/blame/master/pyxform/xls2json.py#L1119

```
                if row.get("choice_filter"):
                    if select_type == "select one external":
                        new_json_dict["query"] = list_name
                    else:
                        new_json_dict["itemset"] = list_name
                        json_dict["choices"] = choices
                        if choices.get(list_name):
                            new_json_dict["list_name"] = list_name
                            new_json_dict[constants.CHOICES] = choices[list_name]
```

This assignment: ```json_dict["choices"] = choices``` is adding the whole choices sheet to the json.

pyxform/tests_v1/test_xml_internal.py
```diff
@@ -0,0 +1,103 @@
...
+            |         | states             | IN         | India     |            |                   |
+            |         | fruits             | apple      | Apple     |            |                   |
+            |         | fruits             | banana     | Banana    |            |                   |
+            """
+        survey = self.md_to_pyxform_survey(md_raw=md)
+        survey_xml = survey._to_pretty_xml()
+
+        self.assertContains(survey_xml, '<instance id="yes_no">', 1)
+        self.assertContains(survey_xml, '<instance id="states">', 1)
+        self.assertContains(survey_xml, '<instance id="fruits">', 1)
```
lognaturel: `fruits` isn't used anywhere so I wouldn't expect it be generated. I think this is existing behavior so maybe we don't need to change it here?

pyxform/tests_v1/test_xml_internal.py
```diff
@@ -0,0 +1,103 @@
...
+
+        self.assertPyxformXform(
+            md=md,
+            model__contains=[
+                '<instance id="yes_no">',
```
lognaturel: The two assertion blocks seem redundant to me. Perhaps it's best to go with the first variant so the count can be verified.

PR #385, opened 2019-10-04
Adding support for dynamic defaults.

### LINKED ISSUE #338: Add support for dynamic defaults
Add support for dynamic defaults by expanding what the exiting `default` column does. See [the forum thread](https://forum.opendatakit.org/t/xlsform-spec-proposal-add-syntax-to-make-it-easy-to-use-a-value-from-the-last-saved-instance/18814/21) for discussion.

If the contents of the default column are not a literal value, the expression is passed through to the `value` attribute of a [`setvalue` action](https://opendatakit.github.io/xforms-spec/#action:setvalue) triggered by the `odk-instance-first-load` event. In the case of nodes that are not in repeats, this action is added as a child of the `model` and ideally right after the `bind` for the node of the value it sets (this is not a hard requirement -- it can be anywhere in the `model`).

In the case of repeats, the action is added as a child of the repeat form control (in the `body`) and the triggering event is `odk-instance-first-load odk-new-repeat`.

Given the following row from a survey sheet:

type | name | label | default
-- | -- | -- | --
date | event_date | Event date | today()

The following XForm content should be generated in the `model`:
```xml
<bind nodeset="/data/event_date" type="date"/>
<setvalue ref="/data/event_date" value="today()" event="odk-instance-first-load"/>
```

Given the following rows from a survey sheet:

type | name | label | default
-- | -- | -- | --
begin repeat | event | Event |
date | event_date | Event date | today()
| end repeat | event | |

The following XForm content should be generated in the `body`:
```xml
<repeat nodeset="/data/event">
  <setvalue event="odk-instance-first-load odk-new-repeat" ref="/data/event/event_date" value="today()"/>
  <input ref="/data/event/event_date">
    <label>Event date</label>
  </input>
</repeat>
```

I believe the best approach for identifying static vs. dynamic defaults will be to use a regular expression. It should be fine for static values to end up as the value for the `value` attribute so not a big deal if the regex catches some things that are not actually intended to be evaluated. However, there is the possibility that a user will want to literally use text that could be evaluated. For example, a user could want the literal string "today()" as a default and that would no longer be possible. We may want to add some kind of escape strategy for that case but I propose we wait until a user requests it because it seems unlikely to me.

nribeka:
Fix #338.

This will add dynamic default handler for the form. This failed validation on both enketo and ODKValidator I think.


pyxform/survey_element.py
```diff
@@ -335,6 +341,42 @@ def needs_itext_ref(self):
...
+                    return False
+                else:
+                    expression.append(expression_element)
+        return contains_dynamic
+
+    def xml_dynamic_default(self, in_binding=False):
+        if not self.default or not self.dynamic_default():
+            return
+
+        if self.parent.__class__.__name__ == "Survey" and in_binding:
```
nribeka: If anyone have comment on how to better do this, reply to this please.

lognaturel:
The tests look good.

For someone who wants to think about this at a high level, the proposed approach is to treat a default as dynamic if it contains one or more balanced pair of brackets ([]), parens (()) or curly braces ({}).

The only odd case I can think of that this would miss is an expression on numeric literals like `365/7`. Of course, someone could just do that math and put in the result instead but it can be nice to use expressions to add meaning.

I also am not certain that it's so important for brackets/parens/braces to be balanced. I'd tend to treat anything that contains any of `[](){}*/-` as a dynamic expression.

nribeka:
@lognaturel you are thinking of all of the arithmetic ops there, right?

Adding arithmetic ops will make it think ```abc-data``` dynamic. I could add checking for text around the operators, but that will not work for ```func_a() - func_b()``` then. I hope my question make sense.

nribeka:
@lognaturel  and the second question, does code-cov update the coverage or just checking it once?

pyxform/survey_element.py
```diff
@@ -335,6 +341,51 @@ def needs_itext_ref(self):
...
         )
 
+    def dynamic_default(self):
+        if not isinstance(self.default, basestring):
+            return False
+
+        expression = []
+        contains_dynamic = False
+        arithmetic_construct = {"*", "/", "+", "-"}
+        if self.type == "date":
```
lognaturel: If all clients supported dynamic defaults, I would say that this case should be removed because it could be useful to have math in a date default. Given that not all clients support them, this does seem like the safest thing to do.

pyxform/survey_element.py
```diff
@@ -335,6 +341,51 @@ def needs_itext_ref(self):
...
+            arithmetic_construct.remove("/")
+            arithmetic_construct.remove("-")
+
+        expression_construct = {"[", "]", "{", "}", "(", ")"}
```
lognaturel: I still don't think matching brackets/braces/parens is necessary and that a regex would be sufficient. Then the documentation can say something like 

> if a default contains any of "*", "/", "+", "-", "[", "]", "{", "}", "(", ")", it is treated as a dynamic default. Date defaults with "/" and/or "-" are treated as static defaults.

With this current implementation, the docs would say something like
 
> if a default contains any of "*", "/", "+", "-" and/or balanced sets of [], {} and/or (), it is treated as a dynamic default. If a default contains any unbalanced "[", "]", "{", "}", "(", ")", it is considered a static default. Date defaults with "/" and/or "-" are treated as static defaults.

I don't feel terrifically strongly about this. It doesn't seem very likely that an expression would contain unbalanced brackets/braces/parens so the behavior would generally be the same. The only case I can think of is if a literal string containing a single one used as an argument to an expression (something like a `:(` or `:)`)

nribeka:
@lognaturel added the warning. I moved the checking for dynamic to utils instead.

pyxform/survey_element.py
nribeka: Waiting for @MartijnR comment on this.

MartijnR: Unbalanced bracket matching seems fine to me (and balanced too). 

"/" as operator should be replaced by "div" I think (in `arithmetic_construct`).  "mod" should be added.

nribeka: Should we try to do the whole xpath operators then @MartijnR ?

tiritea: we can only do so much to hide XPath from XLSForm writers... eg a flat namespace and ${foo} helps, but pretty much that's pretty much the extent of it; all 'advanced' stuff (calculations, relevants, ...) are all raw XPath. Rather than hacking something together we might be better off just tagging dynamic defaults as another 'advanced' feature in XLSForm which requires a modicum of XPath knowledge.

So my initial thought might be we could treat _all_ defaults as XPath expressions (!), much like we do with calculations. So if you actually want a default string of `{[foo*` then you have to enclose it in quotes/escape it accordingly.

lognaturel: > "/" as operator should be replaced by "div"

Thank you!!

> So my initial thought might be we could treat all defaults as XPath expressions

I agree with you in principle but given that clients will get support for dynamic defaults at different times and that there's also a long tail of old client versions in production use, I think that would be too disruptive. That is, a form that has static defaults today and works great in Collect < v1.24/Enketo/all clients would not work in those same client versions if converted again with a `pyxform` version that always generates dynamic defaults.

tiritea: Wont this only be a problem if the (static) default happens to have weird chars in it? Presumably the same validation that pyxform is using to check calculations/relevants can be applied to defaults, and if it throws an error put a '...' around it to treat it as a static string?

lognaturel: I should be more precise. Collect only got support for all the necessary events to do dynamic defaults in v1.24. I don't believe Enketo has support for the `setvalue` action yet. So if `pyxform` generates forms with those for static defaults, users have lost access to existing functionality in existing platforms.

tiritea: my (silly) error: I thought, against an XML document `<foo></foo>` that the XPath expression `bar` would just eval to 'bar'... It doesn't. Scratch my idea.

MartijnR: > Should we try to do the whole xpath operators then @MartijnR ?

I'm not sure if more are needed. I believe you already do `and` and `or`. The others will return booleans which doesn't seem likely to be used. It won't hurt to include all of course, if it doesn't make the matching more fragile.

pyxform/utils.py
```diff
@@ -234,3 +234,37 @@ def get_languages_with_bad_tags(languages):
...
+    Dynamic value for now is defined as:
+    * Contains arithmetic operator (except '/' and '-' for date type).
+    * Contains balanced set of [] and/or {} and/or ().
+    """
+    if not isinstance(element_default, basestring):
+        return False
+
+    expression = []
+    contains_dynamic = False
+    arithmetic_construct = {"*", "/", "+", "-"}
```
lognaturel: "/" should be `div`. You can then remove the "/" exception from date.

nribeka: @lognaturel updated with ```div```, ```mod``` and ```|```. I also removed the ```/``` from the date data type.

pyxform/tests_v1/test_dynamic_default.py
```diff
@@ -0,0 +1,156 @@
...
+
+        self.assertPyxformXform(
+            md=md,
+            name="dynamic",
+            id_string="id",
+            model__contains=[
+                "<first_name/>",
+                "<last_name>not_func$</last_name>",
+                "<age/>",
+                '<setvalue event="odk-instance-first-load" ref="/dynamic/first_name" value="first_name${idea}"/>',
```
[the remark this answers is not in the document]
lognaturel: No processing of `${idea}`, correct. I think the value would be the literal string `first_name${idea}` so this would be equivalent to `<first_name>first_name${idea}</first_name>` in the instance. It's an example of a possible category of false positives.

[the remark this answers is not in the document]
MartijnR: Thanks!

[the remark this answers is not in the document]
lognaturel: I hadn't looked at this test carefully enough. In fact, `${idea}` should have been replaced by an XPath path. `${}` references were not expanded at all which is fixed by https://github.com/XLSForm/pyxform/pull/417/commits/17ca29e37ea27e58b38acda0d5956f2cef00a20f.

PR #389, opened 2019-10-21
Passing filename as the default form id parameter.

### LINKED ISSUE #388: Form title and form_id should not default to data
https://github.com/XLSForm/pyxform/pull/375 uses a default root node name of `data` which is great. It also sets the form title and the form ID to `data` which it should not.

Setting the form ID to `data` is really problematic because the form ID is what both clients and servers use to uniquely identify forms. I think the best option is to default to the filename. I don't see any restriction placed on the `id` structure either at https://opendatakit.github.io/xforms-spec/#primary-instance or at https://docs.opendatakit.org/openrosa-form-list/#query-parameters so it should be ok to include spaces, unicode, and any other characters that a filename could contain. Note that prior to https://github.com/XLSForm/pyxform/pull/375, the filename had to be XML-safe but that this is no longer the case. I have confirmed that Collect and Central have no problem with a form id that contains spaces and unicode.

I think users would expect the form title to match the filename unless they explicitly override it. I can't think of any problems with doing this.

I believe this will mean passing the filename separately from the root node name.

CC @nribeka @ukanga @MartijnR

nribeka:
Fix #388.

Introduce a new parameter for the ```workbook_to_json``` by passing the xls filename as the default value for the form_id. This will be replaced with value from the ```settings``` tab if supplied.

### DIFF
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -299,7 +299,11 @@ def process_range_question_type(row):
 
 
 def workbook_to_json(
-    workbook_dict, form_name=None, default_language="default", warnings=None
+    workbook_dict,
+    form_name=None,
+    form_id=None,
+    default_language="default",
+    warnings=None,
 ):
     """
     workbook_dict -- nested dictionaries representing a spreadsheet.
@@ -388,7 +392,7 @@ def workbook_to_json(
         )
 
     # Here we create our json dict root with default settings:
-    id_string = settings.get(constants.ID_STRING, form_name)
+    id_string = settings.get(constants.ID_STRING, form_id)
     sms_keyword = settings.get(constants.SMS_KEYWORD, id_string)
     json_dict = {
         constants.TYPE: constants.SURVEY,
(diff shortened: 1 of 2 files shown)


pyxform/xls2json.py
```diff
@@ -299,7 +299,11 @@ def process_range_question_type(row):
 
 
 def workbook_to_json(
-    workbook_dict, form_name=None, default_language="default", warnings=None
+    workbook_dict,
+    form_name=None,
+    form_id=None,
```
lognaturel: How about calling this new parameter `fallback_form_name` so it's clear that it's a companion to `form_name` that's only used as a fallback?

pyxform/xls2json.py
```diff
@@ -1316,7 +1320,10 @@ def parse_file_to_json(
     if warnings is None:
         warnings = []
     workbook_dict = parse_file_to_workbook_dict(path, file_object)
-    return workbook_to_json(workbook_dict, default_name, default_language, warnings)
+    default_form_id = unicode(get_filename(path))
```
lognaturel: I think calling this local value `fallback_form_name` would also be more helpful.

pyxform/xls2json.py
nribeka: Updated to fallback_form_name. I think it makes sense to call it that way.

PR #394, opened 2019-11-08
Add pyxform version to XForm output, closes #393

### LINKED ISSUE #393: Output `no-repeat-reference-error` in some manner or the pyxform version or the xforms version
For discussion. 

Some advanced XPath enthusiasts (@ln) are running into issues caused by Enketo's workaround for a major XForm syntax issue: https://opendatakit.github.io/xforms-spec/#a-big-deviation-with-xforms.

Pyxform recently started generating relative repeat node references, so we're on the verge of  being able to [remove this nasty deviation](https://github.com/opendatakit/xforms-spec/issues/251) from the ODK XForms spec (I think). 

However, since we still need to be able to run forms that were transformed by older versions of pyxform, it will be many years before we can remove the code workarounds in Enketo and Collect. This means it is a challenge, for Enketo, to facilitate support for those very cool advanced forms.

This may be a good reason to add some versioning or featuring in the XForms output. E.g. the output could contain a `no-repeat-reference-error` class or a `pyxform-version` attribute. Data collection clients could disable the bug-causing workaround for the buggy XForms syntax based on the presence of this class or the value of the attribute.

I imagine there will be more valid use cases in the future, so perhaps a version attribute is the easiest?

MartijnR:
Please check extra carefully, as I didn't really know what I was doing.

This code assumes each tag has the format 'v'+semantic version, as we've been using so far (except for vv0.88). If you think that may possibly change, I'd be happy to add a few lines to chop off any non numeric starting characters. Or we could not omit any characters (except trimming) and leave it up the data collection clients to do.

### DIFF
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -9,6 +9,7 @@ import json
 import os
 import re
 import sys
+import subprocess
 from collections import Counter
 
 from pyxform import aliases, constants
@@ -406,6 +407,7 @@ def workbook_to_json(
         # problems for formhub.
         # constants.VERSION : datetime.datetime.now().strftime("%Y%m%d%H"),
         constants.CHILDREN: [],
+        constants.PYXFORM_VERSION: get_git_describe_tags(),
     }
     # Here the default settings are overridden by those in the settings sheet
     json_dict.update(settings)
@@ -1434,6 +1436,12 @@ def get_parameters(raw_parameters, allowed_parameters):
             )
 
     return params
+
+
+def get_git_describe_tags():
+    return subprocess.check_output(
+        ["git", "describe", "--tags", "HEAD"], universal_newlines=True
+    ).strip()[1:]
 
 
 class SpreadsheetReader(object):
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -407,7 +407,7 @@ def workbook_to_json(
         # problems for formhub.
         # constants.VERSION : datetime.datetime.now().strftime("%Y%m%d%H"),
         constants.CHILDREN: [],
-        constants.PYXFORM_VERSION: get_git_describe_tags(),
+        constants.GENERATED_BY: ("pyxform " + get_git_describe_tags()).strip(),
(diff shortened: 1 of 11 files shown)


MartijnR:
Oops. Will try to fix all those failing tests.

yanokwa:
> I didn't really know what I was doing.

Don't worry about it. None of us know what we are doing and yet here we are...

lognaturel:
YEAH! @MartijnR out here writing Python!!

I was hoping this would be really straightforward and get you excited about more pyxform contributions 😉 but hadn't anticipated the test problem. It seems it might be tricky to address because the old style tests do literal comparisons against expected XForm output and the version is dynamic. Darn. I wonder if there's a way to exclude that attribute in the diff? Or maybe it's time to switch those tests over to the new style?

MartijnR:
Haha, I was excited initially, indeed!

> to address because the old style tests do literal comparisons against expected XForm output and the version is dynamic. Darn. I wonder if there's a way to exclude that attribute in the diff? Or maybe it's time to switch those tests over to the new style?

Exactly what I have finally concluded! Converting to new style tests would seem great. I am trying to remove the attribute from `assertXFormEqual` atm, but that seems quite a bad solution

yanokwa:
I spoke with @tiritea on Slack and he said, "Nothing new from me. I'm good."

@MartijnR I've gone ahead and squashed your commits and added a new commit that replaces pyxform_version with generated_by in the PR.

MartijnR:
Thank you! That saves me from having to fix my python environment which would have probably taken many hours.

pyxform/xls2json.py
```diff
@@ -1395,6 +1397,12 @@ def get_parameters(raw_parameters, allowed_parameters):
     return params
 
 
+def get_git_describe_tags():
+    return subprocess.check_output(
```
lognaturel: This should be wrapped in a try/except for the unlikely case that git is not installed on the host machine. I think that in that case we should return something like `unknown pyxform version` or simply `pyxform`.

PR #397, opened 2019-11-16
Better Python 2/3 unicode support for strings

yanokwa:
The use of a byte string in raising `ODKValidateError` has strange downstream behavior. For example, if you use jsonify to output that string, you get `b` added to your output like so:
Screen Shot 2019-11-15 at 7 17 41 PM

Instead, I use the futures library, which is designed for Python 2/3 code to do the right thing with the string. 

While I was poking around for other Python 2/3 string problems, I saw the use of `sys.version_info.major` and that seemed unnecessary so I removed it.


yanokwa:
Thanks for the approval, @dorey. Keep riding that dragon. 🐉

dorey:
<h1>🖥️🐉</h1>

PR #400, opened 2019-11-28
Add track-changes-reason audit attribute

seadowg:
This adds the `track-changes-reasons` as defined in [this spec](https://docs.google.com/document/d/1NyBFhASCOAqOSmz70diQObKpymVE59_MuwjOUPCKlR8/edit#).

PR #406, opened 2019-12-12
Identify first argument of pulldata by splitting on comma

### LINKED ISSUE #398: Double quotes in pulldata parameter expression results in wrong instance ID and CSV filename being generated
I designed a form to pull data from a CSV as described in the [XLSForm docs](https://xlsform.org/en/#how-to-pull-data-from-csv); see https://github.com/opendatakit/central/issues/115 for details of what I tried. And the conversion results in a reference to a "RC.csv" file despite the XLSForm xlsx referencing a filename of "registered": `pulldata('registered','first','rcid',concat("RC",${rcid}))` 

What I did differently from the example which may be causing the issue was the use of the `concat()` within the `pulldata()` when providing the lookup value, and the use of double quotes to provide a string as part of that?

lognaturel:
Closes #398

In the issue, I mentioned using `csv` to handle getting the arguments from the `pulldata` call but I realized it's not all that helpful because it only deals with double quotes (and not single quotes) and because the first argument can't be dynamic anyway. That is, the `pulldata` implementation as it exists in Collect does allow for the first argument to be dynamic so you could have it be the result of e.g. a `concat` function call to dynamically use different files with the same prefix depending on some selection. But that won't work when authoring XLSForms now that a secondary instance declaration is added because the value of that first argument is used as the instance ID and needs to be a valid XML identifier. All that to say that we can just split on commas and strip quotes around the first argument which is what I've done here.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -315,15 +315,21 @@ class Survey(Section):
                 instance=node("instance", id=file_id, src=uri),
             )
 
-        formulas = get_pulldata_functions(element)
-        if len(formulas) > 0:
-            formula_instances = []
-            for formula in formulas:
-                pieces = formula.split('"') if '"' in formula else formula.split("'")
-                if len(pieces) > 1 and pieces[1]:
-                    file_id = pieces[1]
-                    formula_instances.append(get_instance_info(element, file_id))
-            return formula_instances
+        pulldata_calls = get_pulldata_functions(element)
+        if len(pulldata_calls) > 0:
+            pulldata_instances = []
+            for pulldata_call in pulldata_calls:
+                pulldata_arguments = re.sub("pulldata\s*\(\s*", "", pulldata_call)
+                parsed_pulldata_arguments = pulldata_arguments.split(",")
+                if len(parsed_pulldata_arguments) > 0:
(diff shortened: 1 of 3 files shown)


lognaturel:
Apologies to anyone who may have seen the force pushes coming in. I'm done now.

lognaturel:
Ok, @seadowg, what did you do to get black happy? I'm running `black --target-version=py27 pyxform` and all looks good locally.

seadowg:
@lognaturel I needed to install the exact version of black that CI uses (I did so with `pip install -r requirements.pip`). You'll see in #400 that I updated the docs with that as an instruction on environment setup. It seems really suspicious to me that a formatting tool can have that level of variation in behaviour. If it was up to me I'd get rid of it Immediately.

yanokwa:
@lognaturel In https://github.com/XLSForm/pyxform/pull/408 I've changed the black command so it prints what files need to be reformatted. 

@seadog I'm not sure the problem you had, but I think black has been a net positive. I've made some changes so we go with its defaults and I'm hopeful it's more helpful.

seadowg:
@yanokwa I ran into problems with different Black versions providing different results. It's potentially just a problem with the version mismatch I had but our introduction has left me with a bad attitude™️ towards it.

Issue #407, opened 2019-12-19
Dev install fails on Python 3.8 because of typed-ast


yanokwa:
According to https://github.com/python/mypy/issues/7001#issuecomment-502468009, typed-ast does not work on Python 3.8. We should upgrade to from typed-ast 1.3.5 to a version that does.

Output from `pip install -r requirements.pip`
```
Installing collected packages: appdirs, typed-ast, lazy-object-proxy, wrapt, six, astroid, attrs, click, toml, black, entrypoints, pycodestyle, mccabe, pyflakes, flake8, formencode, isort, linecache2, mock, nose, pip-tools, pylint, traceback2, unicodecsv, unittest2, xlrd, yapf
    Running setup.py install for typed-ast ... error
    ERROR: Command errored out with exit status 1:
     command: /usr/local/pyenv/versions/3.8.0/envs/pyxform-env/bin/python -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/3r/3bt8vq7s1d98y0_lpd8w8twr0000gn/T/pip-install-9wxnjtkx/typed-ast/setup.py'"'"'; __file__='"'"'/private/var/folders/3r/3bt8vq7s1d98y0_lpd8w8twr0000gn/T/pip-install-9wxnjtkx/typed-ast/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /private/var/folders/3r/3bt8vq7s1d98y0_lpd8w8twr0000gn/T/pip-record-57r8r1oy/install-record.txt --single-version-externally-managed --compile --install-headers /usr/local/pyenv/versions/3.8.0/envs/pyxform-env/include/site/python3.8/typed-ast
         cwd: /private/var/folders/3r/3bt8vq7s1d98y0_lpd8w8twr0000gn/T/pip-install-9wxnjtkx/typed-ast/
    Complete output (58 lines):
    running install
    running build
    running build_py
    creating build
    creating build/lib.macosx-10.15-x86_64-3.8
    creating build/lib.macosx-10.15-x86_64-3.8/typed_ast
    copying typed_ast/conversions.py -> build/lib.macosx-10.15-x86_64-3.8/typed_ast
[...]

PR #408, opened 2019-12-20
Remove unsupported or old Python CI builds, add new ones

yanokwa:
Following the table at https://en.wikipedia.org/wiki/History_of_Python#Table_of_versions, I made these changes: 
* Python 2.7 is dropped because it will receive no security fixes starting Jan 2020 so we should not test on or target formatting to it.
* Python 3.5 is dropped because you can't run black in it. Also it reached end of full support in August 2017 and will receive no security fixes starting Sept 2020. Also, it makes builds slower to have so many Pythons.
* Python 3.6 is dropped because it will reach end of full support at the end of this year. Also, it makes builds slower to have so many Pythons.
* Python 3.8 has been released so we should test on it

Then I 
* Simplified formatting check now that we've dropped Python 2.7 and 3.5. Now if it returns an error code, we list the files so folks can see what files CI thinks they should run black on and exit.

Also
* The version of typed-ast we had didn't support Python 3.8, so went ahead and ran `pip-compile --output-file=requirements.pip requirements.in` to get the most correct requirements.

Fixes https://github.com/XLSForm/pyxform/issues/407

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,15 +24,7 @@ jobs:
       - run:
           name: Test code is well formatted
           command: |
-            if ! type /home/circleci/.local/bin/black > /dev/null; then
-                echo "Black is not supported in this python version :("
-            else
-                output="$(/home/circleci/.local/bin/black pyxform --check  2>&1 >/dev/null)"
-                if [[ $output == *"would be reformatted"* ]]; then
-                    echo 'The source code could use a bit more black.'
-                    exit 1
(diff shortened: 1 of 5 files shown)


lognaturel:
Ack, I didn't notice that this never ran on CircleCI. It looks like tests fail on Python 3.8 because some ordering has changed. See https://github.com/XLSForm/pyxform/pull/394#pullrequestreview-335538975

PR #410, opened 2019-12-20
Output a string bind type for selects

### LINKED ISSUE #168: bind for select1 nodeset is "select1" and this throws an XForms syntax error
bind for select1 nodeset is "select1" and this throws an XForms syntax error. 

I think it should be the XML Schema data type of the value/s

This is a simple example output illustrating how XLSForms binds giving the type as "select1"

```
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <h:head>
    <h:title>foo</h:title>
    <model>
      <submission action="http://xformstest.org/cgi-bin/echoinstance.sh" method="form-data-post"/>
      <instance>
        <instance_xmlns_test id="instance_xmlns_test" xmlns="1234">
          <yes_or_no>yes</yes_or_no>
          <age/>
          <meta>
            <instanceID/>
          </meta>
        </instance_xmlns_test>
      </instance>
      <bind nodeset="/instance_xmlns_test/yes_or_no" type="select1"/>
      <bind constraint=". &gt;=18" jr:constraintMsg="Respondent must be 18 or over" nodeset="/instance_xmlns_test/age" type="int"/>
      <bind calculate="concat('uuid:', uuid())" nodeset="/instance_xmlns_test/meta/instanceID" readonly="true()" type="string"/>
    </model>
  </h:head>
  <h:body>
    <select1 ref="/instance_xmlns_test/yes_or_no">
      <label>Yes or No?</label>
      <hint>Look on the signboard</hint>
      <item>
        <label>Yes</label>
        <value>yes</value>
      </item>
      <item>
        <label>No</label>
        <value>no</value>
      </item>
    </select1>
    <input ref="/instance_xmlns_test/age">
      <label>How old are you?</label>
      <hint>Check your birth certificate</hint>
    </input>
  </h:body>
</h:html>

```

yanokwa:
Closes #168

### DIFF
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -134,7 +134,7 @@ class MultipleChoiceQuestion(Question):
             choice.validate()
 
     def xml_control(self):
-        assert self.bind["type"] in ["select", "string", "odk:rank"]
+        assert self.bind["type"] in ["string", "odk:rank"]
         survey = self.get_root()
         control_dict = self.control.copy()
         # Resolve field references in attributes
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -134,7 +134,7 @@ class MultipleChoiceQuestion(Question):
             choice.validate()
 
     def xml_control(self):
-        assert self.bind["type"] in ["select", "select1", "odk:rank"]
+        assert self.bind["type"] in ["select", "string", "odk:rank"]
         survey = self.get_root()
         control_dict = self.control.copy()
         # Resolve field references in attributes
--- a/pyxform/question_type_dictionary.py
+++ b/pyxform/question_type_dictionary.py
@@ -47,7 +47,7 @@ QUESTION_TYPE_DICT = {
     },
     "add select multiple prompt using": {
         "control": {"tag": "select"},
-        "bind": {"type": "select"},
+        "bind": {"type": "string"},
     },
     "add note prompt": {
         "control": {"tag": "input"},
@@ -59,7 +59,7 @@ QUESTION_TYPE_DICT = {
     "text": {"control": {"tag": "input"}, "bind": {"type": "string"}},
     "select all that apply from": {
         "control": {"tag": "select"},
-        "bind": {"type": "select"},
+        "bind": {"type": "string"},
     },
     "simserial": {
         "bind": {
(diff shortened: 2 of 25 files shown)


pyxform/question.py
```diff
@@ -134,7 +134,7 @@ def validate(self):
             choice.validate()
 
     def xml_control(self):
```
yanokwa: Fairly confident this will introduce a bug somewhere...

lognaturel: I don't think that this makes sense. It seems that the control type should be checked for selects, now, instead of the bind type.

lognaturel: I stand corrected. This is an assert and it's in the `MultipleChoiceQuestion` implementation. So it is ok.

lognaturel:
I am running the failing test with `nosetests pyxform.tests.xform2json_test.DumpAndLoadXForm2JsonTests`. It looks like `SurveyElement._link_children` is hitting a case where something that previously had `children` is now a `dict` that doesn't.

For example, in the `specify_other.xls` case, I see the following json generated for the male choice when the tests pass:
`{'name': 'male', 'compact_tag': '', 'sms_field': '', 'sms_option': '', 'label': {'English': 'Male'}, 'hint': '', 'guidance_hint': '', 'default': '', 'type': '', 'appearance': '', 'parameters': '', 'intent': '', 'jr:count': '', 'bind': {}, 'instance': {}, 'control': {}, 'media': {}, 'parent': None, 'children': [], 'itemset': '', 'choice_filter': '', 'query': '', 'autoplay': '', 'flat': False, 'action': '', 'list_name': ''}`

This suggests to me that select choices are being represented as `SurveyElement` objects which do have `children` attributes.

When the tests fail, select choices are represented as dictionaries:
`{'name': 'male', 'label': {'English': 'Male'}}`

I don't know why a select choice would be represented as a `SurveyElement` or why that json output would be desirable. I don't believe that any of the `SurveyElement` attributes could be populated. The fact that all the other tests pass suggests that the XForms output is not affected.

lognaturel:
The XForms to json functionality was using the bind type rather than the body type to identify selects. I have added a commit to override this. I have run `nosetests` locally and confirmed they pass. I have also run `black pyxform`.

I fully expect that some other downstream integrations use the bind type this way. We just need to be ready to react when folks run into issues.

PR #413, opened 2020-01-06
Ensure stable ordering of attributes for tests

yanokwa:
In 5c7c5a0f14cab112cd6aa267d94fd67897503678, we use the same code from https://github.com/XLSForm/pyxform/pull/412 to produce what I call "testable" XML (XML with a stable ordering). I couldn't find an easy way to de-duplicate the code without invasive changes. A more experienced Pythonista might know better.

9bec7ffe00c7984d67cca8e73bf104cdb034f72e, I narrowed the test to not depend on attribute order.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -716,6 +716,21 @@ class Survey(Section):
     def _to_ugly_xml(self):
         return '<?xml version="1.0"?>' + self.xml().toxml()
 
+    def _to_testable_xml(self):
+        """Preserves attribute ordering across all Python versions.
+
+        See python_test_case.reorder_attributes for related code.
+        """
+        tree = ETree.fromstring(self.xml().toxml())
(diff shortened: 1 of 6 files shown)


pyxform/survey.py
```diff
@@ -713,8 +713,20 @@ def date_stamp(self):
         """Returns a date string with the format of %Y_%m_%d."""
         return self._created.strftime("%Y_%m_%d")
 
-    def _to_ugly_xml(self):
-        return '<?xml version="1.0"?>' + self.xml().toxml()
+    def _to_testable_xml(self):
```
yanokwa: Note that we don't include the `<?xml version="1.0"?>` stuff. Maybe we should?

lognaturel: I don't think it matters at this level.

PR #417, opened 2020-01-28
Add support for last-saved

### LINKED ISSUE #337: Introduce ${last-saved#<question_name>} to refer to a question in the last saved instance
See https://forum.opendatakit.org/t/xlsform-spec-proposal-add-syntax-to-make-it-easy-to-use-a-value-from-the-last-saved-instance/18814/22

If an XLSForm definition uses the `${last-saved#<question_name>}` construct anywhere, `<instance id="__last-saved" src="jr://instance/last-saved" />` will be added to the XForm definition.

`${last-saved#<question_name>}` will expand to `instance('__last-saved')/root_node/question_name` (with additional group/repeat levels as needed -- same as with the `${<question_name>}` construct).

The `__` prefix convention should be documented as a way to avoid collisions when pyxform needs to create ids for secondary instances.

lognaturel:
Closes #337

The primary purpose of this PR is to add support for last-saved as described in #337. I also fixed some things in `pulldata` and dynamic defaults support that I noticed while working on parallel functionality with last-saved. I hope that this will be straightforward to review commit by commit but let me know if I should split the PR up.

To verify that this works as expected, I ran the tests and made sure the existing test suite passes. I also ran Validate against the new tests. Some fail due to "self-references" because of known issue https://github.com/opendatakit/javarosa/issues/449.

@MartijnR and @tiritea, I would really appreciate your eagle-eyed review of [the test cases](https://github.com/XLSForm/pyxform/pull/417/files#diff-c8c7db56f67a0cce55f2e9406753b57c) if you have a moment.


.python-version
```diff
@@ -0,0 +1 @@
+3.8.1
```
yanokwa: This file should be removed.

pyxform/tests_v1/test_last_saved.py
```diff
@@ -0,0 +1,162 @@
...
+            xml__contains=[
+                '<instance id="__last-saved" src="jr://instance/last-saved"/>',
+                '<setvalue event="odk-instance-first-load" ref="/last-saved/foo" value=" instance(\'__last-saved\')/last-saved/foo "/>',
+            ],
+        )
+
+    # This is not really a useful thing to do because we get an unqualified reference. Clients should use the first
+    # node in the repeat in an expression like this one where a single value is needed or the full nodeset in an
+    # expression where a nodeset is required. An alternative to consider is to generate an expression with position(..).
+    def test_last_saved_in_repeat(self):
```
yanokwa: This test feels too complicated to me. I'd rather see a repeat test with `foo` and `bar` where `bar` tries to get last-saved of `foo`. Then another where `bar` gets last-saved outside the repeat. And then maybe a two level repeat.

pyxform/tests_v1/test_last_saved.py
```diff
@@ -0,0 +1,162 @@
...
+            |        | type       | name     | label | required               |
+            |        | integer    | foo      | Foo   | ${last-saved#foo} < 12 |
+            """,
+            xml__contains=[
+                '<instance id="__last-saved" src="jr://instance/last-saved"/>',
+                "required=\" instance('__last-saved')/last-saved/foo  &lt; 12\"",
+            ],
+        )
+
+    def test_last_saved_in_choice_filter_does_not_use_current_or_relative_ref(self):
```
yanokwa: I have no idea what this test does. Can it be renamed to clarify why we don't need to use current/relative?

pyxform/tests_v1/test_last_saved.py
```diff
@@ -0,0 +1,194 @@
+# -*- coding: utf-8 -*-
+"""
+The last-saved virtual instance can be queried to get values from the last saved instance of the form being authored.
+"""
+from pyxform.tests_v1.pyxform_test_case import PyxformTestCase
+
+
+class LastSavedTest(PyxformTestCase):
```
yanokwa: I'd like to see `${last-saved}` in a label that refers to itself and the same for a calculate.

yanokwa:
I used `if(${last-saved#patient_count}   == '', 0, ${last-saved#patient_count} + 1)` in a default and got this error in conversion.

```
  File "/usr/local/pyenv/versions/pyxform-3.8.1/bin/xls2xform", line 11, in <module>
    load_entry_point('pyxform', 'console_scripts', 'xls2xform')()
  File "/Users/yanokwa/Documents/Work/Nafundi/Code/yanokwa/pyxform/pyxform/xls2xform.py", line 172, in main_cli
    warnings = xls2xform_convert(
  File "/Users/yanokwa/Documents/Work/Nafundi/Code/yanokwa/pyxform/pyxform/xls2xform.py", line 38, in xls2xform_convert
    json_survey = xls2json.parse_file_to_json(xlsform_path, warnings=warnings)
  File "/Users/yanokwa/Documents/Work/Nafundi/Code/yanokwa/pyxform/pyxform/xls2json.py", line 1388, in parse_file_to_json
    return workbook_to_json(
  File "/Users/yanokwa/Documents/Work/Nafundi/Code/yanokwa/pyxform/pyxform/xls2json.py", line 632, in workbook_to_json
    default_is_dynamic(question_default, question_type)
  File "/Users/yanokwa/Documents/Work/Nafundi/Code/yanokwa/pyxform/pyxform/utils.py", line 268, in default_is_dynamic
    if expression and expression.pop() != expression_pair[expression_element]:
KeyError: '{'
```

PR #418, opened 2020-01-31
Clarify duplicate choice error message

### LINKED ISSUE #373: Duplicate choice name message is confusing
#310 introduced a `allow_choice_duplicates` setting. The error message when that setting is not set and a list has duplicate choice names is: "There does not seem to be a `allow_choice_duplicates` column header defined in your settings sheet. You must have set `allow_choice_duplicates` setting in your settings sheet to have duplicate choice list names in your choices sheet"

The message should lead with what is wrong with the form (a choice list has duplicate choice names).

I propose something like:
"One or more of your choice lists contains duplicate choice names. That means the choices with the same name will be impossible to differentiate in analysis. If this is intentional, you can set the `allow_choice_duplicates` setting. Read more at `<url once documented>`."

Having a URL means there can be an example which would be helpful to users who may not understand the problem with duplicate choice names.

It might also be helpful to identify the lists and/or the choices that are affected. https://github.com/XLSForm/pyxform/pull/310/files#diff-b29aeb282b118614b8333cdfc19f700dR494 attempts to do this but it should be revisited because it only happens if the `allow_choice_duplicates` is defined in the settings sheet but not set to yes|Yes.

lognaturel:
Closes #373

Introduces a clearer message for duplicate choices in the same list, linking out to documentation provided at https://github.com/XLSForm/xlsform.github.io/pull/184. 

For some reason, the case where no `allow_choice_duplicates` setting was defined was previously less informative than the case where the setting existed but was not set to `yes`. These now show the same message.

### DIFF
--- a/pyxform/tests_v1/test_fields.py
+++ b/pyxform/tests_v1/test_fields.py
@@ -42,7 +42,7 @@ class FieldsTests(PyxformTestCase):
             error__contains=["There are more than one survey elements named 'age'"],
         )
 
-    def test_duplicate_choices_without_settings(self):
+    def test_duplicate_choices_without_setting(self):
         self.assertPyxformXform(
             md="""
             | survey  |                 |          |          |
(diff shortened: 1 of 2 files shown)


WinnyTroy:
Looks good to me!

lognaturel:
Thanks, @WinnyTroy!

PR #420, opened 2020-02-06
Remove support for generated-by reverting #394

yanokwa:
There is an ongoing discussion among other maintainers (@MartijnR, @lognaturel, @tiritea) about the placement of `generated-by` in the DOM as well as how that feature will work with the user-agent string we are thinking about adding. 

Although `generated-by` is in the spec, I thought it wise to pull the implementation in pyxform until we are absolutely certain we want to proceed. If we decide we want to proceed, it will be easy to release this in a point release.

Reverts #394

### DIFF
--- a/pyxform/constants.py
+++ b/pyxform/constants.py
@@ -28,7 +28,6 @@ COMPACT_DELIMITER = "delimiter"
 COMPACT_TAG = "compact_tag"
 
 VERSION = "version"
-GENERATED_BY = "generated_by"
 PUBLIC_KEY = "public_key"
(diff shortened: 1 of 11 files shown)


lognaturel:
@yanokwa and I have been planning to do a release today and we think we should proceed because we have time now to go through the release process carefully. We don't want to release `generated-by` because there is an open conversation and it's a tough thing to walk back.

@MartijnR I promise you that as soon as we get a chance for another quick round of discussion we'll get a new version out with whatever we end up with.

PR #430, opened 2020-02-13
Add xforms version

### LINKED ISSUE #393: Output `no-repeat-reference-error` in some manner or the pyxform version or the xforms version

MartijnR:
closes #393


MartijnR:
@lognaturel, there is [one pesky test](https://github.com/XLSForm/pyxform/blob/master/pyxform/tests/xform2json_test.py#L45) that outputs `odk:xforms-version=""` instead of `odk:xforms-version="1.0.0"`. Do you perhaps have an idea to could put me in the right direction to fix this?

lognaturel:
Taking a look now! Those XForm2json tests are really frustrating. Do you know of anyone using that code to attempt to parse XForms to get JSON?

lognaturel:
At a high-level, I imagine that `xform2json.py` needs to explicitly save the value of that attribute when it parses an XForm. You probably figured that. Digging deeper...

lognaturel:
https://github.com/lognaturel/pyxform/commit/f392273ca9cdd7c43fb8bc854c288c6f56d2c188 is a fix.

MartijnR:
Thank you so much!

Sorry, I messed up the branch. Indeed, it shouldn't be an XLSForm setting!

MartijnR:
Thank you! Turned out to be just a simple change, but it got me to explore the exciting  world of pyxform. Ready for my next assignment! (No, I'm joking, I need a lengthy rest first!)

lognaturel:
> I need a lengthy rest first!

`pyxform` makes me feel like that a lot, too. It's definitely a heavy thing maintaining old code bases when the original masterminds aren't around. But everything we do here adds a lot of value to users so that keeps me going.

PR #441, opened 2020-04-28
Improve repeat count XForm output

### LINKED ISSUE #435: Improve repeat count XForm output
I think there ~~are 2~~ is an issue with how pyxform transforms repeat_count values:

1.~~It places a new count node as sibling of the repeat. This means that any relevant condition on an ancestor of that repeat clears the repeat_count values and removes all repeats. This is not always desirable for implementations that want to keep irrelevant values stored in the model.~~

2. The output is more verbose than necessary for simple ${node} references or static numbers.

See example below with the 3 different types of repeat_count values we support. I propose to change the output ~~of all 3~~ the first and the third.

type | name | label | repeat_count
-- | -- | -- | --
integer | count | Enter count |  
begin repeat | rep1 | Repeat with user entered count | ${count}
text | text1 | Enter text |  
end repeat |   |   |  
begin repeat | rep2 | Repeat with calculated count | ${count} + 2
text | text2 | Enter text |  
end repeat |   |   |  
begin repeat | rep3 | Repeat with fixed count | 4
text | text3 | Enter text |  
end repeat |   |   |  

[Google Sheets](https://docs.google.com/spreadsheets/d/1zVTaUNcp_62z1TlH0Xi3UYnis5JPQ4ExjUYgDTgHCA0/edit?usp=sharing)

current output (relevant parts):

```xml
        <model odk:xforms-version="1.0.0">
            <instance>
                <data id="repeat-count-cases">
                    <count/>
                    <group>
                        <rep1_count/>
                        <rep1 jr:template="">
                            ...
                        </rep1>
                        <rep1>
                            ...
                        </rep1>
                        <rep2_count/>
                        <rep2 jr:template="">
                            ...
                        </rep2>
                        <rep2>
                            ...
                        </rep2>
                        <rep3_count/>
                        <rep3 jr:template="">
                            ...
                        </rep3>
                        <rep3>
                            ...
                        </rep3>
                    </group>
                    ...
                </data>
            </instance>
            <bind nodeset="/data/count" type="int"/>
            <bind calculate=" /data/count " nodeset="/data/group/rep1_count" readonly="true()" type="string"/>
            <bind calculate=" /data/count  + 2" nodeset="/data/group/rep2_count" readonly="true()" type="string"/>
            <bind calculate="4" nodeset="/data/group/rep3_count" readonly="true()" type="string"/>
            ...
        </model>
    </h:head>
    <h:body>
        <input ref="/data/count">
            <label>Enter count</label>
        </input>
        <group ref="/data/group">
            <group ref="/data/group/rep1">
                <label>Repeat with user entered count</label>
                <repeat jr:count=" /data/group/rep1_count " nodeset="/data/group/rep1">
                    ...
                </repeat>
            </group>
            <group ref="/data/group/rep2">
                <label>Repeat with calculated count</label>
                <repeat jr:count=" /data/group/rep2_count " nodeset="/data/group/rep2">
                    ...
                </repeat>
            </group>
            <group ref="/data/group/rep3">
                <label>Repeat with fixed count</label>
                <repeat jr:count=" /data/group/rep3_count " nodeset="/data/group/rep3">
                    ...
                </repeat>
            </group>
        </group>
    </h:body>
```

A cleaner, and less problematic, output would be:

```xml
<model odk:xforms-version="1.0.0">
            <instance>
                <data id="repeat-count-cases">
                    <count/>
                    <group>
                        <rep1 jr:template=""><!-- NOTE: see body and bind, no additional node or calculation has to be created by pyxform for this case-->
                            ...
                        </rep1>
                        <rep1>
                            ....
                        </rep1>
                        <rep2_count/>
                        <rep2 jr:template=""><!-- NOTE: unchanged -->
                            ....
                        </rep2>
                        <rep2>
                           ...
                        </rep2>
                    
                        <rep3 jr:template=""><!-- NOTE: see body and bind, no additional node or calculation has to be created by pyxform for this case-->
                           ....
                        </rep3>
                        <rep3>
                           ...
                        </rep3>
                           ...
                      </group>
                </data>
            </instance>
            ....

            <bind calculate=" /data/count  + 2" nodeset="/data/group/rep2_count" readonly="true()" type="string"/>
            
             ....
        </model>
    </h:head>
    <h:body>
        <input ref="/data/count">
            <label>Enter count</label>
        </input>
        ....
            <repeat jr:count=" /data/count " nodeset="/data/rep1">
                 .....
            </repeat>
       .....
            <repeat jr:count=" /data/group/rep2_count " nodeset="/data/rep2">
                 .....
            </repeat>
        ....
            <repeat jr:count="4" nodeset="/data/rep3">
                .....
            </repeat>
    </h:body>
```

gushil:
Closes #435 

Note: Got error on ODK Validation. Already reported by @MartijnR with https://github.com/getodk/javarosa/issues/399#issuecomment-620085526

### DIFF
--- a/pyxform/tests/test_expected_output/repeat_date_test.xml
+++ b/pyxform/tests/test_expected_output/repeat_date_test.xml
@@ -7,7 +7,6 @@
         <repeat_date_test id="repeat_date_test">
           <generated_note_name_2/>
           <repeat_count/>
-          <repeat_test_count/>
           <repeat_test jr:template="">
             <table_list_3/>
             <table_list_4/>
@@ -24,7 +23,6 @@
       </instance>
       <bind nodeset="/repeat_date_test/generated_note_name_2" readonly="true()" type="string"/>
       <bind calculate="1" nodeset="/repeat_date_test/repeat_count" type="string"/>
-      <bind calculate=" /repeat_date_test/repeat_count " nodeset="/repeat_date_test/repeat_test_count" readonly="true()" type="string"/>
       <bind nodeset="/repeat_date_test/repeat_test/table_list_3" type="string"/>
       <bind nodeset="/repeat_date_test/repeat_test/table_list_4" type="string"/>
       <bind nodeset="/repeat_date_test/generated_note_name_8" readonly="true()" type="string"/>
@@ -37,7 +35,7 @@
     </input>
     <group ref="/repeat_date_test/repeat_test">
       <label></label>
-      <repeat jr:count=" /repeat_date_test/repeat_test_count " nodeset="/repeat_date_test/repeat_test">
+      <repeat jr:count=" /repeat_date_test/repeat_count " nodeset="/repeat_date_test/repeat_test">
         <select1 ref="/repeat_date_test/repeat_test/table_list_3">
           <label>Q1</label>
           <item>
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -591,8 +591,6 @@ def workbook_to_json(
     # Rows from the survey sheet that should be nested in meta
     survey_meta = []
 
-    survey_children_array = stack[0]["parent_children"]
-
     repeat_behavior_warning_added = False
     dynamic_default_warning_added = False
     for row in survey_sheet:
@@ -953,7 +951,7 @@ def workbook_to_json(
                         repeat_count_expression
                     ):
                         generated_node_name = new_json_dict["name"] + "_count"
-                        survey_children_array.append(
+                        parent_children_array.append(
                             {
                                 "name": generated_node_name,
                                 "bind": {
@@ -1331,6 +1329,7 @@ def workbook_to_json(
             "control": {"bodyless": True},
             "children": meta_children,
         }
+        survey_children_array = stack[0]["parent_children"]
         survey_children_array.append(meta_element)
 
     # print_pyobj_to_json(json_dict)
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -586,8 +586,12 @@ def workbook_to_json(
         r"(?P<osm_command>(" + "|".join(aliases.osm.keys()) + r")) (?P<list_name>\S+)"
     )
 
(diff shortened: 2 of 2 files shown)


MartijnR:
Looks great to me, but needs:

* markdown tests for all three cases in #435
* hold off on merge until about a year after getodk/javarosa#399 is fixed

lognaturel:
Thanks, all! Aside from the javarosa issue, this does change the form definition schema. Is it worth it? While the new XML is cleaner, I don't see any major advantage for standard ODK XForms clients. Is there still an advantage in the special case where irrelevant repeats may be kept, @MartijnR?

MartijnR:
@lognaturel, it's indeed just the syntax improvement (no advantage for irrelevant usage in OpenClinica any more).

I think it's worth it, but that's because I'm assuming that removing nodes from the model does not cause a problem, or not a serious one. Though, please do let me know if I'm missing something. My thinking is that:

a.  in the case of a user-entered number, there is still another node that contains the number that represents the repeat count - which seems the logical one to use for analysis
b. in the case of a static number, I don't think there would be an analysis issue, because there is nothing to analyze wrt a constant

lognaturel:
> I'm assuming that removing nodes from the model does not cause a problem, or not a serious one

I know it's hard to believe but it blocks form updates for Aggregate which continues to be broadly deployed. 😭 That means that if someone wanted to update label text in a form with a repeat count, they'd be blocked from doing so.

I'd really prefer not to make this change. I think the user benefit is low and the potential pain is high. On top of that, it increases complexity in `pyxform` and the chance of invalid forms being produced. For example, the current solution's regex for identifying what needs to go in a calculate rather than inline would not pick up on a function call which is a perfectly valid expression for a repeat count. In this case there's an easy fix -- identify integer values and consider those static -- but it still means having two cases.

Ideally someone would have noticed when implementing pyxform that there was a JavaRosa bug and filed it them. But maybe they didn't even try to have separate cases for static and dynamic values because, as I said before, it does add complexity.

lognaturel:
Thanks all. Closing after conversation at https://github.com/getodk/javarosa/issues/399#issuecomment-637578276.

PR #447, opened 2020-06-04
Trigger column

### LINKED ISSUE #70: Add non-string calculation types
**EDIT: see https://github.com/XLSForm/pyxform/issues/70#issuecomment-599660252 for the latest proposed approach.**

Initial bug reported at https://groups.google.com/d/msg/opendatakit/eYFsVCLet1U/Wb4wZumGBAAJ

"The problem seems to be that the total_feld_calc field is stored as a "string" field, so the subsequent sum(${total_field_calc}) will fail with a type conversion failure."

### LINKED ISSUE #438: Add "trigger" column to use value changes as triggers for calculations

gushil:
Closes #438 
Closes #70


pyxform/question.py
```diff
@@ -40,7 +66,10 @@ class InputQuestion(Question):
     dates, geopoints, barcodes ...
     """
 
-    def xml_control(self):
+    def build_xml(self):
+        if ("calculate" in self.bind or self.trigger) and not (self.label or self.hint):
```
lognaturel, line 70: This is part of the approach for https://github.com/XLSForm/pyxform/issues/70. It ensures no body XML is generated if there's no label or hint and there is a calculate or trigger. Should this actually be in `Question`? @MartijnR it would be fine now to have any type for a calculate, not just input types, right?

lognaturel:
Here's something else I wanted to mention -- there's no support added to xform2json and little thought has been given to the JSON representation. I think in general it's up to folks who rely on those tools to participate in review or submit the functionality after the fact if they need it.

MartijnR:
Thank you! Will check it out and am aiming for tomorrow!

pyxform/tests_v1/test_trigger.py
```diff
@@ -0,0 +1,218 @@
...
+            |        | type               | name | label       | calculation        | trigger |
+            |        | select_one choices | a    | Some choice |                    |         |
+            |        | integer            | b    |             | string-length(${a})| ${a}    |
+            | choices|                    |      |             |                    |         |
+            |        | list_name          | name | label       |
+            |        | choices            | a    | A           |
+            |        | choices            | aa   | AA          |
+            """,
+            xml__contains=[
+                '<setvalue event="xforms-value-changed" ref=" /trigger-select_trigger/b " value="string-length( /trigger-select_trigger/a )"/>\n    </select1>'
```
MartijnR, line 216: Ah, so that's how you test the hierarchy. When I looking at these comprehensive tests, I realized it would be really great if we could have a test function that parses the XML and we can do something like `xml_tree_contains = [ [select1, setvalue] ]` (in document order), or better: `xml_query_finds = [ 'select1[ref="/trigger-select_trigger/a"] > setvalue[ref="/trigger-select_trigger/b"] ]` (if there is a python library that supports `querySelector`)

pyxform/tests_v1/test_typed_calculates.py
```diff
@@ -0,0 +1,59 @@
...
+        self.assertPyxformXform(
+            name="no-label",
+            md="""
+        | survey |          |      |             |
+        |        | type     | name | label       |
+        |        | integer  | a    |             |
+        """,
+            errored=True,
+            error__contains="The survey element named 'a' has no label or hint.",
+        )
```
MartijnR, line 59:
```python
self.assertPyxformXform(
            name="calculate-datetime",
            md="""
            | survey |          |      |             |             |
            |        | type     | name | label       | calculation |
            |        | dateTime | a    |             | now()       |
            """,
            xml__contains=[
                '<bind calculate="now()" nodeset="/calculate-datetime/a" type="dateTime"/>'
            instance__contains=["<a/>"],
            xml__excludes=["input"],
        )

self.assertPyxformXform(
            name="calculate-geopoint",
            md="""
            | survey |          |      |             |             |
            |        | type     | name | label       | calculation |
            |        | geopoint | a    |             | 0 0 0 0     |
            """,
            xml__contains=[
                '<bind calculate="0 0 0 0" nodeset="/calculate-geopoint/a" type="geopoint"/>'
            instance__contains=["<a/>"],
            xml__excludes=["input"],
        )
```

MartijnR: I'm comfortable with this coverage, and letting any future bug fixes add more.

MartijnR:
>  there's no support added to xform2json and little thought has been given to the JSON representation. I think in general it's up to folks who rely on those tools to participate in review or submit the functionality after the fact if they need it

This makes sense to me. Didn't realize JSON wasn't used internally at all.

Thanks @lognaturel and @gushil!

pyxform/tests_v1/test_trigger.py
lognaturel: Absolutely, I thought it was important to check the hierarchy here but agree that the way I did it is pretty terrible. How about filing an issue?

MartijnR: Yes, will do!

lognaturel:
> JSON wasn't used internally at all

It is as an intermediary step. If someone actually wanted to use the JSON for something directly, they might need to introduce an index from triggers to calculations. We did that after the JSON representation.

pyxform/question.py
lognaturel: Great, I added a commit making sure all question types can be used as calculates. `xml-external` isn't treated as a question type internally.

pyxform/tests_v1/test_typed_calculates.py
lognaturel: Excellent. I didn't use those tests because pyxform represents question types based on body elements and integer, date, geopoint, etc all use the same type. I added a test with a select instead.

MartijnR:
Yes! :party: Thanks!

gushil:
Yes, of course! Thanks @lognaturel @MartijnR

Issue #446, opened 2020-05-26
<output> inside repeat has absolute path instead of relative


MartijnR:
type | name | label::English | calculation
-- | -- | -- | --
begin repeat | member |   |  
calculate | pos |   | position(..)
text | member_name | Name of ${pos} |  
end repeat |   |   |  

[xlsform](https://docs.google.com/spreadsheets/d/1QGpOlm2sTrd-C93Kyr9V2xW7fPghHd6BlaH8Duio4to/edit?usp=sharing)

The above form contains `<output value=" /data/member/pos "/>` instead of the expected `<output value=" ../pos "/>`.

This is only the case when the output is placed as a descendant of `<itext>`. The above form with a `label` column instead of `label::English` works fine.

MartijnR:
Would be great to wait for this fix before the next release **if** we get a PR for this fix soon, say this week. Am discussing with major user about perhaps supporting this as a priority.

gushil:
Hi @MartijnR I've created PR #450 to address this issue. Waiting for the review.

Thanks.

lognaturel:
Thanks again, @gushil! Quick Github tip -- if you use a [closing word](https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords#about-issue-references) before tagging the corresponding issue in a PR description, merging that PR will automatically close the issue.

Issue #403, opened 2019-12-05
Show warning for attempts to use ${} name values in choices sheet.


MartijnR:
<img width="445" alt="Screen Shot 2019-12-05 at 11 24 34 AM" src="https://user-images.githubusercontent.com/627350/70262568-d31e7580-1751-11ea-9b5c-9cc86dc2361f.png">

MartijnR:
Or is there some rogue ODK Collect feature that evaluates ${}? https://forum.opendatakit.org/t/dynamic-choice-list/22903/9?u=martijnr

MartijnR:
Error would be better

lognaturel:
I tried this a while ago in Enketo and it works on some level but doesn't dynamically update. I can't remember the details but I think it's something like it's always one refresh cycle behind as repeat instances are added.

MartijnR:
In this case the XForm output contains `${references}`, and those would not be resolved in Enketo.

lognaturel:
> In this case the XForm output contains ${references}

That's if there are no translations for the form. It used to be that `pyxform` generated an `itext` block and pulled all strings out no matter what. `pyxform` v1 changed that so `itext` is only generated if needed. That's what I explained [on the forum](https://forum.getodk.org/t/xlsform-index-repeat-calculate-filter-choice-odk-collect-v1-27-crash/27411/7?u=ln).

MartijnR:
Oh ok, thanks for clarifying. I missed that. One day I should figure out what is not working in Enketo and create a bug report.

pbowen-oc:
Hi @lognaturel & @MartijnR -
We do not use references in choice names, but we do use them in choice labels to make pseudo-dynamic choice lists. This currently works for us with the minor exception that Pyxform forces us to have some static text in the label even if want it to be solely an item reference. 

We also have cases where we just want to show item text as part of a label to help a user answer questions.

Here is an example with both cases of dynamic choice label text:
[dynamic label test_minimal.xlsx](https://github.com/XLSForm/pyxform/files/4935867/dynamic.label.test_minimal.xlsx)

![DynamicLabels](https://user-images.githubusercontent.com/25179620/87751450-063ae400-c7cc-11ea-8ec9-c2529ac1bfb9.png)

lognaturel:
[...]

PR #450, opened 2020-06-10
Fixed <output> inside repeat has absolute path instead of relative

gushil:
#446

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -593,23 +593,18 @@ class Survey(Section):
             for d in element.get_translations(self.default_language):
 
                 translation_path = d["path"]
-                translation_key = "form"
+                form = "long"
 
                 if "guidance_hint" in d["path"]:
                     translation_path = d["path"].replace("guidance_hint", "hint")
-                    translation_key = "guidance"
+                    form = "guidance"
 
                 self._translations[d["lang"]][translation_path] = self._translations[
                     d["lang"]
                 ].get(translation_path, {})
 
                 self._translations[d["lang"]][translation_path].update(
-                    {
-                        translation_key: {
-                            "text": d["text"],
-                            "output_context": d["output_context"],
-                        }
-                    }
+                    {form: {"text": d["text"], "output_context": d["output_context"],}}
                 )
 
         # This code sets up translations for choices in filtered selects.
@@ -752,7 +747,7 @@ class Survey(Section):
                             )
                         continue
 
-                    if media_type == "form":
+                    if media_type == "long":
                         # I'm ignoring long types for now because I don't know
                         # how they are supposed to work.
                         itext_nodes.append(
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -603,7 +603,14 @@ class Survey(Section):
                     self._translations[d["lang"]][d["path"]] = self._translations[
                         d["lang"]
                     ].get(d["path"], {})
-                    self._translations[d["lang"]][d["path"]].update({"long": d["text"]})
+                    if "element" in d:
+                        self._translations[d["lang"]][d["path"]].update(
+                            {"long": {"text": d["text"], "element": d["element"]}}
+                        )
+                    else:
+                        self._translations[d["lang"]][d["path"]].update(
+                            {"long": d["text"]}
+                        )
 
         # This code sets up translations for choices in filtered selects.
         for list_name, choice_list in self.choices.items():
@@ -722,34 +729,40 @@ class Survey(Section):
                     raise Exception()
 
                 for media_type, media_value in content.items():
-                    # There is a odk/jr bug where hints can't have a value
-                    # for the "form" attribute.
-                    # This is my workaround.
-                    if label_type == "hint":
-                        value, output_inserted = self.insert_output_values(media_value)
-
-                        if media_type == "guidance":
-                            itext_nodes.append(
-                                node(
-                                    "value",
-                                    value,
-                                    form="guidance",
-                                    toParseString=output_inserted,
-                                )
+                    if label_type == "hint" or media_type == "long":
+                        if "text" in media_value and "element" in media_value:
+                            value, output_inserted = self.insert_output_values(
+                                media_value["text"], context=media_value["element"]
                             )
                         else:
+                            value, output_inserted = self.insert_output_values(
+                                media_value
+                            )
+                        # There is a odk/jr bug where hints can't have a value
+                        # for the "form" attribute.
+                        # This is my workaround.
+                        if label_type == "hint":
+                            if media_type == "guidance":
+                                itext_nodes.append(
+                                    node(
+                                        "value",
+                                        value,
+                                        form="guidance",
+                                        toParseString=output_inserted,
+                                    )
+                                )
+                            else:
+                                itext_nodes.append(
+                                    node("value", value, toParseString=output_inserted)
+                                )
+                            continue
+
+                        if media_type == "long":
+                            # I'm ignoring long types for now because I don't know
+                            # how they are supposed to work.
                             itext_nodes.append(
                                 node("value", value, toParseString=output_inserted)
                             )
-                        continue
-
-                    if media_type == "long":
-                        value, output_inserted = self.insert_output_values(media_value)
-                        # I'm ignoring long types for now because I don't know
-                        # how they are supposed to work.
-                        itext_nodes.append(
-                            node("value", value, toParseString=output_inserted)
-                        )
                     elif media_type == "image":
                         value, output_inserted = self.insert_output_values(media_value)
                         if value != "-":
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -591,26 +591,31 @@ class Survey(Section):
         self._translations = defaultdict(dict)  # pylint: disable=W0201
         for element in self.iter_descendants():
             for d in element.get_translations(self.default_language):
+
+                translation_path = d["path"]
+                translation_key = "long"
+
                 if "guidance_hint" in d["path"]:
-                    hint_path = d["path"].replace("guidance_hint", "hint")
-                    self._translations[d["lang"]][hint_path] = self._translations[
-                        d["lang"]
-                    ].get(hint_path, {})
-                    self._translations[d["lang"]][hint_path].update(
-                        {"guidance": d["text"]}
+                    translation_path = d["path"].replace("guidance_hint", "hint")
+                    translation_key = "guidance"
+
+                self._translations[d["lang"]][translation_path] = self._translations[
+                    d["lang"]
+                ].get(translation_path, {})
+
+                if "output_context" in d:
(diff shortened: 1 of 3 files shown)


MartijnR:
@ukanga, just pinging to check if you're able to look at this Ona-reported issue. :)

pyxform/survey.py
```diff
@@ -722,34 +729,40 @@ def itext(self):
...
-
-                        if media_type == "guidance":
-                            itext_nodes.append(
-                                node(
-                                    "value",
-                                    value,
-                                    form="guidance",
-                                    toParseString=output_inserted,
-                                )
+                    if label_type == "hint" or media_type == "long":
```
lognaturel, line 732: Why this test?

pyxform/survey.py
```diff
@@ -603,7 +603,14 @@ def _setup_choice_translations(name, choice_value, itext_id):
                     self._translations[d["lang"]][d["path"]] = self._translations[
                         d["lang"]
                     ].get(d["path"], {})
-                    self._translations[d["lang"]][d["path"]].update({"long": d["text"]})
+                    if "element" in d:
+                        self._translations[d["lang"]][d["path"]].update(
+                            {"long": {"text": d["text"], "element": d["element"]}}
```
lognaturel, line 608: Making the context part of the `long` form doesn't make sense to me (and also it doesn't work for guidance hints). Could it be a sibling to the form? It seems it applies to the lang/path combination and any form should use it.

MartijnR:
Thanks for your review @lognaturel! Just wanted to let you know OpenClinica has slightly delayed deploying the new pyxform version (they use a very old version), and I've heard @gushil may pick this up in a few weeks.

pyxform/survey.py
```diff
@@ -603,7 +603,14 @@ def _setup_choice_translations(name, choice_value, itext_id):
                     self._translations[d["lang"]][d["path"]] = self._translations[
                         d["lang"]
                     ].get(d["path"], {})
-                    self._translations[d["lang"]][d["path"]].update({"long": d["text"]})
+                    if "element" in d:
```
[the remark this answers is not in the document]
gushil: This is for checking whether there is an "element" key in d. the "element" is used for storing a reference to the current translated element that will be used for a context in the insert_output_values method.

pyxform/survey.py
gushil: If you don't mind, any suggestion about it? Changing dictionary key is doable. I'll change that with the next commit.

pyxform/survey.py
gushil: This is just a little code refactor. Will be changed in the next commit.

pyxform/tests_v1/test_repeat.py
```diff
@@ -266,3 +266,22 @@ def test_hints_are_present_within_groups(self):
     </group>"""  # noqa
 
         self.assertPyxformXform(md=md, xml__contains=[expected], run_odk_validate=True)
+
+    def test_inside_repeat_has_relative_path(self):
```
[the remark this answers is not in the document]
gushil: Will do. Thanks for the suggestion.

pyxform/survey.py
lognaturel: Whoops, sorry about that!

pyxform/survey.py
lognaturel: Unfortunately what I had in mind would require some bigger changes to `itext`. Let's go with what you have.

lognaturel:
@MartijnR I don't think there's anything in the spec that specifically addresses `output` in a `text` translation for labels in secondary instances. It seems wacky but people use it in lieu of #38: https://forum.getodk.org/t/xlsform-index-repeat-calculate-filter-choice-odk-collect-v1-27-crash/27411. Is that supposed to be supported? If so, this PR should generate relative refs for those.

MartijnR:
@lognaturel, you're referring to the output below, right? (using the test.xlsx form from that forum thread)

```xml
 <instance id="hm_name">
                <root>
                    <item>
                        <filter>1</filter>
                        <label>${name1}</label><!-- uh oh -->
                        <name>1</name>
                    </item>
```

I tried to figure out a syntax below. Not sure if I succeeded in that, and whether this implementable in pyxform. It wouldn't work in Enketo in the near future. It also wouldn't work for external instances, which to me is a fair argument to not support it all.

It is extremely convoluted xlsform (and xform) syntax indeed. I agree that https://github.com/XLSForm/pyxform/issues/38 would seem so much more sensible.

```xml
<itext>
      <text id="a1">
             <value><output value="../name1"></value>
      </text>
</itext>

 <instance id="hm_name">
                <root>
                    <item>
                        <filter>1</filter>
                        <itextId>a1</itextId>
                        <name>1</name>
                    </item>

...

            <itemset nodeset="instance('hm_name')/root/item[filter &lt; ( /data/qb_1 +1)]">
                <value ref="name" />
                <label ref="itext(itextId)" />
            </itemset>
</select>

```

MartijnR:
Oh LOL https://github.com/XLSForm/pyxform/issues/403

lognaturel:
Ahhh! I knew I'd seen something fly by related to this but couldn't think of where. Let's not worry about it here and instead try to get #38 which would be really great. So I think this is very close to being ready.

pbowen-oc:
@lognaturel @MartijnR - I can't tell from the discussion above - will this update remove the existing behavior that allowed item references in choice labels? We have used this to very good effect (never in a translated form configuration). It is very helpful for allowing the user to (for example) associate a medication and an adverse event or associate a procedure with a previously identified tumor.

In our experience, the only restriction we found was that we needed some text in the label other than the item reference itself.

lognaturel:
Can you please share a minimal XLSForm on #403 that illustrates the pattern you want to use, @pbowen-oc? This PR doesn’t change existing behavior around this so let’s continue the conversation there.

pbowen-oc:
@lognaturel - Thanks, I have added an example file to #403

pyxform/tests_v1/test_repeat.py
```diff
@@ -187,6 +187,25 @@ def test_indexed_repeat_relative_path(self):
             ],
         )
 
+    def test_output_with_translation_relative_path(self):
+        md = """
+        | survey |              |             |                |              |
+        |        | type         | name        | label::English | calculation  |
+        |        | begin repeat | member      |                |              |
+        |        | calculate    | pos         |                | position(..) |
+        |        | text         | member_name | Name of ${pos} |              |
```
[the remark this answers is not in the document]
gushil: Ok, will add this test.

pyxform/survey.py
```diff
@@ -591,19 +591,31 @@ def _setup_choice_translations(name, choice_value, itext_id):
...
-                    self._translations[d["lang"]][hint_path].update(
-                        {"guidance": d["text"]}
+                    translation_path = d["path"].replace("guidance_hint", "hint")
+                    translation_key = "guidance"
+
+                self._translations[d["lang"]][translation_path] = self._translations[
+                    d["lang"]
+                ].get(translation_path, {})
+
+                if "output_context" in d:
```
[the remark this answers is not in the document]
gushil: It looks like some old tests (not in tests_v1) is not passed if I remove this test. However, I've found the cause of those errors and will be fixed by next commit.

pyxform/survey.py
```diff
@@ -591,19 +591,31 @@ def _setup_choice_translations(name, choice_value, itext_id):
...
+                                "output_context": d["output_context"],
+                            }
+                        }
                     )
                 else:
-                    self._translations[d["lang"]][d["path"]] = self._translations[
-                        d["lang"]
-                    ].get(d["path"], {})
-                    self._translations[d["lang"]][d["path"]].update({"long": d["text"]})
+                    self._translations[d["lang"]][translation_path].update(
```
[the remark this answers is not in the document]
gushil: Ok, will remove this.

pyxform/survey.py
```diff
@@ -722,12 +734,18 @@ def itext(self):
...
 
                 for media_type, media_value in content.items():
-                    # There is a odk/jr bug where hints can't have a value
-                    # for the "form" attribute.
-                    # This is my workaround.
-                    if label_type == "hint":
+                    if (
+                        isinstance(media_value, dict)
+                        and "text" in media_value
+                        and "output_context" in media_value
```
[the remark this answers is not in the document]
gushil: Got it.

pyxform/survey.py
```diff
@@ -591,19 +591,31 @@ def _setup_choice_translations(name, choice_value, itext_id):
         self._translations = defaultdict(dict)  # pylint: disable=W0201
         for element in self.iter_descendants():
             for d in element.get_translations(self.default_language):
+
+                translation_path = d["path"]
+                translation_key = "long"
```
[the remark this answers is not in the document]
gushil: It looks like I have to make changes to one other place if I changed translation_key. You'll see in my next commit.

pyxform/survey.py
```diff
@@ -722,12 +734,18 @@ def itext(self):
...
+                    if (
+                        isinstance(media_value, dict)
+                        and "text" in media_value
+                        and "output_context" in media_value
+                    ):
+                        value, output_inserted = self.insert_output_values(
+                            media_value["text"], context=media_value["output_context"]
+                        )
+                    else:
                         value, output_inserted = self.insert_output_values(media_value)
```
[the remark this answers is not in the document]
gushil: Ok

pyxform/survey.py
```diff
@@ -743,15 +752,13 @@ def itext(self):
                             )
                         continue
 
-                    if media_type == "long":
-                        value, output_inserted = self.insert_output_values(media_value)
+                    if media_type == "form":
```
gushil, line 755: I have to make changes to this line if I changed `translation_key` to `form`.

pyxform/survey.py
```diff
@@ -593,7 +593,7 @@ def _setup_choice_translations(name, choice_value, itext_id):
             for d in element.get_translations(self.default_language):
 
                 translation_path = d["path"]
-                translation_key = "long"
+                translation_key = "form"
```
lognaturel, line 596: I meant changing the name of the variable to `form`. `long` and `guidance` are values for the form attribute.

pyxform/survey_element.py
```diff
@@ -271,6 +271,7 @@ def get_translations(self, default_language):
                 for lang, text in constraint_msg.items():
                     yield {
                         "path": self._translation_path("jr:constraintMsg"),
+                        "output_context": self,
```
lognaturel, line 274: Oh, great! There was a bug before because all of these need relative paths as well.

lognaturel:
I took the liberty of renaming the variable that represents the translation form and adding tests for the additional translated columns. Unfortunately it looks like I forgot to run black and I had to quickly leave my computer so things are in a weird state. I’ll fix and merge this evening. 

Thanks for doing this and for taking my feedback, @gushil!

PR #455, opened 2020-06-24
Add Github templates for issues and PRs

lognaturel:
It's hard to review pull requests without context around how much thought a contributor put into their solution, how confident they are about it, what alternatives they considered, etc. The proposed PR template is used across ODK tools and generally serves us well. When I'm on the contribution side, I find it keeps me honest and prevents me from being lazy. When I'm on the reviewer side, it helps me know what to pay attention to and how to think about the solution.

I don't feel as strongly about adding the issue template because issues have generally been fine. I think it can be helpful to have some structure there but @MartijnR files most of the issues and has his own excellent structure. @MartijnR, if the issue template annoys you or you'd like to propose an alternative, I don't mind removing it from this PR.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,21 @@
+<!-- 
+
+Thank you for taking the time to report a pyxform issue!
+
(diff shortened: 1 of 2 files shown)


lognaturel:
> Issue template won't bother me, I think

You can always remove it and keep doing what you're doing! It's mostly there for people who would be tempted to file issues with body text like "doesn't work."

MartijnR:
Exactly! I know, it may be helpful. I still haven't done it for Enketo, because it has been manageable so far and it does increase the threshold for reporting a little. But totally see that for large volumes of issues, it is needed.

Issue #454, opened 2020-06-24
type=calculate without a calculation column but with a dynamic default should be allowed


MartijnR:
Currently there is an error when using:

type | name | label  | calculation | default
-- | -- | -- | -- | --
calculate | a |   |   | random() 

<img width="247" alt="Screen Shot 2020-06-24 at 2 50 22 PM" src="https://user-images.githubusercontent.com/627350/85626396-38708000-b62a-11ea-99ac-cc976d0a0995.png">

But I think this should be allowed as it is the equivalent of 

type | name | label  | calculation | default
-- | -- | -- | -- | --
text | a |   |   | random()

gushil:
@MartijnR PR #456 available to fix this issue.

Thanks.

PR #456, opened 2020-07-11
Fix 454 calculate with dynamic default

gushil:
Closes #
#454 
I think this is the simplest solution.
I don't think there is any.
No


MartijnR:
Thanks! Looks good to me judging from the tests.

Issue #452, opened 2020-06-12
Triggers referring to question without form control, are silently ignored


MartijnR:
This issue refers to the not-yet-released master with trigger support.

type | name | label | trigger | calculation
-- | -- | -- | -- | --
calculate | one |   |   | 5 + 4
calculate | one-ts |   | ${one} | now()
note | note | timestamp: ${one-ts} |   |  

[XLSForm](https://docs.google.com/spreadsheets/d/16Dddg9MXyr-gtXJO68tGtl3oeyLGj2VoSx1NQ4TH2Og/edit?usp=sharing)

In this case there is no `<setvalue>` in the XForm output. This is correct, because ${one} has no form control (i.e. no label and no hint, and therefore no body element).

The form designer should get some feedback. E.g. "Trigger was added for ${one-ts} that refers to hidden question. This is not allowed."

I think an error would be more helpful than a warning in this case. What do others think?

pbowen-oc:
I'd prefer if this were an error. Otherwise, we'll have no way of letting users know that their form won't work as intended.

MartijnR:
Thanks. Looks like this would be ready for a PR.

pbowen-oc:
I think OC can take this.

gushil:
@MartijnR I've created a PR https://github.com/XLSForm/pyxform/pull/457 to resolve this issue. Please review.

Thanks!

PR #457, opened 2020-07-13
Triggers referring to hidden question will produce an error.

gushil:
Closes #
#452 
The simplest one.
None.
None.

### DIFF
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -32,6 +32,16 @@ class Question(SurveyElement):
 
     def xml_control(self):
         if ("calculate" in self.bind or self.trigger) and not (self.label or self.hint):
+            nested_setvalues = self.get_root().get_setvalues_for_question_name(
+                self.name
+            )
+            if nested_setvalues:
+                for setvalue in nested_setvalues:
+                    msg = (
+                        "Trigger was added for ${%s} that refers to hidden question. This is not allowed."
+                        % setvalue[0]
+                    )
+                    raise PyXFormError(msg)
             return None
 
         xml_node = self.build_xml()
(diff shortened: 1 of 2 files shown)


MartijnR:
Thanks Agus! I have not been able to break this.

pyxform/question.py
```diff
@@ -32,6 +32,16 @@ def xml_instance(self, **kwargs):
 
     def xml_control(self):
         if ("calculate" in self.bind or self.trigger) and not (self.label or self.hint):
+            nested_setvalues = self.get_root().get_setvalues_for_question_name(
+                self.name
+            )
+            if nested_setvalues:
+                for setvalue in nested_setvalues:
+                    msg = (
+                        "Trigger was added for ${%s} that refers to hidden question. This is not allowed."
```
lognaturel, line 41: I found this error message hard to parse. How about flipping it around so it's something like "${%s} is not a user-visible question so it can't be used as a calculation trigger for question ${%s}"? The first string would be the trigger (`self.name`) and the second would be the question that the `trigger` is set for (`setvalue[0]`). Hopefully I got that right.

I don't feel too strongly about it. If you prefer the current message, "to hidden question" -> "to a hidden question"

pyxform/tests_v1/test_trigger.py
```diff
@@ -158,6 +158,22 @@ def test_handling_trigger_column_with_no_calculation_no_label_no_hint(self):
...
+            |        | calculate | one    |                      |         | 5 + 4       |
+            |        | calculate | one-ts |                      | ${one}  | now()       |
+            |        | note      | note   | timestamp: ${one-ts} |         |             |
+            """,
+            errored=True,
+            error__contains=[
+                "Trigger was added for ${one-ts} that refers to hidden question. This is not allowed."
+            ],
+        )
+
```
lognaturel, line 176: Were you thinking that a test for a typed calculate is not necessary because `test_non_calculate_type_with_calculation_and_no_label_has_no_control` checks the branch in `question.py`? I'd still add one for completeness (`test_typed_calculate_cant_be_used_as_trigger` or something).

Issue #463, opened 2020-08-14
Duplicate label translations for secondary itemsets


lognaturel:
#### Software and hardware versions 
all

#### Problem description
When a XLSForm definition includes a select with a choice filter, that select is output as a XForm secondary instance. If the choices have labels in multiple languages, the `itext` machinery is used and each label gets included twice in the `translation` block. For example, for [this form](https://docs.google.com/spreadsheets/d/13RlgVIT3U70kmXOWVoWaDsXSFv2UAE7QGnbLwS2ARPQ/edit?usp=sharing):
```
<translation lang="French">
	<text id="/data/foo/a:label">
		<value>Ah</value>
	</text>
	<text id="/data/foo/b:label">
		<value>Bé</value>
	</text>
	<text id="/data/foo/c:label">
		<value>Cé</value>
	</text>
	<text id="static_instance-list-0">
		<value>Ah</value>
	</text>
	<text id="static_instance-list-1">
		<value>Bé</value>
	</text>
	<text id="static_instance-list-2">
		<value>Cé</value>
	</text>
</translation>
```

#### Steps to reproduce the problem
Write a form with a select with translated labels and a choice filter.

#### Expected behavior
Label strings should only be included once per language.

#### Other information 
Note that `itext` and secondary instances aren't complementary. [This ODK forum thread](https://forum.getodk.org/t/proposal-customize-name-and-label-columns-in-selects-from-external-data-in-xlsform/27265/5?u=ln) has more on that. The solution might be to swap all of it out at some point but before clients are all on the same page it would be good to only output each label once.

lognaturel:
Closed by #468

PR #467, opened 2020-09-03
Fixed duplicate label translations for secondary itemsets and use itextID for selects with choices that have media specified

### LINKED ISSUE #464: Select with single-language labels and choice filter can't specify choice images
all

#### Problem description
Images in ODK XForms are displayed using the `itext` machinery. When a select has labels in a single language and a choice filter, the choices are represented as a secondary instance and the labels are taken from the `label` element for each `item` in the instance. In other words, `itext` isn't used. This means there's no way for images to be displayed by clients.

Create an XLSForm with a select that has single-language labels and a choice filter such as [this one](https://docs.google.com/spreadsheets/d/13RlgVIT3U70kmXOWVoWaDsXSFv2UAE7QGnbLwS2ARPQ/edit#gid=1903566534).

#### Expected behavior
It should be possible to specify images for selects with single-language labels.

#### Other information 
If there's media specified for choices, `itext` should always be used for specifying the choice labels. #463 is in the same general space.

DavisRayM:
Closes #464, #463 

None.

No

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -227,11 +227,12 @@ class Survey(Section):
         """
         instance_element_list = []
         multi_language = isinstance(choice_list[0].get("label"), dict)
+        has_media = bool(choice_list[0].get("media"))
         for idx, choice in enumerate(choice_list):
             choice_element_list = []
             # Add a unique id to the choice element in case there is itext
             # it references
-            if multi_language:
+            if multi_language or has_media:
                 itext_id = "-".join(["static_instance", list_name, str(idx)])
                 choice_element_list.append(node("itextId", itext_id))
 
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -228,12 +228,13 @@ class Survey(Section):
         instance_element_list = []
         multi_language = isinstance(choice_list[0].get("label"), dict)
         has_media = bool(choice_list[0].get("media"))
-        for idx, choice in enumerate(choice_list):
+        for choice in choice_list:
             choice_element_list = []
             # Add a unique id to the choice element in case there is itext
             # it references
             if multi_language or has_media:
-                itext_id = "-".join(["static_instance", list_name, str(idx)])
+                name = choice.get("name").replace(" ", "_")
+                itext_id = "-".join(["static_instance", list_name, name])
                 choice_element_list.append(node("itextId", itext_id))
 
             for name, value in sorted(choice.items()):
@@ -611,11 +612,13 @@ class Survey(Section):
         # This code sets up translations for choices in filtered selects.
         for list_name, choice_list in self.choices.items():
             multi_language = isinstance(choice_list[0].get("label"), dict)
-            if not multi_language:
+            has_media = bool(choice_list[0].get("media"))
+            if not multi_language and not has_media:
                 continue
-            for idx, choice in zip(range(len(choice_list)), choice_list):
+            for choice in choice_list:
                 for name, choice_value in choice.items():
-                    itext_id = "-".join(["static_instance", list_name, str(idx)])
+                    name = choice.get("name").replace(" ", "_")
+                    itext_id = "-".join(["static_instance", list_name, name])
                     if isinstance(choice_value, dict):
                         _setup_choice_translations(name, choice_value, itext_id)
                     elif name == "label":
@@ -656,8 +659,14 @@ class Survey(Section):
             self._translations = defaultdict(dict)  # pylint: disable=W0201
 
         for survey_element in self.iter_descendants():
-
-            translation_key = survey_element.get_xpath() + ":label"
+            parent_element = survey_element.get("parent")
+            if parent_element and parent_element.get("list_name"):
+                name = survey_element.get("name").replace(" ", "_")
+                translation_key = "-".join(
+                    ["static_instance", parent_element.get("list_name"), name,]
+                )
+            else:
+                translation_key = survey_element.get_xpath() + ":label"
             media_dict = survey_element.get("media")
 
             # This is probably papering over a real problem, but anyway,
--- a/pyxform/survey_element.py
+++ b/pyxform/survey_element.py
@@ -326,7 +326,9 @@ class SurveyElement(dict):
 
             if type(label_or_hint) is dict:
                 for lang, text in label_or_hint.items():
-                    yield {
+                    parent = self.get("parent")
+                    list_name = parent.get("list_name")
+                    translation = {
                         "display_element": display_element,  # Not used
                         "path": self._translation_path(display_element),
                         "element": self,  # Not used
@@ -334,6 +336,11 @@ class SurveyElement(dict):
                         "lang": lang,
                         "text": text,
                     }
+                    if list_name:
+                        name = self.get("name").replace(" ", "_")
+                        itextId = "-".join(["static_instance", list_name, name])
+                        translation.update({"path": itextId})
+                    yield translation
 
     def get_media_keys(self):
         """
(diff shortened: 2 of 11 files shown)


pyxform/survey.py
```diff
@@ -610,11 +625,14 @@ def _setup_choice_translations(name, choice_value, itext_id):
...
-            if not multi_language:
+            has_media = bool(choice_list[0].get("media"))
+            if not multi_language and not has_media:
                 continue
-            for idx, choice in zip(range(len(choice_list)), choice_list):
+            for choice in choice_list:
                 for name, choice_value in choice.items():
-                    itext_id = "-".join(["static_instance", list_name, str(idx)])
+                    itext_id = "-".join(
+                        ["static_instance", list_name, str(choice.get("name"))]
```
DavisRayM, line 634: By default, the itextId is generated using `list_name` and `index` but I'm not entirely sure of a good way to retrieve the child's index from here. A possibility would be to iterate through the parents 'children'  key and determine this child's index but that seemed like a lot of processing time?

pyxform/survey_element.py
```diff
@@ -369,7 +369,17 @@ def xml_label(self):
...
             # If there is a dictionary label, or non-empty media dict,
             # then we need to make a label with an itext ref
-            ref = "jr:itext('%s')" % self._translation_path("label")
+            parent = self.get("parent")
+            # Make a label with an itext ref in the format
+            # static_instance-<list_name>-<choice_name> if the element
+            # is a choice
+            if parent.get("list_name"):
+                itextId = "-".join(
+                    ["static_instance", parent.get("list_name"), self.get("name")]
```
DavisRayM, line 378: By default, the itextId is generated using list_name and index but I'm not entirely sure of a good way to retrieve the child's index from here. A possibility would be to iterate through the parents 'children' key and determine this child's index but that seemed like it would take a lot of processing time?

pyxform/tests_v1/test_translations.py
```diff
@@ -75,3 +75,41 @@ def test_missing_media_itext(self):
             ],
             itext__excludes=['<value form="audio">jr://audio/-</value>'],
         )
+
+    def test_select_with_choice_filter_translation(self):
```
lognaturel, line 79: This test would pass on `master`, right? You want to make sure to have a test that would have failed so that it tests the new code paths. I think you can do that by removing the `label::Latin` column. I highly recommend using a test-driven approach in this code base -- typically it's possible to add tests without much refactor which is great! That way you can really go through the red-green-refactor cycle.

I would expect at the very least `test_select_with_media_and_choice_filter_and_no_translations_generates_media` and `test_select_with_choice_filter_and_translations_generates_single_translation` or something like that to capture the desired behavior for the two issues you're closing.

pyxform/survey.py
```diff
@@ -227,12 +227,14 @@ def _generate_static_instances(list_name, choice_list):
...
-        for idx, choice in enumerate(choice_list):
+        has_media = bool(choice_list[0].get("media"))
+        for choice in choice_list:
             choice_element_list = []
             # Add a unique id to the choice element in case there is itext
             # it references
-            if multi_language:
-                itext_id = "-".join(["static_instance", list_name, str(idx)])
+            if multi_language or has_media:
+                name = choice.get("name").replace(" ", "_")
```
lognaturel, line 236: Why add this? Wouldn't a name with a space fail earlier because it's not a valid XML identifier? If it's possible to have a name with a space in it, please add a test to cover that case.

pyxform/survey.py
```diff
@@ -227,12 +227,14 @@ def _generate_static_instances(list_name, choice_list):
...
+                itext_id = "-".join(["static_instance", list_name, name])
```
lognaturel, line 237: Is this change from index to name so that all identifiers have a consistent structure? Why is this important? Do you know what the `static_instance` prefix is supposed to mean (I don't). What are the implications of changing this for all translations?

pyxform/xform2json.py
```diff
@@ -546,6 +547,15 @@ def _set_translations(self):
         assert "text" in self.translations[0]
         assert "lang" in self.translations[0]
 
+    def _get_list_name(self, itemset):
```
lognaturel, line 550: This doesn't look right. You should be able to get that instance name from the instance declaration.

lognaturel (changes requested):
Thanks! It would be really helpful to get a high-level view of the decision points you came up to and some risk analysis. There are clearly ramifications if downstream tools use the itext ids in some way. I don't know how likely that is. DId you think through that?

If we're going to be changing the structure of the itext ids, we should consider alternatives. For example, we could use uuids for those identifiers instead of using the list and choice names. I don't really understand the `static_instance` prefix. If it's just noise, could we remove it?

pyxform/survey.py
DavisRayM: It seemed to be possible for this [test](https://github.com/XLSForm/pyxform/blob/f79ade90b8196e1794ff8e078558c298a9771ea1/pyxform/tests_v1/test_inline_translations.py#L44), I'll need to confirm this though. I'll confirm and add a test if this is the case.

pyxform/survey.py
DavisRayM: It's to make the ID construction more predictable. Noticed that the itext entry creation happens in multiple areas https://github.com/XLSForm/pyxform/blob/f79ade90b8196e1794ff8e078558c298a9771ea1/pyxform/survey.py#L594-L610 https://github.com/XLSForm/pyxform/blob/f79ade90b8196e1794ff8e078558c298a9771ea1/pyxform/survey.py#L613-L629 https://github.com/XLSForm/pyxform/blob/f79ade90b8196e1794ff8e078558c298a9771ea1/pyxform/survey.py#L651 using the name makes it a bit easier to recreate the IDs and stop duplicate entries from being created in other areas.

_Not entirely sure what the `static_instance` prefix means... Changing this for all translations related to choices(with multiple languages / media) would mean that all itext entries will have the `static_instance-<list_name>-<name>` format which seemed a bit more consistent and predictable, I don't believe this would affect any other functionality_

pyxform/xform2json.py
DavisRayM: Mhm I might have been looking at the wrong area, I might have to look at the functionality that converts the XML to a dict. By this [point](https://github.com/XLSForm/pyxform/blob/50921a8e48b853565abd48b543558b5e24d6c4f4/pyxform/xform2json.py#L219) the `list_name`info doesn't seem to have been retrieved... Seems to just be a breakdown of the various sections of the XLSForm, didn't contain the `list_name` within it's contents.

DavisRayM:
> Thanks! It would be really helpful to get a high-level view of the decisions you made and some risk analysis. There are clearly ramifications if downstream tools use the itext ids in some way. I don't know how likely that is. Did you think through that?
> 
> If we're going to be changing the structure of the itext ids, we should consider alternatives. For example, we could use uuids for those identifiers instead of using the list and choice names. I don't really understand the `static_instance` prefix. If it's just noise, could we remove it?

I didn't think this would have any ramifications to be honest which might have been a rather naive viewpoint on my part. It seemed like some itext ids for choices where being generated in the `static_instance-<list_name>-<index>` format when a choice had multiple languages or choice filters and others generated using translation path of the element which didn't seem quite consistent which led to additional itext entries being created in other code locations. The `static_instance` format seemed a bit more easier to predict as such I kind of leaned towards that... I believe we can remove it... I'll work on this.

DavisRayM:
Closing this in favor of #468. _I didn't probably document my steps here or my reasoning for some of the changes. Most changes made here are actually not needed in order to fix the tagged issues. #468 is cleaner and I believe a bit more documented._

PR #468, opened 2020-09-09
Fixed duplicate label translations for secondary itemsets and use itextID for selects with choices that have media specified

### LINKED ISSUE #464: Select with single-language labels and choice filter can't specify choice images

DavisRayM:
Based on #467 
Closes #464, #463 

The simplest approach(_Least amount of modification to already existing functionality_). Ensures that we only generate translations for choices in filtered selects in a controlled manner and that itext is used for selects with media

Considered utilizing uuids, but decided otherwise since uuid generation is unpredictable; _Translation entries are created/linked in multiple areas i.e `_setup_media` & `_setup_translations` using uuids may lead to more duplicate entries and I'm not entirely sure of an easy way to implement uuids and correctly link them to the question elements_

None

None


pyxform/survey.py
```diff
@@ -227,12 +227,13 @@ def _generate_static_instances(list_name, choice_list):
         """
         instance_element_list = []
         multi_language = isinstance(choice_list[0].get("label"), dict)
+        has_media = bool(choice_list[0].get("media"))
         for idx, choice in enumerate(choice_list):
             choice_element_list = []
             # Add a unique id to the choice element in case there is itext
             # it references
-            if multi_language:
-                itext_id = "-".join(["static_instance", list_name, str(idx)])
```
DavisRayM, line 235: The `static_instance` prefix seems to be a way of denoting that the data held within the `itext` is static data related to an instance.

lognaturel: You're probably right. Not clear why specify static -- I don't think there's a dynamic alternative. Anyway, removing the prefix seems reasonable to me.

@MartijnR do you have any knowledge of a tool that could be relying on the names of itext elements in some way or does it seem safe to treat like an implementation detail that can be modified?

pyxform/survey.py
```diff
@@ -702,6 +707,25 @@ def _setup_media(self):
...
+            self._translations = defaultdict(dict)  # pylint: disable=W0201
+
+        for survey_element in self.iter_descendants():
+            parent = survey_element.get("parent")
+            if parent and not parent.get("choice_filter"):
+                translation_key = survey_element.get_xpath() + ":label"
+                media_dict = survey_element.get("media")
+                _set_up_media_translations(media_dict, translation_key)
+
+        # This code sets up media for choices in filtered selects.
```
lognaturel, line 720: Currently I can comment out this entire code block and all tests pass.

pyxform/survey.py
```diff
@@ -702,6 +707,25 @@ def _setup_media(self):
 
                     translations_trans_key[media_type] = media
 
+        if not self._translations:
+            self._translations = defaultdict(dict)  # pylint: disable=W0201
+
+        for survey_element in self.iter_descendants():
+            parent = survey_element.get("parent")
+            if parent and not parent.get("choice_filter"):
```
lognaturel, line 715: Currently I can comment out the two lines above and all tests pass.

lognaturel (changes requested):
Really liking those tests, @DavisRayM, thanks for the simplified PR! My review process is to look at the tests first, then go through the code and if I can't explain a section, comment it out and run the test suite to see what it does. I've identified two code sections that I don't understand and can currently be commented out without any tests failing. Please illustrate the cases they cover with tests if they are indeed needed.

I like the removal of the `static_instance` prefix but it seems possibly risky so I'd like another reviewer's thoughts on it.

One difference between your prior PR and this one is that the other one updates `xform2json` to read in secondary itemset translations. I generally consider `xform2json` experimental and unsupported so I don't think it's necessary to update. I just wanted to flag it in case that was something you intended to do.

pyxform/survey.py
DavisRayM: Updated the tests to cover this. _This check ensures that the translation is only set up once for choices in selects with choice_filter._

pyxform/survey.py
DavisRayM: Initially, thought this change https://github.com/XLSForm/pyxform/pull/468/files/0963af66a2c696b5655f97430d8800630c24a6f0#diff-642ca628a5111f8d9ff1314acb61ceadR714-R715

would break the translation generation for media files but that doesn't seem to be the case. Seems the translations are still setup in `_setup_translations` I haven't noticed anything quite unique in the setup for the media here, as such I've removed this change in the latest commits

pyxform/survey.py
MartijnR: No, I don't have any knowledge about relying on the `id` attribute to have a predictable name. I think it's safe to change.

DavisRayM:
> Really liking those tests, @DavisRayM, thanks for the simplified PR! My review process is to look at the tests first, then go through the code and if I can't explain a section, comment it out and run the test suite to see what it does. I've identified two code sections that I don't understand and can currently be commented out without any tests failing. Please illustrate the cases they cover with tests if they are indeed needed.
> 
> I like the removal of the `static_instance` prefix but it seems possibly risky so I'd like another reviewer's thoughts on it.
> 
> One difference between your prior PR and this one is that the other one updates `xform2json` to read in secondary itemset translations. I generally consider `xform2json` experimental and unsupported so I don't think it's necessary to update. I just wanted to flag it in case that was something you intended to do.

Thanks, regarding the `xform2json` changes I think I'll need to research a bit more into the functionality. Currently, with these changes the secondary itemset translations don't seem to be a problem(the JSON generated version of the `specify_other.xls` fixture seems to be passing the test). I'll look into it though and open an issue if I spot an issue with the secondary itemset translation on that functionality.

pyxform/survey.py
lognaturel: Great! That seems right -- media's treated the same as any other translation.

pyxform/survey.py
lognaturel: Thanks for the sanity check, @MartijnR!

pyxform/tests_v1/test_inline_translations.py
```diff
@@ -74,3 +74,75 @@ def test_multiple_translations(self):
...
+            md=xform_md,
+            errored=False,
+            debug=False,
+            itext__contains=[
+                '<text id="list-0">',
+                '<text id="list-1">',
+                '<text id="list-2">',
+            ],
+            itext__excludes=[
+                '<text id="/data/foo/a:label">',
```
lognaturel, line 144: If I change the test to remove the last angle brackets for these `text` items, the test fails. That is, it's now generating 
```
<text id="/data/foo/a:label"/>
<text id="/data/foo/b:label"/>
<text id="/data/foo/c:label"/>
```
whereas those items shouldn't be there at all.

lognaturel (approved):
I now feel I can fully explain this. It makes a lot of sense and seems low-risk. Thanks, @DavisRayM! It's great to see some redundancy go away.

Sorry if you saw a weird comment about a remaining issue come through -- I wasn't on the latest commit.

PR #469, opened 2020-09-09
V1.2 prep

lognaturel:
This is a routine update with not much choice. I think this is a good time to remove the repeat warning because the change happened several months ago and it's confusing to folks who didn't build forms before the change. I spent a bit of time seeing if there were any other outdated messages but I can't think of any.

JavaRosa's calculation code was dramatically simplified. The latest release has bee in Collect betas for some time and has gone through careful review so I think the risk to validation behavior is low.

No.

- [N/A] included test cases for core behavior and edge cases in `tests_v1`


yanokwa (approved):
LGTM. I think the codecov complaint isn't important, so I'm ignoring it.

PR #473, opened 2020-10-11
Issue 471 deprecate unused device id methods

### LINKED ISSUE #471: Display warning when using simserial or subscriberid
Related docs issue: https://github.com/XLSForm/xlsform.github.io/pull/208

Web-based clients such as Enketo have never been able to read permanent unique identifiers like simserial or subscriberid. Neither have iOS clients for a long time. Android clients used to be able to access them but won't be able to do so from Play Store installs starting Nov 2020.

That means the only clients that will be able to access these identifiers are Android clients that target and run on Android versions < 11 (e.g. old versions of Collect). 

<strike>@MartijnR pointed out at https://github.com/XLSForm/xlsform.github.io/pull/208#issuecomment-693499567 that maybe we should show an error when a user attempts to use these. Given that some folks might intentionally stay on old client versions to access this behavior, I have a slight preference for warning.

- [x] decide whether to warn or error
- [x] decide on text</strike>

From https://github.com/XLSForm/pyxform/issues/471#issuecomment-693531176:
Warning with text "Simserial is no longer supported on most devices. Only old versions of Collect on Android versions older than 11 still support it." (same for subscriberID)

KeynesYouDigIt:
Closes #471 

This adds warnings where other question type specific warnings are. This pattern is modelled after the special validation work used for 'audit' and 'calculate' question types.

I considered checking this information upstream in the survey creation process, but would have to re-divide the sheet into rows. Downstream, during question creation and final survey creation, is another place where warnings can be generated, but we have the information we need at the point where the warnings were added, so no point in working further down.

I did modify constants.py - any modification of a global config represents some regression risk. However, values were only added, and only values that apply to my use case, and only under a name that does not conflict with another name, so I do not see any great risks.

docs should already be set https://github.com/XLSForm/xlsform.github.io/pull/208/files

- [ x ] included test cases for core behavior and edge cases in `tests_v1`
- [ x ] run `nosetests` and verified all tests pass
- [ x ] run `black pyxform` to format code
- [ x ] verified that any code or assets from external sources are properly credited in comments


pyxform/tests/xlsform_spec_test.py
```diff
@@ -100,7 +100,7 @@ def runTest(self):
             path_to_excel_file, default_name="warnings", warnings=warnings
         )
         self.assertEquals(
-            len(warnings), 21, "Found " + str(len(warnings)) + " warnings"
+            len(warnings), 22, "Found " + str(len(warnings)) + " warnings"
```
KeynesYouDigIt, line 103: I thought about just removing the simserial reference from the excel file used in this test, is that a better way to go since simserial is technically not a part of the spec anymore? Should this file be updated for new spec stuff?

pyxform/tests_v1/test_metadata.py
```diff
@@ -0,0 +1,83 @@
...
+from pyxform.tests_v1.pyxform_test_case import PyxformTestCase
+
+
+class MetadataTest(PyxformTestCase):
+    """
+    Test metadata and related warnings.
+    """
+
+    def test_metadata_bindings(self):
+        self.assertPyxformXform(
```
KeynesYouDigIt, line 17: boy scout rule :) thought I would throw a test in for these fields getting their `preloadParams`, LMK if this is unhelpful in any way.

pyxform/tests/xlsform_spec_test.py
lognaturel: I like your idea of removing it since it's no longer part of the spec!

pyxform/tests_v1/test_metadata.py
lognaturel: 👍 I like it.

pyxform/tests_v1/test_metadata.py
```diff
@@ -0,0 +1,83 @@
...
+        self.assertEqual(warning_expected, warnings[0])
+        os.unlink(tmp.name)
+
+    def test_subscriber_id_deprecation_warning(self):
+        warnings = []
+        survey = self.md_to_pyxform_survey(
+            """
+            | survey |              |                          |                                            |
+            |        | type         | name                     | label                                      |
+            |        | subscriberid | subscriberid             | sub id is not lable optional I guess       |
```
lognaturel, line 69: > sub id is not lable optional I guess

It should be. Does it fail with a real form converted by pyxform master without a label?

lognaturel (changes requested):
Looking good!

pyxform/tests_v1/test_metadata.py
KeynesYouDigIt: its just a warning not an error, but I get this warning. I will clarify that in the label and comments

```row : 2] Question has no label: {'name': 'subscriberid', 'type': 'subscriberid'}```

KeynesYouDigIt: I could also do the work to remove this warning, but it doesnt seem pertinent since sub id is being deprecated

pyxform/tests/xlsform_spec_test.py
KeynesYouDigIt: Should be all done!

lognaturel (approved):
Super, thanks so much!!

PR #479, opened 2020-11-14
Java versions can have patch versions

yanokwa:
Resolves https://forum.getodk.org/t/problem-converting-with-xlsform-online-due-to-too-many-values-to-unpack/31163. This problem is occuring because the default Java on Ubuntu was upgraded to version "11.0.9.1" on 2020-11-04. Our code assumes no patch version.

This change will require a hotfix to pyxform.

Narrowest patch I could think of. I assume there is a major and minor version, and that both are ints. I also assume that the other version information doesn't matter. 

I remove the int wrappers in favor of the int() wrapping in the check that raises the exception

No.


lognaturel (approved):
I think doing a patch with this is good but we should change the strategy. The pythonic way would be to try to use Validate and if it fails, put up an error.

PR #482, opened 2020-11-14
[WORK IN PROGRESS] warn about mising translations issue 157

KeynesYouDigIt:
Closes #

not done yet

- [ ] included test cases for core behavior and edge cases in `tests_v1`
- [ ] run `black pyxform` to format code


pyxform/survey_element.py
```diff
@@ -297,43 +297,29 @@ def get_translations(self, default_language):
...
-                and not (isinstance(label_or_hint, dict))
-                and len(label_or_hint) > 0
-            ):
-                label_or_hint = {default_language: label_or_hint}
-
-            # always use itext for hint if there's a guidance hint
-            if (
-                display_element == "hint"
-                and not (isinstance(label_or_hint, dict))
-                and len(label_or_hint) > 0
```
lognaturel, line 321: Could you please add a test to confirm that removing this doesn't result in an extra translation?

pyxform/survey_element.py
```diff
@@ -297,43 +297,29 @@ def get_translations(self, default_language):
...
-                    yield {
-                        "display_element": display_element,  # Not used
-                        "path": self._translation_path(display_element),
-                        "element": self,  # Not used
-                        "output_context": self,
-                        "lang": lang,
-                        "text": text,
-                    }
+            if label_or_hint:
+                if not (
```
lognaturel, line 301: While you're in here, could you change this to `not(type(label_or_hint) is dict)` to match the test below? It's weird to have the same test expressed differently.

lognaturel (commented):
I agree that this is much clearer. Just one change I think is worth having a test for.

pyxform/survey_element.py
```diff
@@ -297,43 +297,29 @@ def get_translations(self, default_language):
...
-                and label_or_hint
-            ):
-                label_or_hint = {default_language: label_or_hint}
-
-            # always use itext for guidance hints because that's
-            # how they're defined - https://opendatakit.github.io/xforms-spec/#languages
-            if (
-                display_element == "guidance_hint"
-                and not (isinstance(label_or_hint, dict))
-                and len(label_or_hint) > 0
```
lognaturel, line 313: Could you please add a test to confirm that removing this doesn't result in an extra translation? The test form could have two questions, one with a guidance hint and one without. A guidance hint should only be generated for the question that has one.

KeynesYouDigIt: How do these look? I wasn't 100% sure on how to confirm there was _only_ one translation https://github.com/XLSForm/pyxform/pull/482/commits/7427d7d40cd801d709adb87bd922d4aa2d4f1000

KeynesYouDigIt: When I run with Debug=True the xml does look correct (no extra translation elements)

KeynesYouDigIt (commented):
dangit git hub these are responses not a review lol

pyxform/survey_element.py
KeynesYouDigIt: Is this what you meant? https://github.com/XLSForm/pyxform/pull/482/files#diff-f80582185a2782560583edf22dd03e637736b335c77eaf8b69d264af87798153R79

pyxform/survey.py
```diff
@@ -363,9 +363,9 @@ def get_instance_info(element, file_id):
             uri = "jr://file-csv/{}.csv".format(file_id)
 
             return InstanceInfo(
-                type=u"pulldata",
+                type="pulldata",
```
KeynesYouDigIt, line 366: Linter stuff I think. LMK if I should revert.

pyxform/survey.py
```diff
@@ -687,8 +690,37 @@ def _add_empty_translations(self):
...
+                                " Set a default language in the settings tab, or specifiy the language of this column."
+                            )
+                            if lang == "default"
+                            else (f"Translation for {lang} missing for: {column_name}")
+                        )
+                        self._add_warning(missing_translation_warning)
                         self._translations[lang][path][content_type] = "-"
 
+    def _get_column_name_from_translation_path(self, path, content_type):
+        if content_type not in ["long", "guidance"]:
```
KeynesYouDigIt, line 708: Thing 1 I don't like about this approach - were just guessing at the column name, trying to make it more user friendly. Could this miss for a significant amount of content_types or elements? Should I test the whole dang spec just to see what we get? (not sarcasm, a real offer if we think its worth it)

pyxform/survey_element.py
```diff
@@ -70,8 +70,13 @@ class SurveyElement(dict):
         "action": unicode,
         "list_name": unicode,
         "trigger": unicode,
+        # This is a simple data store allowing internal logic to store warning messages.
+        "__internal_warnings__": set,
```
KeynesYouDigIt, line 74: Thing 2 I don't like about this approach - we just kind of hold a stateful thing called warnings then slap it on at the end of everything. The only other option I could think to do is to take in the warnings list that was passed initially and pass it around _everywhere_ but that was frowned upon in my last PR, cause it required a lot of changes (and isnt really any better).

The _other_ option is to assign warnings to a key like this as soon as its passed, but again, thats not really any better from what I can see.

pyxform/tests_v1/test_language_warnings.py
```diff
@@ -89,6 +88,45 @@ def test_default_language_only_should_not_warn(self):
...
+        self.assertTrue(len(warnings) == 8)
+        self.assertIn(
+            "Translation for English (en) missing for: jr:constraintMsg", warnings
+        )
+        self.assertIn("Translation for English (en) missing for: image", warnings)
+        self.assertIn("Translation for English (en) missing for: hint", warnings)
+        self.assertIn("Translation for Spanish (es) missing for: image", warnings)
+        self.assertIn("Translation for Spanish (es) missing for: label", warnings)
+        self.assertIn(
+            "There is no default language set, and no language specified for: choice label for opts, Set a default language in the settings tab, or specifiy the language of this column.",
```
KeynesYouDigIt, line 121: Too wordy? This is the case that created the original issue, so I figured it was worth being clear

PR #483, opened 2020-11-18
Add quality attribute for audio questions

seadowg:
This adds support for the `quality` attribute detailed [here](https://forum.getodk.org/t/form-spec-proposal-configure-audio-quality-for-internal-recording/).

Didn't seem like there were a lot of different ways to go but definitely keen to hear criticism as I'm pretty new to the codebase.

Only audio questions have been changed so that's the only area where something could go wrong (I'd imagine).

Yes but that's been planned separately!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -90,12 +90,15 @@ You can run tests with::
 
     nosetests
 
-
 Before committing, make sure to format your code using `black`::
 
     black pyxform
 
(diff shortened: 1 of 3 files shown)


lognaturel (dismissed):
Good to see a README section about tests.

MartijnR (approved):
Looks good to me 👍

lognaturel (approved):

PR #487, opened 2020-12-04
Support ref in constraint and required messages with single language

### LINKED ISSUE #474: If constraint_message includes a ${} reference and no translations, expanded XPath path should be in <output>
pyxform v1.x.x, Python v1.2.0

#### Problem description
If a form only uses one language and has a `constraint_message` with a ${} reference, no `output` is generated. This is a side effect of https://github.com/XLSForm/pyxform/issues/285.

1. Build a single-language form with a constraint message that has a ${} reference in it:

type | name | label | constraint | constraint_message
-- | -- | -- | -- | --
select_one fruits | Q1 | Select the fruit that you like the most: |   |  
select_one fruits | Q2 | Select (again) the (same) fruit that you like the most (that   you selected in Q1): | .=${Q1} | You chose ${Q1} in Q1 but now it's different!

1. Convert it and see that the constraint message is set to `jr:constraintMsg="You chose  /data/Q1  in Q1 but now it's different!"`. This means the `/data/Q1` reference is not evaluated by clients.

#### Expected behavior
`/data/Q1` gets evaluated by clients.

#### Other information 
A form design workaround is to set a language for the `constraint_message` column to force the generation of an `itext` block. I think the only approach we could take is detect whether there's a `${}` reference in a `constraint_message` and if so generate `itext` for it.

lognaturel:
Closes #474

Discussed approach in the issue and @MartijnR agreed identifying `${}` references and forcing itext creation in that case was our best bet.

In terms of implementation, I tried to reduce redundancy in this section but I couldn't come up with any approach that would be easy to read. I think this is ok because the conditions are linear and read well.

I considered adding tests for relative references in repeats but I think that's sufficiently covered elsewhere and should Just Work.

Everything I added is in a conditional so I think risks are minimal. The worst risk is probably that I would have forgotten some cases.

No, this is a bug fix.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -7,13 +7,17 @@ from __future__ import print_function
 import codecs
 import os
 import re
-from re import split
 import tempfile
 import xml.etree.ElementTree as ETree
 from collections import defaultdict
 from datetime import datetime
 
 from pyxform import constants
+from pyxform.utils import (
+    BRACKETED_TAG_REGEX,
+    LAST_SAVED_REGEX,
+    LAST_SAVED_INSTANCE_NAME,
+)
 from pyxform.errors import PyXFormError, ValidationError
 from pyxform.external_instance import ExternalInstance
 from pyxform.instance import SurveyInstance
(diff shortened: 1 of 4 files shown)


MartijnR (approved):
Thanks!

Looks great to me. Did some real-life testing.

PR #489, opened 2020-12-04
Prevent untyped calculates from being used as calculation triggers

### LINKED ISSUE #486: type=calculate with (ignored) label does not show an error when used as trigger
pyxform v1.2.1

#### Problem description
Users may accidentally add a label to type=calculate. This label is mostly ignored by pyxform and not present in the Pyxform output. That's all good. However, the helpful error message generation logic introduced in https://github.com/XLSForm/pyxform/issues/452 actually gets fooled by this otherwise ignored label.

type | name | label | calculation | trigger
-- | -- | -- | -- | --
calculate | calc2 | calculate item label | 1+1 |  
text | text2 | triggered from calc2 | 4+4 | ${calc2}

#### Expected behavior

To output the same pyxform error message as with this form (since they are essentially exactly the same form) :

type | name | label | calculation | trigger
-- | -- | -- | -- | --
calculate | calc2 |  | 1+1 |  
text | text2 | triggered from calc2 | 4+4 | ${calc2}

#### Other info

I also considered an alternative solution where a type=calculate with a label gets a warning. However, since that should probably be a warning and not an error, I think we would still need the above error that prevents the XForm from being created.

lognaturel:
Closes #486

@gushil I took this on to hopefully give you a little extra time for #485. I think you're best positioned to see if there's a way to make it a little easier to work with. I should hopefully also have a bit of time for it tomorrow.

I didn't consider any alternatives. This extends the existing logic to treat any question with a `calculate` type as ineligible to be a calculation trigger.

The only change is to the logic that for a given question suppresses body tags and produces an error if it's used as a trigger. The worst case would be if this adds cases that are actually suppose to lead to body elements

No.

### DIFF
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -32,7 +32,9 @@ class Question(SurveyElement):
         return node(self.name, **attributes)
 
     def xml_control(self):
-        if ("calculate" in self.bind or self.trigger) and not (self.label or self.hint):
+        if self.type == "calculate" or (
+            ("calculate" in self.bind or self.trigger) and not (self.label or self.hint)
+        ):
             nested_setvalues = self.get_root().get_setvalues_for_question_name(
                 self.name
             )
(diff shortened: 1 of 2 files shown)


gushil:
Thanks @lognaturel

MartijnR (approved):
Perfekt. Thank you!

PR #493, opened 2020-12-09
Make value ref dynamic for selects from repeats

lognaturel:
#472 added support for selects from repeats but it did not make the `value` ref dynamic so only fields with the name `name` could be used.

This matches the way the `label` ref is set and is simple to understand.

One risk I thought of is that if `itemset_value_ref` is in the wrong place, some select cases could end up without a `ref` set.

No.


DavisRayM (approved):
LGTM

PR #496, opened 2020-12-12
Add warning for images without max-pixel

### LINKED ISSUE #475: Add size and bandwidth warning to image questions
The default image sizes these days are huge and we are unnecessarily straining bandwidth and storage (both on local devices and on servers). 

I propose we warn users who have the image data type that they can resize images by using max-pixels (as described at https://getodk.github.io/xforms-spec/#bind-attributes and https://docs.getodk.org/form-question-types/#scaling-down-images).

There are many places to help users reduce file size (e.g., set a default max-pixel unless form designer overrides, add some UI all the form design tools to suggest max-pixel, resize by default in Collect/Enketo) but none of those places solve the problem elegantly and across the whole ecosystem.

One concern I have is that this is more a suggestion, than a warning, but it's also kind of a warning, so 🤷🏾‍♂️. Something like this would work for me.
> Consider specifying a `max-pixel` value to scale down the image and reduce storage and bandwidth requirements. Learn more at https://xlsform.org/en/#image

yanokwa:
Closes #475

It's the narrowest change. I used the shortest possible warning to make it easier to read.

Note there is one unrelated commit to make the Learn/read more consistent throughout pyxform.

None that I can think of. It's just a warning. If you have a lot of images, you could get a lot of these warnings. I toyed with the idea of de-duping it so you only get one generic one, but I think row numbers are more helpful.

No.

### DIFF
--- a/pyxform/tests/xlsform_spec_test.py
+++ b/pyxform/tests/xlsform_spec_test.py
@@ -100,7 +100,7 @@ class WarningsTest(unittest.TestCase):
             path_to_excel_file, default_name="warnings", warnings=warnings
         )
         self.assertEquals(
-            len(warnings), 21, "Found " + str(len(warnings)) + " warnings"
+            len(warnings), 22, "Found " + str(len(warnings)) + " warnings"
         )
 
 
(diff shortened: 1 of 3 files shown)


lognaturel (approved):
Note that this could be annoying for folks who use client settings to configure this. I think in general we do want to steer folks towards making the form as specific with these kinds of requirements as possible. Tagging @danbjoseph in case he wants to chime in since I do remember he told me he used the setting instead of this.

PR #499, opened 2020-12-17
Fixed indexed-repeat argument path evaluation in integer column type

### LINKED ISSUE #484: When indexed-repeat is used within a repeat the first argument nodeset points to single node using relative path
2.2.1

#### Problem description
The first argument of indexed-repeat() calls is relative when called from inside a repeat and referring to a question in that repeat. It should be absolute.

A `../relative/path/to/question` always returns a single node (in ODK XForms) and according to [the spec](https://getodk.github.io/xforms-spec/#fn:indexed-repeat) it should refer to all nodes with that question name, i.e. the nodeset returned by `/absolute/path/to/question`.

_A. using regular calculation with relative index:_

type | name | label | calculation
-- | -- | -- | --
begin_repeat | person | Person |  
integer | pos | | position(..) |
text | name | Enter name |  
text | prev_name | Name in previous repeat instance | indexed-repeat(${name}, ${person}, ${pos}-1)
end_repeat |   |   |  

outputs: 

```xml
<bind calculate="indexed-repeat( ../name ,  /data/person , ../pos -1)"  ... />
```

_B. using dynamic default:_

type | name | label | default
-- | -- | -- | --
begin_repeat | person | Person |  
text | name | Enter name |  
text | prev_name | Name in previous repeat instance | indexed-repeat(${name}, ${person}, position(..)-1)
end_repeat |   |   |  

outputs:

```xml
<setvalue ... value="indexed-repeat( ../name ,  /data/person , position(..)-1)" />
```

_C. using dreaded nested repeats:_

type | name | label | default
-- | -- | -- | --
begin_repeat | family | Family |  
integer | members_number | How many members in this family? |  
begin_repeat | person | Person |  
text | name | Enter name |  
text | prev_name | Non-sensible previous name in first family, 2nd person | indexed-repeat(${name}, ${family}, 1, ${person}, 2)
end_repeat |   |   |  
end_repeat |   |   |  

outputs:

```xml
<setvalue .. value="indexed-repeat( ../name ,  /data/family , 1,  ../../person , 2)" />
```

#### Expected behavior

A:

```xml
<bind calculate="indexed-repeat( /data/person/name ,  /data/person , ../pos - 1)" ... />
```

B: 
```xml
<setvalue ... value="indexed-repeat( /data/person/name ,  /data/person , position(..)-1)" />
```

C: 
```xml
<setvalue .. value="indexed-repeat( /data/family/person/name ,  /data/family , 1, /data/family/person , 2)" />
```

So the first argument of indexed-repeat() is an exception to the relative-path rule used to transform ${node} syntax (and the second argument too but that seems to already be the case).

Or actually since indexed-repeat() has either 3, 5, or 7 arguments, this exception applies to the 1st, 2nd, 4th, and 6th  arguments.

#### Other information 

Previously using indexed-repeat() inside a repeat was a bit of an edge case. Since the introduction of dynamic defaults it is becoming more common, as you could use the answer of a previous repeat question as the dynamic default of that same question in a new repeat instance (example B).

I'm guessing this issue was introduced in 0.12.0 when we changed to relative node references inside repeats.

### LINKED ISSUE #507: dynamic default expression causes KeyError
pyxform v1.3.3

#### Problem description
Follow form produces an unhelpful error. 

type | name | label | calculation | default
-- | -- | -- | -- | --
calculate | r |   |   | random() + 0.5
calculate | one |   |   | if(${r} < 1,'A','B')

Screen Shot 2020-12-28 at 3 05 33 PM

The full error is:

```bash
  File "/Users/martijnvanderijdt/.virtualenvs/local_pyxform/bin/xls2xform", line 11, in <module>
    load_entry_point('pyxform', 'console_scripts', 'xls2xform')()
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/xls2xform.py", line 177, in main_cli
    enketo=args.enketo_validate,
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/xls2xform.py", line 49, in xls2xform_convert
    enketo=enketo,
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/survey.py", line 1055, in print_xform_to_file
    raise error
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/survey.py", line 1051, in print_xform_to_file
    file_obj.write(self._to_ugly_xml())
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/survey.py", line 838, in _to_ugly_xml
    return '<?xml version="1.0"?>' + self.xml().toxml()
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/survey.py", line 243, in xml
    node("h:head", node("h:title", self.title), self.xml_model()),
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/survey.py", line 546, in xml_model
    model_children += self.xml_bindings()
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/survey_element.py", line 479, in xml_bindings
    dynamic_default = e.get_setvalue_node_for_dynamic_default()
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/survey_element.py", line 369, in get_setvalue_node_for_dynamic_default
    default_with_xpath_paths = self.get_root().insert_xpaths(self.default, self)
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/survey.py", line 996, in insert_xpaths
    return re.sub(BRACKETED_TAG_REGEX, _var_repl_function, unicode(text))
  File "/Users/martijnvanderijdt/.virtualenvs/local_pyxform/lib/python3.7/re.py", line 192, in sub
    return _compile(pattern, flags).sub(repl, string, count)
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/survey.py", line 993, in _var_repl_function
    matchobj, context, use_current, reference_parent
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/survey.py", line 976, in _var_repl_function
    if _is_return_relative_path():
  File "/Users/martijnvanderijdt/Code/pyxform/pyxform/survey.py", line 960, in _is_return_relative_path
    and "indexed-repeat" in context["bind"]["calculate"]
KeyError: 'calculate'
```

transform above form, also available [here](https://docs.google.com/spreadsheets/d/1Mj3mT7fVp8yjXE6-ffzI-PD_5BNk35SoMzeRjhn661c/edit#gid=0)

#### Expected behavior

For this expression to be accepted without an error.

#### Other information 

I have thoughts about considering type=calculate to be an alias to type=text (and type=hidden) (which doesn't have this problem). https://github.com/XLSForm/pyxform/issues/508

gushil:
Closes #484 
Closes #507
Previous solution (PR #485) with added evaluation for integer type column.
None.
No.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -912,50 +912,45 @@ class Survey(Section):
             current_matchobj = matchobj
 
             if not last_saved and context:
-                if context["type"] != "calculate":
 
-                    if not is_indexed_repeat:
-                        return True
+                if not is_indexed_repeat:
+                    return True
 
-                    # It is possible to have multiple indexed-repeat in an expression
-                    indexed_repeats_iter = indexed_repeat_regex.finditer(
-                        matchobj.string
-                    )
-                    for indexed_repeat in indexed_repeats_iter:
+                # It is possible to have multiple indexed-repeat in an expression
+                indexed_repeats_iter = indexed_repeat_regex.finditer(matchobj.string)
+                for indexed_repeat in indexed_repeats_iter:
 
-                        # Make sure current ${name} is in the correct indexed-repeat
-                        if current_matchobj.end() > indexed_repeat.end():
-                            try:
-                                next(indexed_repeats_iter)
-                                continue
-                            except StopIteration:
-                                return True
-
-                        # ${name} outside of indexed-repeat always using relative path
-                        if (
-                            current_matchobj.end() < indexed_repeat.start()
-                            or current_matchobj.start() > indexed_repeat.end()
-                        ):
+                    # Make sure current ${name} is in the correct indexed-repeat
+                    if current_matchobj.end() > indexed_repeat.end():
+                        try:
+                            next(indexed_repeats_iter)
+                            continue
+                        except StopIteration:
                             return True
 
-                        indexed_repeat_name_index = None
-                        indexed_repeat_args = (
-                            function_args_regex.match(indexed_repeat.group())
-                            .group(1)
-                            .split(",")
-                        )
-                        name_arg = "${{{0}}}".format(name)
-                        for idx, arg in enumerate(indexed_repeat_args):
-                            if name_arg in arg.strip():
-                                indexed_repeat_name_index = idx
+                    # ${name} outside of indexed-repeat always using relative path
+                    if (
+                        current_matchobj.end() < indexed_repeat.start()
+                        or current_matchobj.start() > indexed_repeat.end()
+                    ):
+                        return True
 
-                        return (
-                            indexed_repeat_name_index is not None
-                            and indexed_repeat_name_index
-                            not in indexed_repeat_relative_path_args_index
-                        )
-                else:
-                    return not ("indexed-repeat" in context["bind"]["calculate"])
+                    indexed_repeat_name_index = None
+                    indexed_repeat_args = (
+                        function_args_regex.match(indexed_repeat.group())
+                        .group(1)
+                        .split(",")
+                    )
+                    name_arg = "${{{0}}}".format(name)
+                    for idx, arg in enumerate(indexed_repeat_args):
+                        if name_arg in arg.strip():
+                            indexed_repeat_name_index = idx
+
+                    return (
+                        indexed_repeat_name_index is not None
+                        and indexed_repeat_name_index
+                        not in indexed_repeat_relative_path_args_index
+                    )
 
             return False
 
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -912,7 +912,7 @@ class Survey(Section):
             current_matchobj = matchobj
 
             if not last_saved and context:
-                if context["type"] == "text":
+                if context["type"] in ["text", "integer"]:
 
                     if not is_indexed_repeat:
                         return True
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -912,7 +912,7 @@ class Survey(Section):
             current_matchobj = matchobj
 
             if not last_saved and context:
-                if context["type"] in ["text", "integer"]:
+                if context["type"] != "calculate":
 
                     if not is_indexed_repeat:
                         return True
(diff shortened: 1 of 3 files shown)


lognaturel:
This doesn't look right. It will keep failing for every type other than text or integer which is not what we want. I think the difference in behavior is between calculates and non-calculates so perhaps that condition should be `if context["type"] != "calculate"`. But I think there's also something wrong with the else branch because it doesn't account for typed calculates. Should it maybe just be `return not("indexed-repeat" in context["bind"]["calculate"])`

gushil:
@lognaturel 
I've made the changes you suggested and all the tests are passed.

Thanks!

pyxform/survey.py
```diff
@@ -955,10 +955,7 @@ def _is_return_relative_path():
                             not in indexed_repeat_relative_path_args_index
                         )
                 else:
```
lognaturel, line 957: Seeing this again, I don't understand the two branches. I think this branch will now make every path in a calculate absolute if there's an `indexed-repeat` anywhere in the expression. All tests pass if the entire `if`/`else` structure is removed. So either there needs to be tests to support that branching or it should be removed.

gushil: Ok. I will ask for more test scenarios.
Meanwhile, let's remove the branch for now.

gushil:
@lognaturel 

Paul told me a couple of indexed-repeat expressions that could cause an issue, like indexed-repeat inside an if-else, or indexed-repeat as an expression, or indexed-repeat as an argument in other function like concat(), substr().

I think those cases are already tested, right?

In the previous commit I've already removed the branching statement, please review it.

Thanks!

gushil:
@lognaturel 

Still waiting for your review/resolve.

Thanks!

lognaturel:
Happy New Year and I'll try to take a look by the end of the week. Please see #507 for a new test to add.

lognaturel (approved):

MartijnR:
Thanks @lognaturel and @gushil!

PR #511, opened 2021-01-08
Change error message if section name is equal to form name

### LINKED ISSUE #510: Proposal to change duplicate section name error message if equal to root node name
pyxform v1.3.3.

#### Problem description
When naming a group "data", the error message can be confusing.

type | name | label
-- | -- | --
begin group | data | A group
text | a | Enter text
end group |   |  

The pyxform error message is "There are two sections with the name data."

Users would generally not understand what the other "data" section is (which is the root node).

#### Expected behavior

It may to helpful to add the following generic check to that section of code:

if duplicate section name equals the name of the root node, use error message:
    "The name "[NAME]" is the same as the form name. Use a different section name (or change the form name in the "name" column of the settings sheet)."

pinging @pbowen-oc

gushil:
Closes #510  
The simplest one.
None.
No.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -186,20 +186,19 @@ class Survey(Section):
         root_node_name = self.name
         section_names = []
         for element in self.iter_descendants():
-            if not isinstance(element, Survey):
-                if element.name == root_node_name:
+            if not isinstance(element, Survey) and element.name == root_node_name:
+                raise PyXFormError(
+                    'The name "%s" is the same as the form name. '
+                    "Use a different section name "
+                    '(or change the form name in the "name" column of the settings sheet).'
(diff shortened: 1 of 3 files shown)


gushil:
It looks like the changes made the old tests (the tests folder, not tests_v1) failed. Will look into it more.

MartijnR:
Thanks! This looks good to me after testing some forms. I think the reported reduction in code coverage is not an actual code coverage reduction (since all newly added code seems to be covered in the added test).

lognaturel (approved):

lognaturel:
> is not an actual code coverage reduction

I think it's just reported as lines covered and the error message is split over three lines with only one tested...

PR #512, opened 2021-01-14
Release v1.3.4

MartijnR:
I noticed there are 10 failing ODK Validate tests (step 2 of release steps), but that was the case in the previous version as well, and I'm guessing may have been the case for much longer. Are we just ignoring this?


lognaturel:
Yes, I've been not very good and just let those slide. I believe all of them just need slight reworks of the tests (e.g. using a valid function name, not having cycles, indexing into all repeats).

yanokwa (approved):

MartijnR:
Thank you!

PR #516, opened 2021-02-10
Generate correct refrence when selecting choices from current repeat question

### LINKED ISSUE #503: Wrong reference when building select one from repeat inside that repeat
pyxform v1.3.3

#### Problem description
When trying to use a select one from repeat inside the current repeat, the wrong reference is produced. It should be `../../repeat` but instead is `../`.

type | name | label | choice_filter | appearance | relevance | calculation
-- | -- | -- | -- | -- | -- | --
begin_repeat | pet | Pet |   | field-list |   |  
calculate | pos |   |   |   |   | position(..)
select_one ${animal_type} | animal_select | Select the animal type. If not in the list, type below. | position() != current()/../pos and animal_type != '' |   |   |  
text | animal_type | Animal type |   |   | ${animal_select} = '' |  
end_repeat | pet |   |   |   |   |  

[XLSForm](https://docs.google.com/spreadsheets/d/17r4UVIKa1GjsIXHmbpQ5_GEOiUCeX6czH09SEanjjlA/edit#gid=0)

```
    def test_choice_from_previous_repeat_in_current_repeat__parents_out_to_repeat(self):
        xlsform_md = """
        | survey  |                           |                           |                          |
        |         | type                      | name                      | label                    |
        |         | begin repeat              | pet                       | Pet                      |
        |         | select_one ${animal_type} | animal_select             | Select the animal type   |
        |         | text                      | animal_type               | Specify the animal type  |
        |         | end repeat                | pet                       |                          |
        """
        self.assertPyxformXform(
            name="data",
            debug=True,
            md=xlsform_md,
            xml__contains=[
                '<itemset nodeset="../../pet">',
            ],
        )
```

DavisRayM:
Closes #503 

Simplest solution with minimal changes

N/A

None

### DIFF
--- a/pyxform/tests_v1/test_repeat.py
+++ b/pyxform/tests_v1/test_repeat.py
@@ -472,7 +472,7 @@ class TestRepeat(PyxformTestCase):
             id_string="some-id",
             md=xlsform_md,
             xml__contains=[
-                '<itemset nodeset="../../household_mem_rep[ ./age  &gt;  current()/../../selected/target_min_age ]">',
+                '<itemset nodeset="../../household_mem_rep[ ./age  &gt;  current()/../target_min_age ]">',
             ],
         )
 
--- a/pyxform/tests_v1/test_repeat.py
+++ b/pyxform/tests_v1/test_repeat.py
@@ -472,7 +472,7 @@ class TestRepeat(PyxformTestCase):
             id_string="some-id",
             md=xlsform_md,
             xml__contains=[
-                '<itemset nodeset="../../household_mem_rep[ ./age  &gt;  current()/../target_min_age ]">',
+                '<itemset nodeset="../../household_mem_rep[ ./age  &gt;  current()/../../selected/target_min_age ]">',
             ],
         )
 
--- a/pyxform/tests_v1/test_repeat.py
+++ b/pyxform/tests_v1/test_repeat.py
@@ -653,3 +653,22 @@ class TestRepeat(PyxformTestCase):
                 """<bind nodeset="/data/family/person/prev_name" relevant=" ../age  &gt; indexed-repeat( /data/family/person/age ,  /data/family , 1,  /data/family/person , 2)" type="string"/>"""  # noqa pylint: disable=line-too-long
             ],
         )
+
+    def test_choice_from_previous_repeat_in_current_repeat_parents_out_to_repeat(self):
+        """Test choice from previous repeat in current repeat produces the correct reference"""
+        xlsform_md = """
+        | survey       |                           |               |                        |                                                      |            |                       |              |
+        |              | type                      | name          | label                  | choice_filter                                        | appearance | relevant              | calculation  |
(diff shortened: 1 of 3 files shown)


lognaturel:
From Slack:
"Especially this change [here](https://github.com/XLSForm/pyxform/pull/516/files#diff-9384652576208ba3535a94e2ff894150086f531f3a55657b2ef6af9cb223dc8fR475). I'm unsure of whether the path change is referencing something different... Would a change like that mean that it won't pick the value from the current repeat ?"

Is that what makes it a WIP or is there anything else that you expect to do here, @DavisRayM?

DavisRayM:
> From Slack:
> 
> "Especially this change [here](https://github.com/XLSForm/pyxform/pull/516/files#diff-9384652576208ba3535a94e2ff894150086f531f3a55657b2ef6af9cb223dc8fR475). I'm unsure of whether the path change is referencing something different... Would a change like that mean that it won't pick the value from the current repeat ?"
> 
> 
> 
> Is that what makes it a WIP or is there anything else that you expect to do here, @DavisRayM?

Yes, that's the only bit of that I wasn't too sure about

pyxform/tests_v1/test_repeat.py
```diff
@@ -472,7 +472,7 @@ def test_choice_from_previous_repeat_answers_in_nested_repeat_uses_current(self)
             id_string="some-id",
             md=xlsform_md,
             xml__contains=[
-                '<itemset nodeset="../../household_mem_rep[ ./age  &gt;  current()/../target_min_age ]">',
+                '<itemset nodeset="../../household_mem_rep[ ./age  &gt;  current()/../../selected/target_min_age ]">',
```
lognaturel, line 475: Your intuition is correct that this shouldn't change. It's important to stay inside the current repeat instance. When you go out of it and then back in (`../selected/`), you're going to end up in the first instance of the repeat. This will happen to have the desired behavior in Collect but it's not strictly supposed to and it won't work in Enketo.

lognaturel:
The general case is "if you need to reference a repeat nodeset, your expression must include the repeat node explicitly". In the case of itemsets from repeats, there's always a predicate expression after a nodeset expression. There are cases like parameters to the `count` function where we need to access the raw nodeset. The problem with a relative expression such as `../[predicate]` in the form in the issue is that `../` references a single repeat instance, not the whole nodeset.

In the case of `${target_min_age}`, we're not referencing the nodeset, we're referencing a specific value in a specific repeat instance.

DavisRayM:
@lognaturel Kindly re-review when you get a chance

pyxform/tests_v1/test_repeat.py
DavisRayM: I've reverted this change in the latest commits and ensured the code passes the test

lognaturel (approved):
Thanks so much, @DavisRayM!

lognaturel (approved):

PR #518, opened 2021-02-17
Options contains ${node} references and choice filter and no translation should generate itext fields.

### LINKED ISSUE #515: Choice_filter with options containing outputs does not expand ${node} references
pyxform v1.3.4

#### Problem description

Using a choice list with labels containing ${node} refs (XForm `<output>`s), does not expand the ${node} refs, if the form contains only 1 language.

Note: ODK Validate does not validate the result, so test with `--skip_validate`.

choices:

list_name | name | label
-- | -- | --
choices | one | One - ${txt}
choices | two | Two - ${txt}

survey:

type | name | label | choice_filter | default
-- | -- | -- | -- | --
text | txt | Enter text |   | default
select_one choices | one | Select one | 1 < 2 |  

The output contains:

```xml
<instance id="choices">
    <root>
        <item>
            <label>1 - ${txt}</label>
            <name>one</name>
        </item>
        <item>
            <label>2 - ${txt}</label>
            <name>two</name>
        </item>
    </root>
</instance>
```

It contains no `<itext>` block.

and: 

```xml
<itemset nodeset="instance('choices')/root/item[1 &lt; 2]">
    <value ref="name" />
    <label ref="label" />
</itemset>
```

#### Expected behavior

The output should contain (I think):

```xml
<itext>
    <translation>
        <text id="choices-0">
            <value>
                One -
                <output value=" /data/txt " />
            </value>
        </text>
        <text id="choices-1">
            <value>
                Two -
                <output value=" /data/txt " />
            </value>
        </text>
    </translation>
</itext>
...
<instance id="choices">
    <root>
        <item>
            <itextId>choices-0</itextId>
            <name>one</name>
        </item>
        <item>
            <itextId>choices-1</itextId>
            <name>two</name>
        </item>
    </root>
</instance>
```

and (see `itext()` in label `ref`):

```xml
<itemset nodeset="instance('choices')/root/item[1 &lt; 2]">
     <value ref="name" />
     <label ref="jr:itext(itextId)" />
</itemset>
```

#### Other information 

When changing the same form to use multiple languages (forcing itext usage), the problem does not occur. See [here](https://docs.google.com/spreadsheets/d/1JClWkcDhDTBekqjEXFZTq9UA0WBkL6z-UPdl4zf4uXA/edit?usp=sharing)

gushil:
Closes #515
 
The simplest one that works.

None. All tests are passed.

No.

### DIFF
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -8,7 +8,13 @@ import re
 from pyxform.errors import PyXFormError
 from pyxform.question_type_dictionary import QUESTION_TYPE_DICT
 from pyxform.survey_element import SurveyElement
-from pyxform.utils import basestring, node, unicode, default_is_dynamic
+from pyxform.utils import (
+    basestring,
+    node,
+    unicode,
+    default_is_dynamic,
+    BRACKETED_TAG_REGEX,
+)
 
 
 class Question(SurveyElement):
@@ -205,14 +211,32 @@ class MultipleChoiceQuestion(Question):
             has_media = False
             is_previous_question = bool(re.match(r"^\${.*}$", self.get("itemset")))
 
+            choice_filter_is_not_empty = choice_filter != ""
+            has_bracketed_tag_in_choice_label = False
             if choices.get(itemset):
                 has_media = bool(choices[itemset][0].get("media"))
+                for choice in choices[itemset]:
+                    if "label" in choice:
+                        choice_label = choice.get("label")
+                        if (
+                            choice_label
+                            and isinstance(choice_label, basestring)
+                            and re.search(BRACKETED_TAG_REGEX, choice_label) is not None
+                        ):
+                            has_bracketed_tag_in_choice_label = True
+            has_choice_filter_and_condition = (
+                choice_filter_is_not_empty and has_bracketed_tag_in_choice_label
+            )
 
             if file_extension in [".csv", ".xml"]:
                 itemset = itemset
                 itemset_label_ref = "label"
             else:
-                if not multi_language and not has_media:
+                if (
+                    not multi_language
+                    and not has_media
+                    and not has_choice_filter_and_condition
+                ):
                     itemset = self["itemset"]
                     itemset_label_ref = "label"
                 else:
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -257,7 +257,9 @@ class Survey(Section):
         return self.setvalues_by_triggering_ref.get("${%s}" % question_name)
 
     @staticmethod
-    def _generate_static_instances(list_name, choice_list):
+    def _generate_static_instances(
+        list_name, choice_list, has_choice_filter_with_condition
+    ):
         """
         Generates <instance> elements for static data
         (e.g. choices for select type questions)
@@ -274,19 +276,23 @@ class Survey(Section):
             choice_element_list = []
             # Add a unique id to the choice element in case there is itext
             # it references
-            if multi_language or has_media:
+            if multi_language or has_media or has_choice_filter_with_condition:
                 itext_id = "-".join([list_name, str(idx)])
                 choice_element_list.append(node("itextId", itext_id))
 
             for name, value in sorted(choice.items()):
-                if isinstance(value, basestring) and name != "label":
-                    choice_element_list.append(node(name, unicode(value)))
-                if (
-                    not multi_language
-                    and isinstance(value, basestring)
-                    and name == "label"
-                ):
-                    choice_element_list.append(node(name, unicode(value)))
+                if has_choice_filter_with_condition:
+                    if isinstance(value, basestring) and name == "name":
+                        choice_element_list.append(node(name, unicode(value)))
+                else:
+                    if isinstance(value, basestring) and name != "label":
+                        choice_element_list.append(node(name, unicode(value)))
+                    if (
+                        not multi_language
+                        and isinstance(value, basestring)
+                        and name == "label"
+                    ):
+                        choice_element_list.append(node(name, unicode(value)))
 
             instance_element_list.append(node("item", *choice_element_list))
 
@@ -447,6 +453,20 @@ class Survey(Section):
             instance=node("instance", id=name, src=uri),
         )
 
+    @staticmethod
+    def _has_bracketed_tag_in_choice_label(choices):
+        has_bracketed_tag_in_choice_label = False
+        for list_name, choice_list in choices.items():
+            for idx, choice in zip(range(len(choice_list)), choice_list):
+                for name, choice_value in choice.items():
+                    if (
+                        name == "label"
+                        and isinstance(choice_value, basestring)
+                        and re.search(BRACKETED_TAG_REGEX, choice_value) is not None
+                    ):
+                        has_bracketed_tag_in_choice_label = True
+        return has_bracketed_tag_in_choice_label
+
     def _generate_instances(self):
         """
         Get instances from all the different ways that they may be generated.
@@ -484,6 +504,7 @@ class Survey(Section):
         """
         instances = []
         generate_last_saved = False
+        choice_filter_is_not_empty = False
         for i in self.iter_descendants():
             i_ext = self._generate_external_instances(element=i)
             i_pull = self._generate_pulldata_instances(element=i)
@@ -493,14 +514,25 @@ class Survey(Section):
             for x in [i_ext, i_pull, i_file]:
                 if x is not None:
                     instances += x if isinstance(x, list) else [x]
+            parent = i.get("parent")
+            if parent and parent.get("choice_filter") != "":
+                choice_filter_is_not_empty = True
 
         if generate_last_saved:
             instances += [self._get_last_saved_instance()]
 
         # Append last so the choice instance is excluded on a name clash.
+        has_bracketed_tag_in_choice_label = self._has_bracketed_tag_in_choice_label(
+            self.choices
+        )
         for name, value in self.choices.items():
             instances += [
-                self._generate_static_instances(list_name=name, choice_list=value)
+                self._generate_static_instances(
+                    list_name=name,
+                    choice_list=value,
+                    has_choice_filter_with_condition=choice_filter_is_not_empty
+                    and has_bracketed_tag_in_choice_label,
+                )
             ]
 
         # Check that external instances have unique names.
@@ -632,11 +664,14 @@ class Survey(Section):
                         )
 
         self._translations = defaultdict(dict)  # pylint: disable=W0201
+        choice_filter_is_not_empty = False
         for element in self.iter_descendants():
             # Skip creation of translations for choices in filtered selects
             # The creation of these translations is done futher below in this
             # function
             parent = element.get("parent")
+            if parent and parent.get("choice_filter") != "":
+                choice_filter_is_not_empty = True
             if parent and not parent.get("choice_filter"):
                 for d in element.get_translations(self.default_language):
                     translation_path = d["path"]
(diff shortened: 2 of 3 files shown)


lognaturel:
The general approach here is to iterate through all the labels for a list that has a choice filter and if any of them has a reference, generate `itext` for that list. This adds yet more special cases and I think will have a noticeable performance impact on forms with really long choice lists that don't use any references in labels. It's not unusual to have thousands of choices for filtered lists and the typical case is not to have references.

Here are a couple more approaches to consider:
- Only check the first item in a list with a choice filter and generate `itext` only if that first choice has a label with a ref. That doesn't slow down the common case and I have a hard time imagining a case where someone would want a mix of labels with references and labels without. We still end up needing some special cases but with comments we can mitigate some of that.
- Always generate `itext` for lists with filters. The big advantage is that it's simpler. I don't love that it will often unecessarily increase the size of the XML and I know Enketo has trouble with forms with `itext` so this is probably not a good option but maybe I'm missing something.

@MartijnR what do you think?

pbowen-oc:
@lognaturel - I do not have a specific real-world example for your first bullet point, but I could definitely see codes like the following:
name label filter
0 "Event not related to any medication" 0
1 "Event related to ${medication1}" 1
2 "Event related to ${medication2}" 2
3 "Event related to ${medication3}" 3

If that approach were taken, I suppose a workaround in this case would be to add an item reference to a null calculated item to the first choice. I think that could work from a form design perspective, but would not be ideal.

MartijnR:
Thanks @gushil and @lognaturel. 

Good point about the potential performance issue with thousands of items (and generally not using outputs, especially not in those long lists). 

Of your suggestions, my preference would be to use a trick such as you proposed in your first approach and keep the XML output clean for the common use cases without outputs. @pbowen-oc's concern that the first option may be the odd one that doesn't use an output, is a good point. So how about we check the first 2 to be 99.9% sure (or first 10 to be 99.999999% sure)? 

I'm guessing a non-global regex test of the unparsed choices **per list** is not an option, right?

lognaturel:
> we check the first 2 to be 99.9% sure

I'm happy with that. I'd also be happy documenting that for a choice list labels to be treated as dynamic, the first item's label must be dynamic. I think @pbowen-oc's example is realistic but it seems the non-dynamic label could also go last.

> non-global regex test of the unparsed choices per list

I'm not sure I know what that means! Do you mean doing a regex on the text of all the choice labels together? I don't think there's a performant way to do that.

MartijnR:
> I'm not sure I know what that means! 

I think maybe it makes more sense in Dutch? 😂 But you actually parsed that correctly. Let's forget about that idea.

lognaturel:
> I think maybe it makes more sense in Dutch? 😂

😄😄 I thought "non-global regex" might be some technical concept I didn't know about. I think it'd work if we had all choices as text but coming out of a spreadsheet they're in a collection.

gushil:
Hi @lognaturel and @MartijnR 

Thanks for the very useful comments. I've made some changes to accommodate that. Hoping to get some review again.

Thanks!

pyxform/survey.py
```diff
@@ -447,6 +460,27 @@ def _get_last_saved_instance():
...
+                is not None
+            )
+            if len(choice_list) > 1:
+                bracketed_tag_in_choice_label = bracketed_tag_in_choice_label and (
+                    choice_list[1].get("label")
+                    and isinstance(choice_list[1].get("label"), basestring)
+                    and re.search(BRACKETED_TAG_REGEX, choice_list[1].get("label"))
+                    is not None
+                )
+        return bracketed_tag_in_choice_label
```
MartijnR, line 482: I think this algorithm is not quite right. If **any** of the first two choices contains a bracketed tag, it should generate outputs. Will propose a tweak to the test below, to catch this.

pyxform/tests_v1/test_inline_translations.py
```diff
@@ -146,3 +146,39 @@ def test_select_with_choice_filter_and_translations_generates_single_translation
...
+        Selects with options contains output and choice filter and no translations should generate itext fields for choices.
+        """
+        xform_md = """
+        | survey |                    |      |            |               |         |
+        |        | type               | name | label      | choice_filter | default |
+        |        | text               | txt  | Enter text |               | default |
+        |        | select_one choices | one  | Select one | 1 < 2         |         |
+        | choices |
+        |         | list_name | name | label        |
+        |         | choices   | one  | One - ${txt} |
```
MartijnR, line 163: Change this label to "One" without a ${txt} reference to catch a bug in the current PR.

MartijnR (changes requested):
I did a partial review, mostly by performing some transformations, and a quick glance at the code. I found one issue.

pyxform/survey.py
gushil: Thanks for the review. Fixed in the next commit.

pyxform/tests_v1/test_inline_translations.py
gushil: Added test to include this case in the next commit.

pyxform/tests_v1/test_inline_translations.py
```diff
@@ -146,3 +146,75 @@ def test_select_with_choice_filter_and_translations_generates_single_translation
...
+            ],
+            model__contains=[
+                "<itextId>choices-0</itextId>",
+                "<name>one</name>",
+                "<itextId>choices-1</itextId>",
+                "<name>two</name>",
+            ],
+            xml__contains=['<label ref="jr:itext(itextId)"/>'],
+            xml__excludes=['<label ref="label"/>'],
+        )
```
MartijnR, line 220: Now, I'm seeing this one fail (accidentally, when randomly testing):

list_name | name | label
-- | -- | --
choices | one | One - {txt}
choices | two | Two
yes_no | yes | Yes  - ${txt}
yes_no | no | No

type | name | label | choice_filter | default
-- | -- | -- | -- | --
text | txt | Enter text |   | default
select_one choices | one | Select one | 1 < 2 |  
select_one yes_no | yn | Yes or no? | 1 < 2 |  

It works for the `yes_no` list but not for the `choices` list.

MartijnR (changes requested):

pyxform/tests_v1/test_inline_translations.py
gushil: @MartijnR I see that the label for choices one is 'One - {txt}' not 'One - ${txt}'. Is that intentional?

MartijnR: No bug. Sorry about that! That explains why I couldn't figure out the cause from the code.

MartijnR (commented):
I think this is ready for a real review!

MartijnR (approved):

pyxform/question.py
```diff
@@ -205,14 +211,32 @@ def build_xml(self):
             has_media = False
             is_previous_question = bool(re.match(r"^\${.*}$", self.get("itemset")))
 
+            choice_filter_is_not_empty = choice_filter != ""
+            has_bracketed_tag_in_choice_label = False
             if choices.get(itemset):
                 has_media = bool(choices[itemset][0].get("media"))
+                for choice in choices[itemset]:
```
lognaturel, line 218: This iterates over all choices.

pyxform/survey.py
```diff
@@ -270,23 +272,34 @@ def _generate_static_instances(list_name, choice_list):
...
             for name, value in sorted(choice.items()):
-                if isinstance(value, basestring) and name != "label":
-                    choice_element_list.append(node(name, unicode(value)))
-                if (
-                    not multi_language
-                    and isinstance(value, basestring)
-                    and name == "label"
-                ):
-                    choice_element_list.append(node(name, unicode(value)))
+                if has_choice_filter_and_referenced_choice_label:
```
lognaturel, line 291: I don't understand what these changes are for.

pyxform/survey.py
```diff
@@ -632,50 +684,68 @@ def _setup_choice_translations(name, choice_value, itext_id):
...
-                    self._translations[d["lang"]][
-                        translation_path
-                    ] = self._translations[d["lang"]].get(translation_path, {})
-
-                    self._translations[d["lang"]][translation_path].update(
-                        {
-                            form: {
-                                "text": d["text"],
-                                "output_context": d["output_context"],
+            if parent:
```
lognaturel, line 693: It's not clear to me what these changes are for.

lognaturel (commented):
I think there are some leftovers from previous solutions. I propose #521 as a simplification.

pyxform/question.py
gushil: I see. I try not to change original code. Yeah, this is iterates over all choices.

pyxform/survey.py
gushil: That is for generating itextId with just name if the choices have a choice_filter applied and referenced label in it.

pyxform/survey.py
gushil: That is from the original code. It will check whether current element we are checking have a parent. From what I saw, this applied to survey element and choices, and not element in settings.

pyxform/survey.py
gushil: I saw that PR #521 still generating instance with itextId, label, name not itextId and name.

pyxform/survey.py
gushil: I use this one to check if the current element has choice_filter in it.

pyxform/survey.py
lognaturel: Ahh, great, that makes sense. Thanks for spotting that. Goes to show how important it is to have test coverage for requirements like that. [`5362a8f` (#521)](https://github.com/XLSForm/pyxform/pull/521/commits/5362a8f25749df9f6692eb78a94c96c279a9772d) should address it and add test coverage.

I think your approach here would drop additional data columns for lists with dynamic labels. I added a test for that case at https://github.com/XLSForm/pyxform/pull/521/commits/5362a8f25749df9f6692eb78a94c96c279a9772d#diff-c18988116c2ca8b0df3613f683c25a7f7eeb871dc823c2697be5db1d3f38a9acR213.

pyxform/survey.py
lognaturel: Can you think of a test that would fail without this change?

pyxform/question.py
lognaturel: I think https://github.com/XLSForm/pyxform/pull/521/files#diff-70c36a8d1a06302bbc77d3189fb1a736d02802a2c789f7687cc9e3458628bd1dR218 achieves maintaining existing code and avoiding the iteration.

pyxform/survey.py
```diff
@@ -447,6 +460,28 @@ def _get_last_saved_instance():
...
+                    choice = choice_list[choice_idx]
+                    choice_label = choice.get("label")
+                    if (
+                        choice_label
+                        and isinstance(choice_label, basestring)
+                        and re.search(BRACKETED_TAG_REGEX, choice_label) is not None
+                    ):
+                        bracketed_tag_in_choice_label = True
+                        break
+                except IndexError:
```
lognaturel, line 481: See https://github.com/XLSForm/pyxform/pull/521/files#diff-633c22eb8bd15e90567dcd0d93daf7e6ad79a27b5d55e827f7335b6178806374R279 for a way to capture the semantics of what we're trying to do without risking an index error.

gushil: That line is perfect 👍 
I learned a lot. Thanks.

pyxform/survey.py
gushil: Hmm, previously I'm not changing this line and I got a couple of test errors. But I check your PR that doing exactly that (not changing anything) and all the tests passed. So I'm assuming all the cases are covered?

pyxform/survey.py
gushil: So, it looks like we don't have to check choice_filter right?

gushil: Ah, so that's why I don't get remaining columns.

gushil: So, it looks like we don't really need to check choice_filter column at all for this case?

pyxform/question.py
gushil: Yes, I agree with that.

pyxform/survey.py
lognaturel: I can't think of a reason why we would. This code appears in `_generate_static_instances` which is only called for select choice lists that were already determined to need an `instance`. That decision is made earlier based on whether there's a choice filter.

I added a test supporting this reasoning at https://github.com/XLSForm/pyxform/pull/521/files#diff-48adf053262e50c0e674ecc7abaf998386e3ba08ff96c4b4327a84b423a307a0R231. I also verified that the test passes on `master` so that behavior hasn't changed.

pyxform/survey.py
lognaturel: All the same tests definitely pass so it's a question of whether we forgot any cases! My guess is that some of this is related to the conversation above and thinking we need to differentiate between selects that need secondary instances and those that don't.

lognaturel:
Let's close this in favor of #521 and apply what we've learned there.

PR #519, opened 2021-02-17
Add support for odk:recordaudio

### LINKED ISSUE #517: Add background-audio type
As described in https://forum.getodk.org/t/form-spec-proposal-add-background-audio-recording/31889.

| type          		| name        	| parameters     |
|----------------------------|-------------	|--------------------|
| background-audio 	| my_recording 	| quality=low|

Will result in a `my_recording` node in the instance with no corresponding `body` element.

In the model (the action doesn't have to be close to the corresponding bind; wherever actions usually land is fine):
```
<bind nodeset="/data/audio_recording" type="binary"/>
<odk:recordaudio event="odk-instance-load" ref="/data/my_recording" odk:quality="low"/>
```

The `quality` parameter is optional. Valid values are `low`, `normal`, `voice-only`. Unlike with foreground recording, `external` should result in an error.

seadowg:
Closes #517

I'm not that familiar with the code base so would not be surprised if there's a better way of parsing the parameters than what I ended up with. Mostly just put this in following how other actions/questions were handled.

Shouldn't be any really as most code refers to a new question type.


pyxform/tests_v1/pyxform_test_case.py
```diff
@@ -317,7 +317,8 @@ def assertContains(self, content, text, count=None, msg_prefix=""):
             )
         else:
             self.assertTrue(
-                real_count != 0, msg_prefix + "Couldn't find %s in content" % text_repr
+                real_count != 0,
+                msg_prefix + "Couldn't find %s in content:\n" % text_repr + content,
```
seadowg, line 321: I found I was really unclear on what exactly I was doing wrong in my implementation and adding the content here really helped me.

lognaturel: If I understand this correctly, I think what I usually do for this is turn on `debug=True`. I find that's a good sanity check anyway and usually have it on when adding a new test.

Would this make more sense as `Couldn't find %s in content %s:\n" % text_repr, content` or something? I should actually try it...

lognaturel: Ok, I see, the format is actually what I'd expect. But I think it is redundant with `debug=True`.

pyxform/question_type_dictionary.py
```diff
@@ -374,4 +374,8 @@ def generate_new_dict():
         "bind": {"type": "geopoint"},
         "action": {"name": "odk:setgeopoint", "event": "odk-instance-first-load"},
     },
+    "background-audio": {
+        "control": {"tag": "action"},
```
lognaturel, line 378: I think you'l also need a `"bind": {"type": "binary"}` here.

pyxform/tests_v1/test_background_audio.py
```diff
@@ -0,0 +1,93 @@
...
+    def test_background_audio(self):
+        self.assertPyxformXform(
+            name="data",
+            md="""
+            | survey |                  |              |
+            |        | type             | name         |
+            |        | background-audio | my_recording |
+            """,
+            xml__contains=[
+                '<odk:recordaudio event="odk-instance-load" ref="/data/my_recording"/>',
```
lognaturel, line 16: Should also test for `<bind ref="/data/my_recording" type="binary"/>`

pyxform/tests_v1/test_background_audio.py
```diff
@@ -0,0 +1,93 @@
...
+            md="""
+            | survey |                  |              |                      |
+            |        | type             | name         | parameters           |
+            |        | background-audio | my_recording | quality=foo |
+            """,
+            errored=True,
+            error__contains=["Invalid value for quality."],
+        )
+
+    @unittest.skip("Required update to Validate to work")
```
lognaturel, line 83: We won't want this as part of the test suite because running Validate is time-consuming. Instead we run all tests with Validate as part of review and as a sanity check before release. Should probably automate that some day but it's not a big deal for now.

lognaturel (changes requested):
Looks good, just a few things!

pyxform/tests_v1/test_background_audio.py
seadowg: Ah actually yeah I was meaning to ask about that. We don't include anything about `bind` in the final example in the [spec discussion](https://forum.getodk.org/t/form-spec-proposal-add-background-audio-recording/31889/21) and we also haven't included it in [tests for Collect](https://github.com/getodk/collect/blob/master/collect_app/src/androidTest/assets/forms/one-question-background-audio.xml). Do we just need it for correctness? Hoping there's nothing wrong with the Collect implementation 🤞.

lognaturel: It's important for tools that might treat answers of different types differently, usually servers. Form filling clients probably don't care. The default bind type is `string` and that's fine for a file names.

seadowg: Ah right but the server needs to know it's `binary`. I see!

pyxform/tests_v1/pyxform_test_case.py
seadowg: It seems strange to me to only output the content optionally. Wouldn't you always want to know if it fails what the resulting XForm looked like? Also, as much as it does the right thing it's a bit of a secret that you can use the `debug` parameter here.

pyxform/tests_v1/test_background_audio.py
lognaturel: Yeah, the way we've written the XForms spec, the action should only have any effect if the `ref` is of type `binary`. We're more lenient than that at the moment in JR/Collect. I think that's ok if manual XML writers really want to do crazy things but it's important that form builders provide all the info.

lognaturel (approved):

pyxform/tests_v1/pyxform_test_case.py
lognaturel: 🤷🏻‍♀️ Fair enough!

pyxform/tests_v1/test_background_audio.py
lognaturel: ("all the info" = "this node actually represents a binary file so treat its value as a filename, not just a string")

lognaturel:
@MartijnR it occurs to me you might like (and it might be a good idea) to do a sanity check on this.

MartijnR:
Thanks. I had a quick look at the tests. All looks good to me!

PR #521, opened 2021-02-25
Use itext for dynamic selects with dynamic labels

### LINKED ISSUE #515: Choice_filter with options containing outputs does not expand ${node} references

lognaturel:
Closes #515

This is a simplification of #518. I aimed to have a single function that identifies the dynamic label case and to make sure that the checks on conditions for using `itext` are easily readable. I considered seeing if we could avoid the redundancy between those checks but that felt too involved for the moment. 

I don't think there's a risk of degraded user experience. I think the worst case would be that we might be generating `itext` too often now but that would generally be safe.

Another risk is related to behavior I changed for consistency that affects selects from itemsets with media. Previously, both `itext` and `label` children were generated in the secondary instance. This is unnecessary but I can imagine some downstream analysis tools or servers relying on it. I think it's clearly better to omit so I think it's ok.

It would be good to explicitly mention that we can use references for labels but not for names or other columns. I've filed https://github.com/getodk/docs/issues/1329. I'm not sure it fits in to xlsform.org.

### DIFF
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -8,7 +8,14 @@ import re
 from pyxform.errors import PyXFormError
 from pyxform.question_type_dictionary import QUESTION_TYPE_DICT
 from pyxform.survey_element import SurveyElement
-from pyxform.utils import basestring, node, unicode, default_is_dynamic
+from pyxform.utils import (
+    basestring,
+    node,
+    unicode,
+    default_is_dynamic,
+    BRACKETED_TAG_REGEX,
+    has_dynamic_label,
+)
 
 
 class Question(SurveyElement):
@@ -212,7 +219,11 @@ class MultipleChoiceQuestion(Question):
                 itemset = itemset
                 itemset_label_ref = "label"
             else:
-                if not multi_language and not has_media:
+                if (
+                    not multi_language
+                    and not has_media
+                    and not has_dynamic_label(choices[itemset], multi_language)
+                ):
                     itemset = self["itemset"]
                     itemset_label_ref = "label"
                 else:
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -210,20 +210,18 @@ class MultipleChoiceQuestion(Question):
             itemset_value_ref = "name"
             itemset, file_extension = os.path.splitext(self["itemset"])
             has_media = False
+            has_dyn_label = False
             is_previous_question = bool(re.match(r"^\${.*}$", self.get("itemset")))
 
             if choices.get(itemset):
                 has_media = bool(choices[itemset][0].get("media"))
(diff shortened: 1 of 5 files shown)


gushil:
@lognaturel @MartijnR 
Should we merge this instead of PR #518 ?

MartijnR:
Up to @lognaturel, but this looks like a cleaner solution. Thanks @lognaturel!

lognaturel:
Yes, please do a final review on this one, @gushil and @MartijnR.

gushil:
> Yes, please do a final review on this one, @gushil and @MartijnR.

I'm Ok. Thanks @lognaturel !

MartijnR (approved):
Great. Thanks!

PR #522, opened 2021-02-25
Add support for csv-external

### LINKED ISSUE #271: Support csv-external for defining an external instance not necessarily used for a select
https://github.com/opendatakit/xforms-spec/issues/88 introduces a new external itemsets in csv feature to the specs. This has been already implemented in JavaRosa with https://github.com/opendatakit/javarosa/pull/397 and it looks like we could implement this in PyXForm now too.

lognaturel:
Closes #271

Uses the same code paths as `xml-external` which reduces risk. I briefly considered treating them differently in some way but that seemed more complex for no payoff.

I tried to come up with the minimal tests we needed to validate the addition. Since the same code is used as with `xml-external`, I think a lot of the cases are already handled but please be on the lookout for anything that might not be covered.

I noticed some tests were configured to always run Validate. We've discussed not doing that to speed up the test suite so I removed that. We always run all tests with Validate before a release. I also added a commit to make sure all tests pass Validate because failures were mostly in the test file I was already in.

This is an additive change and I don't see a regression risk. The only existing code that has changed has to do with `xml-external` and that has really good test coverage.

https://github.com/XLSForm/xlsform.github.io/issues/216

### DIFF
--- a/pyxform/builder.py
+++ b/pyxform/builder.py
@@ -115,7 +115,7 @@ class SurveyElementBuilder(object):
             d = self._sections[section_name]
             full_survey = self.create_survey_element_from_dict(d)
             return full_survey.children
-        elif d["type"] == "xml-external":
+        elif d["type"] in ["xml-external", "csv-external"]:
             return ExternalInstance(**d)
         else:
             self._save_trigger_as_setvalue_and_remove_calculate(d)
--- a/pyxform/question_type_dictionary.py
+++ b/pyxform/question_type_dictionary.py
@@ -369,6 +369,9 @@ QUESTION_TYPE_DICT = {
     "xml-external": {
         # Only effect is to add an external instance.
     },
+    "csv-external": {
+        # Only effect is to add an external instance.
+    },
     "start-geopoint": {
         "control": {"tag": "action"},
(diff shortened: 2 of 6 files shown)


ukanga (approved):

MartijnR (changes requested):
Very exciting feature! Thanks!

We should do `jr://file-csv/name.csv` and this what we use with pulldata in pyxform as well. 

https://getodk.github.io/xforms-spec/#file-endpoints

lognaturel:
> We should do jr://file-csv/name.csv

I completely forgot about that, you're absolutely right! 😬

lognaturel:
Thanks for remembering that, @MartijnR. I'm a little shaken that https://github.com/XLSForm/pyxform/pull/522/files#diff-8ca5f58100688d382c2625d49faf9b321a0edd1f4dd98526bc8e71d0f11ab4e8R176 didn't catch that. I was thinking it would help us know the generated instance declarations would be identical. I now see there's only a check for an instance name conflict so it doesn't really cover us that much.

MartijnR (dismissed):
Thanks!

lognaturel:
I rebased on master which auto fixed merge conflicts.

PR #524, opened 2021-03-03
${} references in predicates should be expanded using current()

### LINKED ISSUE #490: ${} references in predicates not used as choice filters should be expanded using current()
All versions

#### Problem description
It's useful to manually write XPath expressions with predicates and it can be convenient to mix in `${}` references. For example, consider [this form](https://docs.google.com/spreadsheets/d/1HpDBGhe7cQYODmDG9-YkjBTzo9VdUwq1EPaZHnweNhM/edit#gid=1045645704) which looks up items in a repeat. The expression `instance('item')/root/item[itemindex=current()/../item-counter]/label` would be convenient to write as `instance('item')/root/item[itemindex=${item-counter}]/label`. However, this would currently be expanded to `instance('item')/root/item[itemindex=../item-counter]/label` which is incorrect.

Convert a form such as the one linked above. Notice that the expression in the predicate does not use `current()`.

#### Expected behavior
All `${}` expressions in XPath predicates should expand using `current()`. Alternately we could document that we don't allow using `${}` references in XPath predicates.

gushil:
Closes #490 

I use simple predicate testing by checking whether current context start with `instance(`. I'm thinking about using external library for xpath checking, but thought that would be too overkill for this case.

None, but probably still need some more tests. Open to more tests suggestion!

No.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -889,9 +889,15 @@ class Survey(Section):
         name = matchobj.group(2)
         last_saved = matchobj.group(1) is not None
         is_indexed_repeat = matchobj.string.find("indexed-repeat(") > -1
-        is_predicate = matchobj.string.startswith("instance(")
         indexed_repeat_regex = re.compile(r"indexed-repeat\([^)]+\)")
         function_args_regex = re.compile(r"\b[^()]+\((.*)\)$")
+
+        # check if current matchobj is a predicate
+        # (starts with 'instance(' and have square bracket in it)
+        is_secondary_instance = (
+            matchobj.string.startswith("instance(")
+            and re.search(r"\[(.+)\]", matchobj.string) is not None
+        )
 
         def _relative_path(name):
             """Given name in ${name}, return relative xpath to ${name}."""
@@ -976,7 +982,7 @@ class Survey(Section):
 
         if _is_return_relative_path():
             if not use_current:
-                use_current = is_predicate
+                use_current = is_secondary_instance
             relative_path = _relative_path(name)
             if relative_path:
                 return relative_path
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -889,6 +889,7 @@ class Survey(Section):
         name = matchobj.group(2)
         last_saved = matchobj.group(1) is not None
         is_indexed_repeat = matchobj.string.find("indexed-repeat(") > -1
+        is_predicate = matchobj.string.startswith("instance(")
         indexed_repeat_regex = re.compile(r"indexed-repeat\([^)]+\)")
         function_args_regex = re.compile(r"\b[^()]+\((.*)\)$")
 
@@ -974,6 +975,8 @@ class Survey(Section):
             )
 
         if _is_return_relative_path():
+            if not use_current:
+                use_current = is_predicate
             relative_path = _relative_path(name)
             if relative_path:
                 return relative_path
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -891,13 +891,19 @@ class Survey(Section):
         is_indexed_repeat = matchobj.string.find("indexed-repeat(") > -1
         indexed_repeat_regex = re.compile(r"indexed-repeat\([^)]+\)")
         function_args_regex = re.compile(r"\b[^()]+\((.*)\)$")
+        instance_regex = re.compile(r"instance\([^)]+\S+")
 
-        # check if current matchobj is a predicate
-        # (starts with 'instance(' and have square bracket in it)
-        is_secondary_instance = (
-            matchobj.string.startswith("instance(")
-            and re.search(r"\[(.+)\]", matchobj.string) is not None
-        )
+        def _is_secondary_instance():
+            """check if current matchobj is a predicate and part of primary instance"""
+            # It is possible to have multiple instance in an expression
+            instance_match_iter = instance_regex.finditer(matchobj.string)
+            for instance_match in instance_match_iter:
+                if (
+                    matchobj.start() >= instance_match.start()
+                    and matchobj.end() <= instance_match.end()
+                ):
+                    return True
+            return False
 
         def _relative_path(name):
             """Given name in ${name}, return relative xpath to ${name}."""
@@ -982,7 +988,7 @@ class Survey(Section):
 
         if _is_return_relative_path():
             if not use_current:
-                use_current = is_secondary_instance
+                use_current = _is_secondary_instance()
             relative_path = _relative_path(name)
             if relative_path:
                 return relative_path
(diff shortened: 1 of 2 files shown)


pyxform/survey.py
```diff
@@ -889,6 +889,7 @@ def _var_repl_function(
         name = matchobj.group(2)
         last_saved = matchobj.group(1) is not None
         is_indexed_repeat = matchobj.string.find("indexed-repeat(") > -1
+        is_predicate = matchobj.string.startswith("instance(")
```
MartijnR, line 892: This is quite an original solution to 

> figuring out the least hacky/most performant way to identify that we're expanding a reference in a predicate

This `is_predicate` doesn't look at predicates at all! :) However, it works. I think this is because the only way we can use a ${node} inside a path is as a predicate. So checking for square brackets may not be necessary?

If so, it might be good to change the variable to `is_secondary_instance` or something like that.

MartijnR (commented):
I've not been able to break this 3-line solution 😄 

I added some tests to the issue, that may be helpful if this code ever gets refactored.

gushil:
> I added some tests to the issue, that may be helpful if this code ever gets refactored.

Thanks. I will add those tests.

pyxform/survey.py
gushil: > This `is_predicate` doesn't look at predicates at all! :) However, it works. I think this is because the only way we can use a ${node} inside a path is as a predicate. So checking for square brackets may not be necessary (?).
> 

Ok. I will add the check for square brackets.

> It might be good to change the variable to `is_secondary_instance` or something like that.

Ok. Will change the variable name in the next commit.

Thanks for the review.

gushil:
@MartijnR 

Addressed your feedbacks in recently pushed commits.

Thanks.

pyxform/survey.py
```diff
@@ -892,6 +892,13 @@ def _var_repl_function(
         indexed_repeat_regex = re.compile(r"indexed-repeat\([^)]+\)")
         function_args_regex = re.compile(r"\b[^()]+\((.*)\)$")
 
+        # check if current matchobj is a predicate
+        # (starts with 'instance(' and have square bracket in it)
+        is_secondary_instance = (
+            matchobj.string.startswith("instance(")
+            and re.search(r"\[(.+)\]", matchobj.string) is not None
```
MartijnR, line 899: Sorry for my confusing comment. I think the square bracket (i.e. predicate) check may not be necessary (like you had initiallly), since the use of ${node} syntax implies this. But would be good to hear what @lognaturel thinks.

lognaturel (changes requested):
Apologies for the delay on this!

I agree with @MartijnR that looking ONLY for `instance(` as you had before is a clever way to identify expressions that will need `current`. As far as I can tell, the only place there could be a `${}` reference in such an expression is in a predicate.

The issue with this is that it will only work if a complete expression starts with `instance(`. Consider for example changing `test_repeat_using_select_with_reference_path_in_predicate_uses_current` to have `${item-counter} + instance('item')/root/item[itemindex=${item-counter}]/label` (it's nonsense but illustrates the point). Now `current()` isn't used anywhere.

pyxform/tests_v1/test_repeat.py
```diff
@@ -653,3 +653,118 @@ def test_indexed_repeat_math_expression_with_double_variable_in_nested_repeat_re
...
+        """
+        self.assertPyxformXform(
+            name="data",
+            md=xlsform_md,
+            xml__contains=[
+                """<bind calculate="instance('item')/root/item[index= /data/pos1 ]/label" nodeset="/data/item1" type="string"/>"""  # noqa pylint: disable=line-too-long
+            ],
+        )
+
+    def test_repeat_with_reference_path_not_in_predicate_uses_xpath(self,):
```
lognaturel, line 751: `rep2` and `item2` don't exist so it's hard to tell what this should do. I assume you're illustrating the case where one repeat indexes into another parallel repeat? That seems like a good test to have but `rep2` should exist. I think it's also more likely that a form designer would write `${rep2}[number(${pos4})]/item2`

The test name and comment should be updated to reflect what this tests. The reference path IS inside a predicate. The key is that the node with the calculation and referenced node have the same parent.

gushil: I think this test is just for testing reference path that is not part of primary instance. Fixed.

gushil:
> Apologies for the delay on this!
> 
> I agree with @MartijnR that looking ONLY for `instance(` as you had before is a clever way to identify expressions that will need `current`. As far as I can tell, the only place there could be a `${}` reference in such an expression is in a predicate.
> 
> The issue with this is that it will only work if a complete expression starts with `instance(`. Consider for example changing `test_repeat_using_select_with_reference_path_in_predicate_uses_current` to have `${item-counter} + instance('item')/root/item[itemindex=${item-counter}]/label` (it's nonsense but illustrates the point). Now `current()` isn't used anywhere.

Thanks for the review @lognaturel 

Subsequent commit includes fixes for this as well.

pbowen-oc:
@lognaturel - Is there anything @gushil or I can do to help with this issue?

lognaturel:
> Is there anything @gushil or I can do to help with this issue?

@pbowen-oc Y'all have cycles for some of my other work? ;) Taking a look...

pyxform/survey.py
```diff
@@ -891,6 +891,19 @@ def _var_repl_function(
         is_indexed_repeat = matchobj.string.find("indexed-repeat(") > -1
         indexed_repeat_regex = re.compile(r"indexed-repeat\([^)]+\)")
         function_args_regex = re.compile(r"\b[^()]+\((.*)\)$")
+        instance_regex = re.compile(r"instance\([^)]+\S+")
+
+        def _is_secondary_instance():
+            """check if current matchobj is a predicate and part of primary instance"""
```
lognaturel, line 897: I think we're really going to tie ourselves up in knots if this code needs to be revisited so making sure the naming is  precise is important.

Name: _in_secondary_instance_predicate
Comment: check if ${} expression represented by matchobj is in a predicate for a path expression for a secondary instance

@MartijnR any other ideas for tightening the wording?

lognaturel (changes requested):
Thank you, this approach makes sense to me semantically -- check to see if a ${} expression is in a predicate for a path expression that starts with `instance(`. This will slow form conversion down since every single `${}` expression will go through an extra check but I can't think of a great alternative.

I can't think of additional test cases. I would like the naming in the code changes to be a little more precise.

pyxform/survey.py
gushil: Thanks for the suggestion @lognaturel 
Fixed in the next commit.

lognaturel (approved):
Thanks, @gushil!

PR #528, opened 2021-04-05
Node-set arguments are returned as absolute path.

### LINKED ISSUE #525: node-set arguments are converted to relative paths when they should be absolute
#### Problem description
node-set arguments are converted to relative paths if the argument node(s) is/are inside a repeat and referred from the same repeat

type | name | label | calculation | readonly
-- | -- | -- | -- | --
begin repeat | repeat |   |   |  
integer | item1 | item1 |   |  
integer | count1 | count1 | count(${item1}) | yes
integer | countne1 | countne1 | count-non-empty(${item1}) | yes
integer | sum1 | sum1 | sum(${item1}) | yes
integer | max1 | max1` | max(${item1}) | yes
integer | min1 | min1 | min(${item1}) | yes
text | concat1 | concat1 | concat(${item1}) | yes
text | join1 | join1 | join('...', ${item1}) | yes
text | randomize1 | randomize1 | randomize(${item1}) | yes
text | area | area | area(${item1}) | yes
text | distance | distance | distance(${item1}) | yes
end repeat |   |   |   |  

#### Expected behavior

All `item1` should output as `/data/repeat/item1`, and not as `../item1`.

#### Other information 

This is exactly the same problem as the already fixed https://github.com/XLSForm/pyxform/issues/484 for `indexed-repeat()`, so this fix should probably be integrated with that fix.

Perhaps it would be helpful to maintain an object of XPath functions that contain node-set parameters and list the parameter indices that are node-sets in arrays, e.g.

```json
{
     "count": [0],
     "count-non-empty": [0],
     "sum":  [0],
     "max": ["*"],
     "min": ["*"],
     "concat": ["*"],
     "join": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20],
     "randomize": [0],
     "indexed-repeat": [0, 1, 3, 5],
     "area": [0],
     "distance": [0]
}
```
 
I believe `join()` can have an unlimited number of parameters, and only the first is not a node-set. So perhaps that should be handled separately, or a syntax can be invented for 1+. Or just not worry about more than 20, since that's very likely to never be used.

gushil:
Closes #525

The expansion of previous working solution.

None.

No.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -903,7 +903,7 @@ class Survey(Section):
         name = matchobj.group(2)
         last_saved = matchobj.group(1) is not None
         function_args_regex = re.compile(r"\b[^()]+\((.*)\)$")
-        is_begin_function_regex = re.compile(r"\S+\(")
+        is_begin_function_regex = re.compile(r"\S+\s*\(")
 
         xpath_functions_with_node_set_args_indices = {
             "count": [0],
@@ -973,13 +973,18 @@ class Survey(Section):
                         xpath_function
                     ]
 
-                    is_xpath_function = matchobj.string.find(xpath_function + "(") > -1
+                    begin_xpath_function_regex = re.compile(
+                        r"{}\s*\(".format(xpath_function)
+                    )
+                    is_xpath_function = (
+                        begin_xpath_function_regex.search(matchobj.string) is not None
+                    )
                     if not is_xpath_function:
                         continue
 
                     # It is possible to have multiple xpath function in an expression
                     xpath_function_regex = re.compile(
-                        r"{}\([^)]+\)".format(xpath_function)
+                        r"{}\s*\([^)]+\)".format(xpath_function)
                     )
                     xpath_function_iter = xpath_function_regex.finditer(matchobj.string)
                     for xpath_func in xpath_function_iter:
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -902,9 +902,43 @@ class Survey(Section):
 
         name = matchobj.group(2)
         last_saved = matchobj.group(1) is not None
-        is_indexed_repeat = matchobj.string.find("indexed-repeat(") > -1
-        indexed_repeat_regex = re.compile(r"indexed-repeat\([^)]+\)")
         function_args_regex = re.compile(r"\b[^()]+\((.*)\)$")
+        is_begin_function_regex = re.compile(r"\S+\(")
+
+        xpath_functions_with_node_set_args_indices = {
+            "count": [0],
+            "count-non-empty": [0],
+            "sum": [0],
+            "max": ["*"],
+            "min": ["*"],
+            "concat": ["*"],
+            "join": [
+                1,
+                2,
+                3,
+                4,
+                5,
+                6,
+                7,
+                8,
+                9,
+                10,
+                11,
+                12,
+                13,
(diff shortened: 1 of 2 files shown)


pyxform/survey.py
```diff
@@ -930,50 +964,75 @@ def _relative_path(name):
...
-                    return True
+                for xpath_function in xpath_functions_with_node_set_args_indices:
+                    xpath_function_node_set_args_indices = xpath_functions_with_node_set_args_indices[
+                        xpath_function
+                    ]
 
-                # It is possible to have multiple indexed-repeat in an expression
-                indexed_repeats_iter = indexed_repeat_regex.finditer(matchobj.string)
-                for indexed_repeat in indexed_repeats_iter:
+                    is_xpath_function = matchobj.string.find(xpath_function + "(") > -1
```
MartijnR, line 977: This fails when there are spaces before the first bracket as in `concat   (${item1}, "b")`.

pyxform/tests_v1/test_repeat.py
```diff
@@ -602,7 +602,7 @@ def test_mixed_variables_and_indexed_repeat_in_expression_text_type_nested_repea
                 |         | end repeat     |                |                                  |                                                                                                                 |
             """,  # noqa pylint: disable=line-too-long
             xml__contains=[
-                """<bind calculate="concat( ../name , indexed-repeat( /data/family/person/age ,  /data/family , 1,  /data/family/person , 2),  ../age )" nodeset="/data/family/person/prev_name" type="string"/>"""  # noqa pylint: disable=line-too-long
+                """<bind calculate="concat( /data/family/person/name , indexed-repeat( /data/family/person/age ,  /data/family , 1,  /data/family/person , 2),  ../age )" nodeset="/data/family/person/prev_name" type="string"/>"""  # noqa pylint: disable=line-too-long
```
MartijnR, line 605: This change in output in the first argument may sometimes be undesirable. Both relative and absolute path outputs have valid use cases. Not sure how we could please everybody. Just wanted to flag this for @lognaturel.

MartijnR (changes requested):
Thanks! 

Apart from by using space(s) before parens, I wasn't able to break this.

pyxform/tests_v1/test_repeat.py
MartijnR: After discussing with @pbowen-oc, it may best to take concat() out of this feature and leave its behavior as is.

@lognaturel, I would have appetite in removing our concat() customization from the spec... https://getodk.github.io/xforms-spec/#fn:concat

gushil: > After discussing with @pbowen-oc, it may best to take concat() out of this feature and leave its behavior as is.
> 
> @lognaturel, I would have appetite for removing our concat() customization from our spec actually... https://getodk.github.io/xforms-spec/#fn:concat

Ok. I will remove `concat` handling in the next commit.

pyxform/survey.py
gushil: You are right. Will try to fix this issue in the next commit(s).

lognaturel:
After conversation in #525, we've decided not to do this for now because it changes long-standing behavior. The effect is mostly on folks who use nested repeats but it may also result in performance issues in other contexts.

PR #529, opened 2021-04-08
Apply section name check to sections only

jnm:

#511 is a helpful solution to #510, providing less-confusing error text when a group name matches the root node name, but it also ensnared [(unintentionally?)](https://github.com/XLSForm/pyxform/issues/510#issuecomment-763063318) regular question names. Other approaches considered and rejected were:
1. Reverting #511, losing a valuable addition to pyxform and ignoring the problem described in #510;
1. Doing nothing, and thus requiring unnecessary changes to forms that worked with pyxform prior to v1.3.4.

Some unknown tool could depend on the recently-added check that prevents question names from equaling the form name.

No; however, if this PR is undesirable, then the documentation should be updated to reflect the new restriction on question names.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -187,7 +187,7 @@ class Survey(Section):
         root_node_name = self.name
         section_names = []
         for element in self.iter_descendants():
-            if not isinstance(element, Survey) and element.name == root_node_name:
+            if element.type != "survey" and element.name == root_node_name:
                 raise PyXFormError(
                     'The name "%s" is the same as the form name. '
                     "Use a different section name "
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -187,15 +187,17 @@ class Survey(Section):
         root_node_name = self.name
         section_names = []
         for element in self.iter_descendants():
-            if element.type != "survey" and element.name == root_node_name:
-                raise PyXFormError(
-                    'The name "%s" is the same as the form name. '
-                    "Use a different section name "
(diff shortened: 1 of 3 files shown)


lognaturel:
Indeed, thank you! I had missed @pbowen-oc's https://github.com/XLSForm/pyxform/issues/510#issuecomment-763063318 to that effect.

Could you please add a test very similar to https://github.com/XLSForm/pyxform/pull/511/files#diff-2f87b2f8ddd9e41e8ebf5bf45472a415f5d2744e146a487aa0516762c79c47cf that verifies that a field name that matches the root node name is accepted?

jnm:
Sure :) I added one just now. Should I also check the generated XML, or is it enough as-is? I started writing that but threw it away, thinking that it'd add brittleness and duplicate other tests

lognaturel (approved):
🙏 Thank you! Some people I know would be very offended by the lack of explicit assertion and I might add `errored=False` to placate them but I don't mind as-is.

Leaving open for a bit in case @gushil, @MartijnR or @pbowen-oc want to sanity check. Any urgency to release?

jnm:
> I might add `errored=False`

Aye, aye :+1: I don't want to get run out of town or anything :laughing:

MartijnR:
Fine with me. Thanks! 👍

lognaturel (approved):

pbowen-oc:
@lognaturel - We would like this update, but there is no urgency on it.

PR #538, opened 2021-06-07
quick handling of default image values

### LINKED ISSUE #405: Add default value support for annotate question
If an annotate question has a default value (a filename), the XForm should prepend this value with `jr://images/`.

So if the default column for question annotate question `a` contains `body.png`, the output in the XForm instance should be: `<a>jr://images/body.png</a>`. 

Note that this value transformation is the same as what is used for values in the `media::image::` column.

(as agreed [here](https://forum.opendatakit.org/t/proposal-specify-a-fixed-image-to-be-annotated-or-signed/14954/18) - XForms spec PR [here](https://github.com/opendatakit/xforms-spec/pull/257)

KeynesYouDigIt:
Closes #405 

This is the simplest and fastest way to get default support for images. I am creating another pull request now which creates broader support for different binary defaults.

image questions now have another processing step. Nothing in this step appears unsafe, but if there is a regression, the new step would be the first place to look. Given this is a small change, in the unlikely event of a regression, I believe the fix will be easy.

I do not believe so.

### DIFF
--- a/pyxform/aliases.py
+++ b/pyxform/aliases.py
@@ -115,7 +115,7 @@ list_header = {
     "video": "media::video",
 }
 # Note that most of the type aliasing happens in all.xls
-_type = {
+_type_alias_map = {
     "imei": "deviceid",
     "image": "photo",
     "add image prompt": "photo",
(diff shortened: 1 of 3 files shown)


ukanga (approved):
Looks good to me.

PR #539, opened 2021-06-07
[ WIP/ Needs discussion ] handle all binaries with a big constant

### LINKED ISSUE #405: Add default value support for annotate question

KeynesYouDigIt:
Closes #405 

This solution centralizes the handling of Sandbox files in a form while adding support for multiple binary default types.

Users attempting to add sandbox files that don't conform to expectations might see unexpected behavior - I have tried to avoid this but its worth discussing how common it is.

In testing, I found a few CSV and XML handling bugs, but those were cleared up. Still, I dont know if tests are missing for these and other file types being used in an unexpected way.

Overall, my question is: do we have validation on paths to binary files? if not, is now the time to add it? or could we choke out some much needed flexibility?

Only if we add stricter validation of file types, and even then only maybe.

### DIFF
--- a/pyxform/aliases.py
+++ b/pyxform/aliases.py
@@ -115,7 +115,7 @@ list_header = {
     "video": "media::video",
 }
 # Note that most of the type aliasing happens in all.xls
-_type = {
+_type_alias_map = {
     "imei": "deviceid",
     "image": "photo",
     "add image prompt": "photo",
--- a/pyxform/constants.py
+++ b/pyxform/constants.py
@@ -107,15 +107,14 @@ AUDIO_QUALITY_EXTERNAL = "external"
 # TODO - theres key repetion here to handle the diverse use cases.
 # could I be a function? or a DefaultDict?
 SANDBOXED_TYPE_EP_PREFIX_MAP = {
-    'photo': 'jr://images/',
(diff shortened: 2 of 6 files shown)


lognaturel:
Thanks so much for looking into this, @KeynesYouDigIt. It hasn't come up as a user priority recently so I'm going to close for now.

KeynesYouDigIt:
> Thanks so much for looking into this, @KeynesYouDigIt. It hasn't come up as a user priority recently so I'm going to close for now.

It was super fun to hear from you guys! I hope you all are doing well.

lognaturel:
Hi, nice to see you too! 👋  ODK continues to grow and be used for all kinds of cool, impactful projects. Thanks for your contributions over the years!

Issue #439, opened 2020-03-16
Output empty `<label/>` instead of omitting it.


MartijnR:
Currently pyxform requires either a value in the "label" column or in the "hint" column for form controls (ie. questions in the body). This validation rule can remain unchanged.

However, as @tiritea discovered,  a `<label>` element is actually required according to the XForms spec. Therefore, we should include an empty `<label/>` (or `<label></label>`) element in the body output. This is currently not the case.

### Test 1

| type | name | label | hint |
|-----|--------|-----|----|
| text | a        |         | a hint |
| text | b       |A label |        |

```xml
<input ref="/data/a">
    <label/>
    <hint>a hint</hint>
</input>
<input ref="/data/b">
    <label>A label</label>
</input>
```

lindsay-stevens:
@lognaturel / @MartijnR  it looks like the error for no hint comes from ODK Validate rather than pyxform. Should pyxform warn / error about no hint, or just emit a label to comply with the spec and ODK Validate?

It doesn't seem like an error is necessary, since adding the empty label XML node shouldn't affect how the survey works. But maybe it does for some tools?

MartijnR:
Hi @lindsay-stevens. Welcome back!

> Should pyxform warn / error about no hint, or just emit a label to comply with the spec and ODK Validate?

If ODK Validate does not get fooled by an empty label (and still requires either a label or hint text content), then I think outputting an empty label in pyxform is all that is required.

KeynesYouDigIt:
@lindsay-stevens Are you already working on this? If not I might be able to take a swing at it! let me know either way

lindsay-stevens:
@MartijnR thanks!
@KeynesYouDigIt yes I have a draft that I'll create a PR for this week.

PR #542, opened 2021-07-03
Do not warn on unlabeled field-lists

yanokwa:
Too many warnings make it hard for users to see things they should be actually warned about. In this case, users use field-lists to visually group questions and that grouping really doesn't need a label, so we don't need to warn users.

And while I was looking at the warnings, I wanted to improve them by specifying that the warning was about a group or a repeat.

Originally, I wanted to warn on every group and every repeat, but decided to dial it back because...
* Collect uses the repeat label in the UI when asking about what creating repeats
* Groups without a field-list are probably fine to remove a warning, but this current PR feels like a safer first step.

There could be some downstream tools that rely on group labels, but since has always been a warning and not an error, it seems likely that those tools can handle missing group labels.

There doesn't seem to be a fixed list of acceptable ways to say "begin group". There is a regex in https://github.com/XLSForm/pyxform/blob/master/pyxform/xls2json.py#L570 that I built my tests around.

Finally, there isn't a lot of test coverage on groups, so that didn't make me feel great.

No, infact https://xlsform.org/en/#multiple-webpage-forms shows an example of a field-list without a label.


lognaturel:
Is this ready for a look or any particular reason it's still a draft?

lognaturel:
@lindsay-stevens and I discussed and are on board. He had to make some conflicting changes in https://github.com/XLSForm/pyxform/issues/439 so this will need to be reworked some.

yanokwa:
Closing in favor of https://github.com/XLSForm/pyxform/pull/545

PR #543, opened 2021-07-05
Output empty `<label/>` instead of omitting it.

### LINKED ISSUE #439: Output empty `<label/>` instead of omitting it.

lindsay-stevens:
- as per #439
- pyxform should always output a label node even if there is no label specified.
- pyxform had a warning for no label but this is removed since there is no requirement for a label.
- added tests to verify that a label is output and no warning raised
- fixed existing tests that either expected a warning or no label

Closes #439

As discussed in the issue ticket, the XForms spec requires a label node. It would not be feasible to change the spec, so this solution brings pyxform output in to line with the spec.

It's possible that existing users expect there to be a warning if no label is supplied but a hint is supplied. So the regression risk would be in other apps that use pyxform. In this case the other app could add a check for empty labels.

The xlsform.org site says that the survey sheet label is mandatory. However, the update here is specific to pyxform behaviour. It would not be accurate to say that the label is not mandatory, because it is. It's just that for pyxform, in the absence of a label it will emit one in order to comply with the spec.


lognaturel:
From @MartijnR at https://github.com/XLSForm/pyxform/issues/439#issuecomment-867815645:
> If ODK Validate does not get fooled by an empty label (and still requires either a label or hint text content)

The PR description doesn't address this that I can see -- will Validate still throw an error if there's an empty label and no hint? That would be the desired behavior.

lognaturel:
I've come back to try this, and I do confirm that there is still a pyxform error thrown if neither label nor hint is present which is what we want. 👍

I went straight to tests to try to see if that was covered and didn't see it because the forms aren't inline.

pyxform/tests_v1/test_guidance_hint.py
```diff
@@ -73,28 +74,38 @@ def test_multi_language_guidance(self):
 
     def test_guidance_hint_only(self):
         """Test guidance_hint only."""
```
lognaturel, line 76: This is unexpected. Guidance hints are supplemental so it feels like it doesn't make sense to only have guidance. I guess it's a headache to support that case?

pyxform/tests_v1/test_sheet_columns.py
```diff
@@ -38,6 +41,55 @@ def test_missing_label(self):
             error__contains=["no label or hint"],
         )
 
+    def test_label_node_added_when_hint_given_but_no_label_value(self):
```
lognaturel, line 44: We've tried as much as possible to move towards tests where the form is inline using `md_to_pyxform_survey`. Is there a strong reason not to do that here? We also have not been introducing new tests on full forms. Was there something that led you to have the test below which I believe is the same but on a real XLSForm?

lindsay-stevens: The `PyxformTestCase` machinery seems to simulate from the `workbook_to_json` step onwards. These end to end cases use real XLSX files in case that's too late for the relevant errors or warnings to surface (either now or in future). If preferred I can convert them to the md style test - please let me know.

pyxform/tests_v1/test_guidance_hint.py
lindsay-stevens: If it's preferred that a guidance hint alone (without regular hint or label) raises an error, please let me know.

lindsay-stevens:
> will Validate still throw an error if there's an empty label and no hint?

For this case there is an existing test: in `test_sheet_columns.InvalidSurveyColumnsTests.test_missing_label`: https://github.com/XLSForm/pyxform/blob/6bb547b409f46f78bb54af398b6da7e869ed418b/pyxform/tests_v1/test_sheet_columns.py#L36

For reference, the two e2e spreadsheets look like this:
hint_with_no_label_column:
hint_with_no_label_column
hint_with_no_label_value:
hint_with_no_label_value

pyxform/tests_v1/test_sheet_columns.py
lindsay-stevens: One other note - the `xml__contains` option of `PyxformTestCase` takes a sequence of XML snippets and looks for any match in the document. Like issue 448, this test is looking for an exact hierarchy (in case one of the questions doesn't output a label). Once 448 has a solution, this and other output tests would be refactored.

pyxform/tests_v1/test_guidance_hint.py
lognaturel: Ideally a guidance hint alone does raise an error because they can be configured to be hidden in some clients. If it adds a lot of complexity to have that case, it's not worth it. We know guidance hints are not very common.

lindsay-stevens: updated implementation and tests such that an error is thrown for guidance hint only

pyxform/tests_v1/test_sheet_columns.py
lindsay-stevens: removed e2e tests / files and updated newly added test to markdown style

pyxform/survey_element.py
```diff
@@ -414,19 +420,27 @@ def xml_hint(self):
             )
             return node("hint", hint, toParseString=output_inserted)
 
-    def xml_label_and_hint(self):
+    def xml_label_and_hint(self) -> "List[DetachableElement]":
```
lognaturel, line 423: Spiritually, I'm certainly on board with using type annotations. How are you using them in practice?

pyxform/xls2json.py
```diff
@@ -878,24 +878,6 @@ def workbook_to_json(
             )
             raise PyXFormError(error_message)
 
-        if (
-            constants.LABEL not in row
-            and row.get(constants.MEDIA) is None
-            and question_type not in aliases.label_optional_types
```
lognaturel, line 884: Are there tests for this case and the two below?

pyxform/survey_element.py
lognaturel: From @lindsay-stevens: IntelliJ does verification on an ongoing basis. No effect to runtime.

That seems useful enough to me to add as a tool/convention.

pyxform/xls2json.py
lognaturel: This was for warnings, not for errors. It is related to what @yanokwa was doing at https://github.com/XLSForm/pyxform/pull/542 so we can leave it in here and then look at #542 separately.

lindsay-stevens: Restored this but in a slightly different location. #542 can still add in the new check condition, and add the tests. Please see commit msg for 3e4a84e

lindsay-stevens:
Hi @lognaturel, resolving this turned out to be not as simple as restoring the check code block. When it was restored, a couple of tests needed to be updated. In particular there is an old-style test with a large "warnings.xls" fixture, which is now refactored into markdown. This should make it clearer when the label warning will and will not be raised. From here, #542 could add in the new condition to exclude warnings for field-list groups, and add the new tests. More context / info in the 3e4a84e commit message.

lognaturel (approved):
Thank you!

PR #544, opened 2021-07-06
Levenshtein distance check / warning for sheet name misspellings

### LINKED ISSUE #443: Catch basic spelling errors on worksheet names
For the `survey`, `choices`, and `settings` worksheets, it could be nice to catch basic spelling errors and warn the user.

So, eg, if there is no `settings` sheet, but there is some other sheet with a Levenshtein distance < 2 from `settings` (say somebody like @seadowg accidentally labelled it `setting`), a warning could be issued.

Filed on suggestion from @lognaturel.

### LINKED ISSUE #527: Warn if worksheet titles are not lowercase
An XLSForm with a "Survey" sheet fails with "The survey sheet is either empty or missing important column headers", but the same form with a "survey" sheet works.

I'm not sure there is a good reason for this. I propose we make the check case-insensitive, but perhaps convert it to lowercase immediately so we don't break anything downstream. Or maybe we just warn if any of constants.SUPPORTED_SHEET_NAMES are not all lowercase.

lindsay-stevens:
Closes #443
Closes #527

It implements the suggested heuristic for Levenshtein distance <= 2 to detect possible sheet name misspellings. It's possible that this will warn about sheet names for sheets that have auxilliary metadata that is not meant for processing by pyxform. For example the "osm" sheet name may be similar to many other initialisms. In that case, the warning prompts users to prefix such sheet names with an underscore to suppress the warning.

Other approaches could involve more complex spell check algorthims. The current Levenshtein distance algo is well documented and does a decent job at finding similar spellings. It seemed like overkill to add a module dependency just to have more advanced spell check.

As mentioned above, some users may like to store extra metadata in similarly named sheets to those that are part of the SUPPORTED_SHEET_NAMES constant. In which case they would need to rename the sheets to suppress the warning, if the warning is a problem.

I don't think a documentation change is needed.


pyxform/tests_v1/test_xls2json.py
```diff
@@ -0,0 +1,158 @@
...
+            test_dict = get_survey(sheet_name=n)
+            warnings = []
+            observed = xls2json.workbook_to_json(
+                workbook_dict=test_dict, warnings=warnings
+            )
+            self.assertDictEqual(get_expected_dict(), observed)
+            observed = [i for i in warnings if self.err_msg in i]
+            self.assertEqual(0, len(observed), observed)
+
+    def test_workbook_to_json__misspelled_survey_found(self):
```
lognaturel, line 79:
```
    def test_when_survey_sheet_not_found__and_close_names_found__lists_close_names_in_error(self):
        self.assertPyxformXform(
            name="survey_case_insensitive",
            md="""
            | SuRvE   |         |         |       |
            |         | type    | name    | label |
            |         | string  | bar     | baz   |
            | SuRvEys |         |         |       |
            |         | type    | name    | label |
            |         | string  | bar     | baz   |
            | Sruvey  |         |         |       |
            |         | type    | name    | label |
            |         | string  | bar     | baz   |
            """,
            errored=False,
            error__contains=['Could not find a sheet named \'survey\'.',
                             'Please check the sheet names and try again.',
                             'The following sheets with similar names were found: \'surve\', \'surveys\', \'sruvey\''.]
        )
```

pyxform/xls2json.py
```diff
@@ -412,6 +470,11 @@ def workbook_to_json(
 
     # ########## External Choices sheet ##########
 
+    warnings, _ = find_sheet_misspellings(
```
lognaturel, line 473: Is it possible all of these are getting lost/swallowed?

pyxform/xls2json.py
```diff
@@ -540,6 +619,9 @@ def workbook_to_json(
     )
     survey_sheet = dealias_types(survey_sheet)
 
+    warnings, _ = find_sheet_misspellings(
```
lognaturel, line 622: I don't think we should look for matches for OSM. Like you say, it seems pretty easy to get them and this sheet is not very important.

pyxform/xls2json.py
```diff
@@ -426,6 +489,11 @@ def workbook_to_json(
     # ########## Choices sheet ##########
     # Columns and "choices and columns" sheets are deprecated,
     # but we combine them with the choices sheet for backwards-compatibility.
+    warnings, _ = find_sheet_misspellings(
```
lognaturel, line 492: Since this is deprecated, let's not try to surface it.

pyxform/xls2json.py
```diff
@@ -434,11 +502,17 @@ def workbook_to_json(
...
+        key=constants.COLUMNS, dict_keys=workbook_dict.keys(), warnings=warnings
+    )
     columns_sheet = workbook_dict.get(constants.COLUMNS, [])
     columns_sheet = dealias_and_group_headers(
         columns_sheet, aliases.list_header, use_double_colons, default_language
     )
 
+    warnings, _ = find_sheet_misspellings(
+        key=constants.CHOICES, dict_keys=workbook_dict.keys(), warnings=warnings
```
lognaturel, line 514: Ideally this is error text added to the error when a choices sheet is needed but not present.

pyxform/xls2json.py
```diff
@@ -434,11 +502,17 @@ def workbook_to_json(
         default_language,
     )
 
+    warnings, _ = find_sheet_misspellings(
+        key=constants.COLUMNS, dict_keys=workbook_dict.keys(), warnings=warnings
```
lognaturel, line 506: This alias is never documented so best not to surface it again.

pyxform/xls2json.py
```diff
@@ -447,6 +521,11 @@ def workbook_to_json(
         choices_sheet, aliases.list_header, use_double_colons, default_language
     )
     # ########## Cascading Select sheet ###########
+    warnings, _ = find_sheet_misspellings(
+        key=constants.CASCADING_CHOICES,
```
lognaturel, line 525: This alias is never documented so best not to surface it again.

lognaturel (commented):
Great to see this!

Have you seen the messages when converting a real form? I am not seeing any of them. I believe `workbook_dict` is misleadingly described and in fact only includes recognized sheets. So it's always empty in cases where the recognized sheets aren't there. I believe your tests hide that because they explicitly set the dicts. As far as I can tell, all the tests can be converted to the more typical style which better matches real conversion.

Strangely, https://github.com/XLSForm/pyxform/pull/544#discussion_r665703905 does pass but when I use a real spreadsheet I don't see the message about close names. When I print `len(workbook_dict.keys())`, it is empty.

pyxform/xls2json.py
```diff
@@ -298,13 +308,52 @@ def process_range_question_type(row):
...
+        if 2 >= levenshtein_distance(_k.lower(), key)
+        and _k not in constants.SUPPORTED_SHEET_NAMES
+        and not _k.startswith("_")
+    )
+    if 0 < len(candidates):
+        msg += (
+            " The following sheets with similar names were found: {c}. "
+            "To suppress this message, prefix these sheet names with an "
+            "underscore. For example 'setting' becomes '_setting'."
+        ).format(c=str(candidates))
```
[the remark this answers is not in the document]
lindsay-stevens: sure

pyxform/xls2json.py
lindsay-stevens: If it is deprecated, could the code here that processes this sheet be deleted?

pyxform/xls2json.py
lindsay-stevens: In that case, could the code here that processes this sheet be deleted?

pyxform/xls2json.py
lindsay-stevens: In that case, could the code here that processes this sheet be deleted?

lindsay-stevens:
> Have you seen the messages when converting a real form? I am not seeing any of them.

Thanks for this review. It seems like I have put the spell check too far down the pipeline. As you mentioned, only supported sheet names would come through from the actual spreadsheet parsing step before it. I'll have another go and use end to end tests instead. This kind of issue is why I avoided using the `PyxformTestCase` - it's close to, but not fully, end to end.

pyxform/xls2json.py
lognaturel: Yes, let's remove.

pyxform/xls2json.py
lognaturel: Yes, let's remove.

pyxform/xls2json.py
lognaturel: Yes, let's remove.

lindsay-stevens:
Hi @lognaturel  - please see below change list. I think this addresses all of the above and what was discussed a week or so ago.

- update test style from dict + workbook_to_json to assertPyxformXform
- update spelling message to format candidates as quoted strings
- add all sheet names to workbook_dict so they can be used for spell
  check but otherwise only normally process supported sheets
- apply spell check rules:
  - only run spell check for choices or external_choices if required
  - always run spell check for survey (required) and settings (optional)
  - don't run spell check for OSM sheet
- remove code related to deprecated sheets cascading_choices and columns
- add end to end tests for the warnings, and inclusion of all sheet
  names in initial workbook dict
- add test cases for various scenarios with each sheet name with
  different assertions as per above rules

pyxform/tests_v1/test_xls2json.py
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+
+class TestXLS2JSONSheetNameHeuristics(unittest.TestCase):
+
+    err_msg = "The following sheets with similar names were found"
+
+    def test_workbook_to_json__case_insensitive_survey_sheet(self):
+        """Should not warn/error if the survey sheet is not lowercase."""
+        test_names = ("survey", "Survey", "SURVEY")
```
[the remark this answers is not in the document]
lindsay-stevens: Changed tests to markdown style

pyxform/tests_v1/test_xls2json.py
lindsay-stevens: added test that covers this scenario

pyxform/xls2json.py
```diff
@@ -298,13 +308,52 @@ def process_range_question_type(row):
...
+        _k  # thanks to black
+        for _k in dict_keys
+        if 2 >= levenshtein_distance(_k.lower(), key)
+        and _k not in constants.SUPPORTED_SHEET_NAMES
+        and not _k.startswith("_")
+    )
+    if 0 < len(candidates):
+        msg += (
+            " The following sheets with similar names were found: {c}. "
+            "To suppress this message, prefix these sheet names with an "
```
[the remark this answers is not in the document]
lindsay-stevens: updated choices error messages and added tests covering these scenarios

pyxform/xls2json.py
lindsay-stevens: removed related code for "choices and columns"

pyxform/xls2json.py
lindsay-stevens: moved the spell check to where the required-ness of the choices sheet is checked

pyxform/xls2json.py
lindsay-stevens: removed related code for "columns"

pyxform/xls2json.py
```diff
@@ -1077,11 +1118,14 @@ def replace_prefix(d, prefix):
...
+                        similar = find_sheet_misspellings(key=k, keys=workbook_keys)
+                        if similar is not None:
+                            msg = msg + " " + similar
                         raise PyXFormError(
-                            "There should be a choices sheet in this xlsform."
-                            " Please ensure that the choices sheet name is "
-                            "all in small caps and has columns 'list name', "
-                            "'name', and 'label' (or aliased column names)."
+                            msg + " Please ensure that the choices sheet has columns"
+                            " 'list name', 'name', and 'label' (or aliased column names)."
```
lognaturel, line 1128: This hasn't changed from the existing implementation but while we're here, how about using `list_name` which is what is documented and dropping "or aliased column names" since we'd really rather not users know about aliases.

lindsay-stevens: Error message updated in latest commit

lognaturel (approved):

lognaturel:
Apologies for the delay, here. Looks great. 🙏

PR #559, opened 2021-09-15
avoid recursion depth error for label with many references

### LINKED ISSUE #444: Maximum recursion depth can be reached when a label has many references
https://forum.getodk.org/t/error-maximum-recursion-depth-exceeded/26792

I think the first thing to do is to identify what recursive call this his hitting and whether it really does need to be recursive.

lindsay-stevens:
Closes #444

The root cause was the call to `copy.deepcopy`. The `child` argument is a minidom Node type which has a reference back to it's parentNode. The parent in turn has references to it's childNodes. So a deepcopy would try to traverse this recursive structure and error out. A copy of the parentNode is unnecessary anyway, since the copy is being made in order to append to a new parent.

The updated call to `child.cloneNode` seems to be the appropriate minidom API method for making a copy of a Node ([docs](https://docs.python.org/3/library/xml.dom.html#xml.dom.Node.cloneNode)). It doesn't copy the parentNode, thus avoiding the recursion issue. It's not clear from the pyxform git history why this wasn't used earlier. The method seems to have been available for at least 15 years. Maybe there was a bug in the implementation or usage in prior Python versions.

Another approach could be to use the child's `__dict__` or `__slots__` to filter out `parentNode` and copy everything else. But that may not work without making new DOM objects for the copied attributes. Then at that point this strategy essentially becomes the same as what `cloneNode` does.

Should not be any. In case there is an unknown edge/use case that requires a recursive copy, the `cloneNode` call requests a deep copy of childNodes. As noted in the new inline comment, there did not seem to be any tests which involved copying childNodes which themselves had childNodes. 

Probably not. The original bug report was perhaps an extreme use case.

### DIFF
--- a/pyxform/tests_v1/test_survey.py
+++ b/pyxform/tests_v1/test_survey.py
@@ -16,8 +16,7 @@ class TestSurvey(PyxformTestCase):
             |        | note | n    | {n}   |          |
             |        | text | q2   | Q2    | {r}      |
             """.format(
-                n="q1 = ${q1} " * 250,
-                r=" or ".join(["${q1} = 'y'"] * 250)
+                n="q1 = ${q1} " * 250, r=" or ".join(["${q1} = 'y'"] * 250)
             ),
         )
 
--- a/pyxform/tests_v1/test_survey.py
+++ b/pyxform/tests_v1/test_survey.py
@@ -1,0 +1,54 @@
+from pyxform.tests_v1.pyxform_test_case import PyxformTestCase
+
+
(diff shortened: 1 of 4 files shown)


lindsay-stevens:
The `formencode` commit eda825d was cherry-picked from #548 so that the CI tests can pass.

lindsay-stevens:
Hi @yanokwa or @lognaturel would this be OK to merge?

lognaturel (approved):
Excellent! It did seem like there might be something relatively simple at play.

PR #560, opened 2021-09-23
Add allow-mock-accuracy parameter

seadowg:
Needed for getodk/collect#4768

### DIFF
--- a/pyxform/tests_v1/test_allow_mock_accuracy.py
+++ b/pyxform/tests_v1/test_allow_mock_accuracy.py
@@ -12,7 +12,7 @@ class AllowMockAccuracyTest(PyxformTestCase):
             |        | geopoint  | geopoint    | Geopoint | allow-mock-accuracy=true |
             """,
             xml__contains=[
-                'xmlns:orx="http://openrosa.org/xforms"',
+                'xmlns:odk="http://www.opendatakit.org/xforms"',
                 '<bind nodeset="/data/geopoint" type="geopoint" odk:allow-mock-accuracy="true"/>',
             ],
         )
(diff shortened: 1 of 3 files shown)


lognaturel:
The required dependency update has been merged.

pyxform/tests_v1/test_allow_mock_accuracy.py
```diff
@@ -0,0 +1,119 @@
...
+    def test_geopoint(self):
+        self.assertPyxformXform(
+            name="data",
+            md="""
+            | survey |           |             |          |                          |
+            |        | type      | name        | label    | parameters               |
+            |        | geopoint  | geopoint    | Geopoint | allow-mock-accuracy=true |
+            """,
+            xml__contains=[
+                'xmlns:orx="http://openrosa.org/xforms"',
```
lognaturel, line 15: This should be testing for the `http://www.opendatakit.org/xforms` namespace with `odk` prefix.

lognaturel (changes requested):

pyxform/tests_v1/test_allow_mock_accuracy.py
seadowg: Oh I think this is a mistake in `test_audio_quality` as well right (which I'm pretty sure is where I was copying from). That's probably come from somewhere else then got missed. So these tests should be matching on `'xmlns:odk="http://www.opendatakit.org/xforms"'` instead right?

lognaturel (approved):

pyxform/tests_v1/test_allow_mock_accuracy.py
lognaturel: Whoops, good call fixing those.

PR #561, opened 2021-09-24
Fix appveyor errors

seadowg:
The version of formencode we're using seems to be having problems with appveyor. This fix is stolen from https://github.com/XLSForm/pyxform/pull/548.

From looking at the release notes for formencode 2.0.0 ([here](http://www.formencode.org/en/latest/whatsnew-2.0.html)) it looks like it no longer works with Python 2 (or anything under Python 3.6). As far as I understand, pyxform still supports these versions, so I'm not sure if this upgrade actually is the right fix. An alternative might be to downgrade to 1.3.0 if that also works. 

It could also be that there is an entirely different solution! I'm still not quite sure why appveyor is running into problems with formencode 1.3.1.

See above.

Maybe if we're planning to drop support for some Python versions (although just to the README)

### DIFF
--- a/requirements.pip
+++ b/requirements.pip
@@ -12,7 +12,7 @@ black==19.10b0
 click==7.0                # via black, pip-tools
 entrypoints==0.3          # via flake8
 flake8==3.7.7
-formencode==1.3.1
+formencode==2.0.0
 isort==4.3.20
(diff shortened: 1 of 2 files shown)


yanokwa:
Closing in favor of https://github.com/XLSForm/pyxform/pull/548.

PR #562, opened 2021-09-25
clean up dependencies, py2 code, etc 🚿 

### LINKED ISSUE #558: Clean up dependencies
As encountered in #548, there are numerous dependencies which should be cleaned up to avoid confusion during development and to ensure pyxform is using up to date dependencies (fixes / security / etc).

Main ideas for this ticket:

1. `setup.py` declares 5 dependencies, but it seems like only `xlrd` and `unicodecsv` are used by prod code. Usages of `formencode` are in tests only, so it could move into the dev requirements `requirements.in`. The `unittest2` and `functools` don't seem to be required anymore for currently supported python 3.7 / 3.8. `unicodecsv` might be straightforward to replace with stdlib `csv`, if it's the case that the external package was used only because of python 2 / unicode compatibility.
2. `requirements.pip` seems like it may have been generated at one stage then manually updated thereafter. That's my guess since the output format from the latest `pip-tools` (latest 6.2.0 vs used 3.7.0) is somewhat different. In any case the `requirements.in` should pin to current versions and regenerate `requirements.pip` so it's all up to date.
3. Remove the py2 compatibility conditional imports, i.e. instances of `try: import x; except ImportError: import y as x`. There's about 10 of these based on a search for `ImportError`.

Related ideas for separate tickets:

- `xlrd` is pinned to an old version 1.2.0 but this seems to be because afterwards, support for .xslx was removed and users were encouraged to use `openpyxl` for that format instead. So it would likely involve adding a dependency, and code to detect / swap library depending on input format (either by file name or some fancy detection).
- `formencode` is pulled in just for it's `xml_compare` function in the older `tests` package. The function is a pretty neat structural comparison but the dependent tests could probably be refactored to use `tests_v1` style `PyxformTestCase`, meaning that the dependency could be removed altogether.

lindsay-stevens:
Closes #558

todo

todo

todo

- [ ] included test cases for core behavior and edge cases in `tests_v1`
- [ ] run `black pyxform` to format code


lognaturel:
Have taken a quick look and don't see any issues. 👍 I'm hoping it will be a quick rebase on master to fix conflicts. I'll finish the review today or tomorrow.

lindsay-stevens:
resolved merge conflicts

lognaturel (approved):
I went through commit by commit and validated that the action made sense. I also did a quick check for each commit to make sure that it performs the action described by the commit message and does so correctly. Thanks for separating out the commits -- this would have been impossible to do any kind of reasonable review of otherwise.

There are some duplicate commits that were introduced in merges and rebases. Though the history was very helpful in review, I think we should squash this. I'll give a day or two for comments before doing that.

The two things that I did some thinking about are 
* Moving all the tests. It's unfortunate that `blame` won't provide anything useful anymore. I find it occasionally useful to find out when and why a test was introduced and that will be much harder. But it's probably not a huge deal.
* @KeynesYouDigIt has two open PRs that I never seem to be able to get a chance to review. This will cause conflicts. I propose we still merge this first and we can help with conflicts when reviewing.

lindsay-stevens:
Great, thanks @lognaturel !

About the tests and blame, the moves were tracked (`git mv`) so they should still show up in blame. At least, they do for me - both on CLI and GitHub when browsing the PR branch. One exception being 1750d59 where I copied the XFormTestCase definition lines into a different file. But we can still go to the parent commit and blame from there if needed.

About the conflict PRs, I agree it would be simpler to proceed with this PR and resolve conflicts in other PRs; than to try to resolve them first then incorporate into this PR.

About squashing, I don't mind either way. Many of the commits are noisy but some have potentially useful context.

PR #564, opened 2021-10-09
Issue 157 warn about missing translations

### LINKED ISSUE #157: Sometimes users don't apply consistent language tags across all translatable columns

KeynesYouDigIt:
Closes #157 

This approach ensures that the warning work has been done along side the work that puts in placeholders for missing translations. It is focused on making sure we never re-evaluate the form for missing languages but uses what is already implied in the form build.

Last time I attempted to solve this issue, I assumed the existence of certain pieces of the `path` string and tried to build warning messages using those pieces. When the warnings function tried to parse paths without what I assumed to be present, it broke. I do not believe this risk is still present, but I do assume things like path and content type are not null in this function. This feels like a safe assumption, so I dont think the risks are great.

No


KeynesYouDigIt:
@lindsay-stevens Thanks so much for that big PR fixing up the tests!!! it got CI to pass!

@lognaturel I was able to clean up and shorten this PR, Hoping it can get a look soon?

pyxform/survey.py
```diff
@@ -703,8 +705,49 @@ def _add_empty_translations(self):
...
+            # Images, media will hit this block, as well as other content_types missing translations.
+            return content_type
+
+        # In the "path", the actual column that needs the translation added should appear after the ":"
+        survey_column_name_ix = path.find(":")
+        if survey_column_name_ix > -1:
+            return path[path.find(":") + 1 :]
+
+        # if the ":" is not present, there may be a choice label present.
+        # First check for a "-"
```
KeynesYouDigIt, line 733: This relates to the regression that occurred on my first try

KeynesYouDigIt:
@lindsay-stevens 

Thank you so much for the comment and feedback! I have been needing some guidance and collaboration badly here, as you can see :smile: 

Sorry in advance for so much text, the TL:DR is:

- I will reconsider the approach and see if I can make something work more upstream. (do you think we could we move all the translation work up here?)
- I will remove the refactor.
- I internal__warnings was added because of the approach I took,  and `to list on to json` was added to support serializing the internal__warnings. 

Please feel free to @ me on slack if you would rather decline the PR and move me to a simpler or more urgent set of work, or if you want to chat more about what I've done here. I mostly only have time for this on weekends, but could get away from work on Fridays, Thursdays, and maybe Tuesdays.

Longer version - 
> Could you please clarify the current approach and whether the above xls2json.py locations were considered?

I considered and even preferred the upstream approach here, and to be honest now that you bring it up again I think you might be correct. But let me explain why I went downstream:

What caused me to pick this downstream approach is that we actually build the translations themselves downstream. The "state" of a missing translation is detected  and handled at this point, so seems redundant to me to detect it and warn about it earlier. You are right that it forces us to reverse engineer the xpath, which I thought would be easier than it turned out to be :/ 

I am going to spend some time today reconsidering that upstream path again. What I would really like to do is move the _translations_ upstream - when I first had that thought it sounded like too big of a change, but looking at it again I think it would at least be worth spending a few hours seeing what that would look like. Now, if we calculate missing translations upstream, but only create translations downstream, that _feels_ redundant, but I suppose it might not be. (and redundancy might be better than reverse engineering?) 

> It seems a bit peripheral to this PR anyway so maybe it should be moved to a new PR?
It is, at my company we follow the "boy scout rule" (leave it better than you found it) as a way of packing refactors into PRs. I _think_ the tests cover the case you had mentioned, but I am happy to just back it out for simplicity's sake - the "boy scout rule" might doesn't make sense for this project? (small team,  less frequent updates, testing that's more situation-ally driven and doesn't cover small units of work like this particular chunk of code)


> so why not use warnings only?

I don't (think?) ther'es a way to access that warnings object in the "guts" of where I do this work (this is another hazard of the downstream approach I am using). I do think its a bit odd that the internal workings can't pass information back up to the user, so while the approach of adding another mutable/stateful attribute was hardly ideal, this was the quickest way I could make it work.

If I can make the more "upstream" approach work we can remove this for now. My guess is that we will need something that achieves the same later? but I could be wrong. This is a pretty mature codebase and it hasn't needed it before :shrug:

KeynesYouDigIt:
@lindsay-stevens 

Reviewing the sections you linked me to - dealias_and_group_headers is very agnostic as to what the "tokens" it works with actually are, which makes it hard to detect if the token is a language. It also works row-by-row, but translations are added column-by-column, so we would have to review the aggregated data again.

I could certainly add something like `find_uneven_translations(out_dict_array)` at the end of that function, but I am still worried I'd duplicating work with the stuff done in the Survey class (see `_add_empty_translations` and `_setup_translations` ).

I haven't done enough work to see if its possible, but what do you think of trying to move some or all of the work in _setup_translations and/or _add_empty_translations to this upstream point? Perhaps we could build JSON elements that are much easier to work with down stream. Before I spend time trying to code a proof of concept, is that what you were getting at with your earlier comment? Or are we strictly trying to detect unbalanced translations and move on? Again, I think that would require duplication of logic, but I am not sure.

I'll keep mulling this over in my mind, still not sure what the best way forward is, LMK if you have any thoughts or if I've missed something here.

lindsay-stevens:
Hi @KeynesYouDigIt 

The `dealias_and_group_headers` reference was meant as more of an example of examining the XLSForm data and processing/validating. To implement this I think a new function would be appropriate, particularly since `dealias_and_group_headers` is used by multiple sheets (Settings, Survey, Choices, External Choices, OSM) which don't all have translatable columns.

There are finite translatable column types, as noted in [this comment](https://github.com/XLSForm/pyxform/issues/157#issuecomment-623650826). So after splitting column names on `::`, any columns not in a 'translatable' list / mapping could be ignored. Even if it's not currently possible to determine with all certainty what all the translatable columns are, the referenced comment identifies what are likely the most commonly used ones, so the warning would still be valuable.

About the duplication concerns, I think these may overlap code-wise but the purpose is different. The goal for #157 is a warning to the user, which they may ignore at their peril. The existing `_add_empty_translations` seems to be about structural integrity in the XForm XML. For example if I comment out the body of `_add_empty_translations`, then in `pyxform_test_case.py` set `run_odk_validate = kwargs.get("run_odk_validate", True)`, then run all the tests, there are 5 failures from errors from ODK Validate.

I would recommend aiming to address #157 as directly as possible. If it looks like a refactor is needed or potentially beneficial, then please create a separate issue/PR to describe/discuss/implement it. My concern is that there aren't many tests around translations behaviour, so embarking on that at the same time could stymie the PR or cause regressions.

lognaturel:
Thanks @KeynesYouDigIt for sticking with this one and exploring from multiple angles. I'd like to propose @lindsay-stevens take this one over and put up a new PR. Then you can use what you've learned to do an initial review and evaluate his approach. 

As @lindsay-stevens says, if there's an area you feel really needs some refactor, please file an issue for now.

lindsay-stevens:
@lognaturel sounds good I'll take a look at this in a new PR.

KeynesYouDigIt:
@lindsay-stevens / @lognaturel Sounds good to me as well, thanks for all the help!

lindsay-stevens:
closing in favour of #571

PR #570, opened 2021-11-01
Improve xml hierarchy test methods, refactor test_trigger.py tests to use them

### LINKED ISSUE #563: Refactor test_trigger.py tests to use xml hierarchy test methods
PR #548 added xml hierarchy test methods `xml__xpath_exact` and `xml__xpath_count`. The originating issue #448 describes cases related to trigger tests where the `xml__contains` approach was fragile. Many of the tests in `test_trigger.py` could be improved by using the new test methods.

lindsay-stevens:
Closes #563

This follows on from #548 to add the discussed `xml__xpath_match` kwarg for `PyxformTestCase` and to switch to the lxml package from ElementTree for the purposes of test XPath evaluation. The benefit of lxml is that it's XPath 1.0 processor can do a lot more than ElementTree's XPath subset. Gains include being able to write compound predicates (with `and`) in one predicate instead of like `.//bind[1=1][1=1]`, the use of axes, and numerous XPath functions. Also it allows the old code for switching XML tokenizers for Python 3.7 to be removed. More thorough tests for the `xml__xpath*` matchers are added as well.

The way around the default namespace limitation in lxml (mentioned in #548) was to tell the XPath processor about the same namespace URI/URL but assign a non-empty and not-None prefix to it. The lxml docs say the namespace match is actually on the URI not the prefix, so it's a bit like SQL aliases in different SELECT clauses. Since the default namespace is `http://www.w3.org/2002/xforms` I went with a prefix of `x` for this purpose. The assignment happens in `pyxform_test_case.py` on L231 which is finally used at L556. This means that in any usages of the `xml__xpath*` test matchers, XPath expressions referring to elements in the default namespace need to be written like `.//x:bind` instead of `.//bind`. Attributes without a prefix don't require this `x` prefix. There are numerous examples in `test_pyxform_test_case.py`.

This PR also delivers on the main purpose of #563 to refactor the `test_trigger.py` tests to use the new `xml__xpath*` matchers. With above improvements and addition of `xml__xpath_match`, it was all possible with `xml__xpath_match`.

Further details in the commit messages.

This is a new testing feature so there would be no regressions unless other branches had already started using it. 

It's possible that the refactors of `test_trigger.py` test are not perfect, despite being careful. Each of these tests passes. The XPath results can be checked by setting `debug=True`, which prints out the XML document, each XPath expression and what it matched.

Not applicable.


lognaturel (approved):
I focused mostly on the updated tests which look good to me. I agree with the rationale for introducing `lxml`. The approach for the default namespace makes sense.

I'm wondering whether we should strongly recommend using `xpath_match` for new tests now that all this investment has gone into making it nice? The biggest benefit I see is that it removes the attribute order issue entirely and makes tests more robust against changes to things like whitespace in the XForms output. If you agree, @lindsay-stevens, would be great to add to the README.

lindsay-stevens:
Thanks @lognaturel I agree and have added a section to the README for this. I've focussed on `xml__xpath_match` because after having used it I think that will probably be all that's needed in >95% of cases.

lindsay-stevens:
Appveyor build error seems to be random pip error, I only added to the Readme since the passing build in baec985

PR #575, opened 2021-11-30
use openpyxl for XLSX files; upgrade xlrd to v2

### LINKED ISSUE #501: Explore alternatives for reading Excel files because xlrd has been deprecated for a long time and is reported to be unstable
xlrd has been passively maintained since 2016: https://groups.google.com/g/python-excel/c/P6TjJgFVjMI/m/g8d0eWxTBQAJ

xlrd 2.0 was [recently released](https://groups.google.com/g/python-excel/c/IRa8IWq_4zk) and removes xlsx support. The recommended package for reading xlsx is openpyxl: http://www.python-excel.org/

The former maintainer of xlrd says "the xlsx reading in xlrd has become unreliable in Python 3.9".

Some questions to investigate:
- [x] what would be the impact of dropping xls support? -- no go at the moment, there are still too may floating around
- [x] is openpyxl actually any good? -- seems to be what folks have rallied around
- [ ] how big of a job would it be to replace xlrd with openpyxl?
- [ ] when is this likely to become a real problem?

sheppard:
Closes #501

See discussion in #501.

There should be no noticeable changes for most users of pyxform.  I started from a line-for-line translation of the xlrd integration and optimized some of the logic.  One potential change is that dictionary keys will not contain consecutive inner spaces, since they are now populated from the cleaned headers rather than a separate step.

There is a new dependency, but this should be transparent to most users.

- [X] run `black pyxform tests` to format code


lindsay-stevens:
Hi @sheppard thanks for adding this PR! I have a few requests to consider please. If you don't expect to have the bandwidth to address these please let us know, and they could be done in a follow-up PR.

- Pin setup.py dependencies to latest versions. Pyxform users may have older versions which may not necessarily work: `xlrd==2.0.1` and `openpyxl==3.0.9`.
- A requirement for adding openpyxl should be to include defusedxml, which seems to involve:
  - add a setup.py dependency pinned to `defusedxml==0.7.1` so it is importable by openpyxl.
  - `os.environ["OPENPYXL_DEFUSEDXML"] = "True"` in xls2json_backends.py, as per: [openpyxl src](https://foss.heptapod.net/openpyxl/openpyxl/-/blob/5645879037fdf6f51292ec40edc31aa64e0ec1f6/openpyxl/xml/__init__.py) and [docs](https://openpyxl.readthedocs.io/en/stable/index.html#security).
- Add a test to test_xls2json_backends.py to explicitly exercise XLS (xls_to_dict) and XLSX (xlsx_to_dict) reading. This could be copy of [CsvReaderEquivalencyTest](https://github.com/XLSForm/pyxform/blob/66c0d5b9ad4a34bce47fe52a60e4366c19ee457c/tests/xls2json_tests.py#L210-L227) for XLS / XLSX instead of CSV / XLS.
- Move utils.py function-level imports to module level, since they are either internal imports or package-level dependencies. I'm not sure why there was a function-level import in sheet_to_csv.

sheppard:
I updated the setup.py dependencies, added equivalency tests, and moved the imports to the module level in utils.py.

It looks like openpyxl's defusedxml integration is active by default as long as:
 * defusedxml is installed, and
 * `OPENPYXL_DEFUSEDXML` is either `True` **or not set**.  (In spite of the name, `defusedxml_env_set()` returns `True` if the key does not exist in the dictionary.)

I confirmed this by adding a test to `assertTrue(openpyxl.DEFUSEDXML)` in test_xls2json_backends.py.

Setting `os.environ` would only be necessary to override the preferences of a user that had explicitly defined `OPENPYXL_DEFUSEDXML=False` in their environment.  In that case, the user would still be able to disable defusedxml by importing openpyxl before pyxform had a chance to update `os.environ`.

lindsay-stevens (approved):
Great! Thanks very much

PR #577, opened 2022-01-14
461: Add value and label params to select_.._from_file

### LINKED ISSUE #461: Add value and label params to select_.._from_file
#### Problem description

When creating select options from external (and internal) data in XLSForm, the "name" and "label" data columns are hardcoded in the XForm output (and therefore required in the data), when using:

* select_one_from_file filename.csv & select_one_from filename.xml
* select_multiple_from_file filename.csv & select_multiple_from filename.xml

This means that you'd normally have to modify the data file to rename columns (or XML nodes), which is not great.

This is purely a pyxform/xlsform restriction because Collect, Enketo, iXForms etc are perfectly capable to deal with other names.

Discussion: https://forum.getodk.org/t/proposal-customize-name-and-label-columns-in-selects-from-external-data-in-xlsform/27265

#### Feature

Customize value and label names using the `parameters` column:

##### 1
| type | parameters | name | label |
|-------|-----------------|---------|--------|
| select_multiple_from_file hh-data.csv | value=hh_number, label=hh_name | a | Select |

##### 2
| type | parameters | name | label |
|-------|-----------------|---------|--------|
| select_one_from_file hh-data.csv | value=hh_number, label=hh_name | a | Select |

##### 3
| type | parameters | name | label |
|-------|-----------------|---------|--------|
| select_multiple_from_file hh-data.xml | value=hh_number, label=hh_name | a | Select |

##### 4
| type | parameters | name | label |
|-------|-----------------|---------|--------|
| select_one_from_file hh-data.xml | value=hh_number, label=hh_name | a | Select |

Output includes (for all 4 cases): 

```xml
<itemset nodeset="instance('hh-data')/root/item">
       <value ref="hh_number" />
       <label ref="hh_name" />
</itemset>
```

lindsay-stevens:
Closes #461

- `xls2json.py`: Added value, label to allowed params for select_from_file question types.
- `question.py`: Updated to use params (if any) as the default itemset_ref value. Other logic around dynamic labels etc then layers on top.
- `test_external_instances_for_selects.py`: 
  - Added tests for select_from_file question types (csv or xml) without params, with params, with choice_filter, and both params and choice filter. 
  - Refactored existing select_one_external tests to check the same scenarios as above, though they only work with choice_filter anyway. Also replaced an end-to-end test for itemsets.csv generation that had associated XLSX/XML/CSV fixture files with a new `md_table_to_workbook` method since openpyxl can write files.
  - Added test for parameter value check (see below) for both select_from_file question types (csv or xml) and various good / bad param values.

ODK Validate rejects certain characters for the `value` and `label` parameters. Presumably the `ref` value is placed into a XPath, and characters that are XPath operators cause a problem. The error from ODK Validate looks like the below.

```
>> Something broke the parser. See above for a hint.
org.javarosa.xpath.XPathException: XPath evaluation: Parse error in XPath path: [val*].
Bad node: org.javarosa.xpath.parser.ast.ASTNodeAbstractExpr@63e2203c
```

Since this isn't super clear, the parameter values are validated by pyxform with the regex `^[a-zA-Z_][a-zA-Z0-9\-_\.]*$`. The error message is similar to other design issues, e.g.

> [row : 2] Parameter 'value' has a value which is not valid. Values must begin with a letter or underscore. Subsequent characters can include letters, numbers, dashes, underscores, and periods.

None are anticipated.

- The [main XLSForm.org page](https://xlsform.org/en/#multiple-choice-from-file) should be reworded, to clarify that either the default names can be used, or custom names can be used by specifying the new parameters.
- The [Reference Table](https://xlsform.org/en/ref-table/) for select_one_from_file and select_multiple_from_file should mention the available parameters in the notes (as well as `randomize` and `seed`).
- PR for the above here: https://github.com/XLSForm/xlsform.github.io/pull/223

- [X] run `black pyxform tests` to format code


lognaturel:
I've updated Validate to https://github.com/getodk/validate/releases/tag/v1.17.0 and run `nosetests` without the new test erroring so I think it should work. I'm not as confident as I'd like to be because I wasn't able to run the automatic Validate update script (something has changed there?) and I'm having trouble running tests in my IDE. But hopefully you can do the update and move forward, @lindsay-stevens.

lognaturel:
I've taken a first look through and this looks great to me. I'll take a deeper dive asap. @MartijnR perhaps you'll be interested in a peek as well? I think you were particularly interested in this change at some point.

MartijnR:
Thanks! Very cool! I played around converting some forms and wasn't able to break it.

pyxform/xls2json.py
```diff
@@ -1115,9 +1115,16 @@ def workbook_to_json(
                 new_json_dict = row.copy()
                 new_json_dict[constants.TYPE] = select_type
 
-                # Look at parameters column for randomization parameters
+                select_params_allowed = ["randomize", "seed"]
+                if parse_dict["select_command"] in (
+                    "select_one_from_file",
```
lognaturel, line 1120: I don't think this limitation is strictly necessary. That is, you could have an internal secondary instance with arbitrary column names if it's a resource pasted in from somewhere else. I'm not sure how compelling that is since then it's so easy to just rename the columns so fine by me to initially release with the restriction. Now we have a record of this thought and can be on the lookout for users trying to use arbitrary column names from internal datasets.

lognaturel (approved):
One comment for posterity, one merge for the present.

pyxform/xls2json.py
lindsay-stevens: One reason for the limitation is that when a translation is added, the XForm choice `@ref` attributes at e.g. `body/select1/itemset/value/@ref` (or `../label/@ref`) point to the itext e.g. `jr:itext(itextId)`, so the name can't be customised in the way implemented here for select_.._from_file.

lognaturel: Ah yes, indeed. I've filed https://github.com/XLSForm/pyxform/issues/593 for specifying localized columns for `select_.._from_file`. I guess it probably won't be practical to harmonize internal and external instances in this respect.

PR #578, opened 2022-01-17
495: Default string values with dashes are mistakenly treated as dynamic defaults

### LINKED ISSUE #495: Default string values with dashes are mistakenly treated as dynamic defaults
Since dynamic defaults 

#### Problem description
Dynamic defaults use a regular expression that looks for `-` and other operators. This can lead to bad behavior when a string with dashes is used. This is particularly likely when setting a default for a select because it seems form designers favor dashes over underscores for separating words in choice names (e.g. `a-choice`).

Create a form with a dynamic default that has a dash in it meant to be part of a string. For example, [this one](https://docs.google.com/spreadsheets/d/1VTtwlCIlOofTgoLBVeoIauGj6TJg8TwDmhvClfdZKUU/edit#gid=0). See that the resulting XML when converting has a `setvalue` when the value should be used literally.

#### Expected behavior
The literal default value should be set in the `instance`.

#### Other information 
I wasn't too worried about the regular expression when we initially did this because I was thinking that literal values would work fine with `setvalue`. However, literal values in XPath must be surrounded by quotes. I'm not sure what the best solution is.

lindsay-stevens:
Closes #495

At this stage just a draft PR with tests to reproduce the issues described in 495 and 533.

TBC

TBC


lindsay-stevens:
Hi @lognaturel there may be some more to do on this wrt. tests but I think it is ready for a look. Thanks!

tests/test_dynamic_default.py
```diff
@@ -337,3 +339,203 @@ def test_dynamic_default_on_calculate(self):
...
+            # TEXT
+            # Literal text containing URI; https://github.com/XLSForm/pyxform/issues/533
+            self.Case(False, "text", """https://my-site.com""", ""),
+            # Literal text containing brackets.
+            # French translation "Oui" here to satisfy Validate for jr:itext call below.
+            self.Case(False, "text", """(https://mysite.com)""", "Oui"),
+            # Literal text containing URI.
+            self.Case(False, "text", """go to https://mysite.com""", ""),
+            # Literal text containing various non-operator symbols.
+            self.Case(False, "text", """Repeat after me: ~!@#$%^&()_""", ""),
```
lognaturel, line 384: lol

tests/test_dynamic_default.py
```diff
@@ -337,3 +339,203 @@ def test_dynamic_default_on_calculate(self):
...
+            # Node paths with operator.
+            self.Case(True, "text", """../t2 - ./t4""", ""),
+            # Math expression.
+            self.Case(True, "text", """1 + 2 - 3 * 4 div 5 mod 6""", ""),
+        )
+        # Additional cases needing different approach for markdown / xpath assertions.
+        self.case_data_extras = (
+            # TODO: pyxform references appear in result as resolved XPaths
+            # Pyxform reference.
+            self.Case(True, "text", """${q0}""", ""),
```
lognaturel, line 456: I'm not clear on what this todo is about? Is the issue that the test XML output will have `/data/q0` in it? Actually I can't figure out how it's passing?

lognaturel (commented):
Definitely more involved than I would have gone for but more correct, too!

Here are some cases I played around with:
```
            self.Case(True, "date", """concat('2022-03', '-14')""", ""),

            self.Case(False, "text", """f-4""",""),
            self.Case(False, "text", """./f-4""",""),
            self.Case(True, "text", """./f - 4""",""),
            self.Case(False, "integer", """7-4""",""),
            self.Case(True, "integer", """7 - 4""",""),
```

The behavior with `-` is a change so that could possibly break some form updates. I can't imagine these types of cases are common but it might be good to have the tests as explicit documentation. And as you said, also add something to user-facing docs.

lindsay-stevens:
Thanks for the review! I will update the tests as mentioned above to complete this PR.

tests/test_dynamic_default.py
lindsay-stevens: Yep the issue is that the output has the resolved XPath e.g. `/data/q0` form rather than `${q0}`. Possibly a matter of reworking the tests to allow for this case specifying an alternative expected value. The test passes with this Case because the `self.case_data_extras` items are only passed through `utils.default_is_dynamic` in the shorter test, not `assertPyxformXform` in the longer test.

tests/test_dynamic_default.py
```diff
@@ -337,3 +339,203 @@ def test_dynamic_default_on_calculate(self):
...
+        |         | c1        | a-1  | Choice A-1 |
+        |         | c1        | a-2  | Choice A-2 |
+        |         | c2        | 1-1  | Choice 1-1 |
+        |         | c2        | 2-2  | Choice 1-2 |
+        |         | c3        | a-b  | Choice A-B |
+        """
+        self.assertPyxformXform(
+            name="test",
+            md=md,
+            run_odk_validate=True,
```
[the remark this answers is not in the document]
lindsay-stevens: Added tests for this

tests/test_dynamic_default.py
lindsay-stevens: Resolved this by updating the test strategy for cases that initially or after rendering contain single quotes, namely to use xpath_exact to compare outside of xpath, instead of xpath_match. There is one remaining "extra" case for the union operator, but the fate of that case depends on https://github.com/getodk/xforms-spec/issues/293

lindsay-stevens:
Should be good to go now. New commit is all about tests, per commit message for [ba7b196](https://github.com/XLSForm/pyxform/pull/578/commits/ba7b1967973da39002a89287bf27dde8cb6e7f15):

- deleted existing tests where covered by case in TestDynamicDefaultSimpleInput
- updated tests using string matchers to use xpath instead
- added pyxform_test_case ability to escape literal pipe in values
- clarified existing translation performance test description
- added util function for coalesce
- added tests suggested by @lognaturel and performance tests
- updated test strategy for cases that initially or after rendering
  contain single quotes, namely to use xpath_exact to compare outside
  of xpath, instead of xpath_match.

The performance tests are:
- `test_dynamic_default_performance__time`: processing time for 500, 1000, 2000, 5000 dynamic default questions shouldn't take much longer than no-op dynamic defaults check (results were approx 0 to 6%).
- `test_dynamic_default_performance__memory`: question with 2000 dynamic default questions shouldn't increase memory usage more than x2.

tests/test_dynamic_default.py
lognaturel: Ah yes, intriguing.

lognaturel (approved):

lognaturel:
🥳 🥳 🥳

Issue #568, opened 2021-10-20
Consider removing CircleCI, codecov


yanokwa:
Placeholder for the questions I raised at https://github.com/XLSForm/pyxform/pull/567.

We'd remove CircleCI in favor of Github Actions, remove codecov because it's unused can be the source of [security issues](https://about.codecov.io/security-update/), and move the formatting check into Github Actions as well.

PR #580, opened 2022-01-20
555: Add support for Python 3.9

### LINKED ISSUE #555: Add support for Python 3.9
Python 3.9 has been out since Oct 2020. We should add support for it in our build scripts along with Python 3.7 and 3.8.

- [x] Figure out what to do about xls support (https://github.com/XLSForm/pyxform/issues/501)
- [ ] .appveyor.yml
- [ ] .circleci/config.yml
- [ ] .github/workflows/pypi.yml

lindsay-stevens:
Closes #555

A GitHub Actions yaml file was in place already with a `release` job for creating a build with Py3.8 and linux (ubuntu) and uploading it to PyPI. This PR updates the yaml to add jobs for:

- `lint`, with Py3.8 on linux, by running `black`, `isort`, `flake8`, and `pycodestyle`
- `test`, with Py3.7, 3.8, and 3.9 on linux, macos, and windows, by running `nosetests`
  - At the end of the `test` job the sdist and wheel are uploaded as Artifacts ([example here](https://github.com/lindsay-stevens/pyxform/actions/runs/1724468708)).

The above more or less covers what the AppVeyor and CircleCI configurations are doing. So if this all looks good I could add removal of those CI dotfiles to this PR to close out #568 as well.

Some minor changes were required for Py3.9 compatibility. These are noted in e9c77ab and 3b21ce9, and involved removing a method call and a kwarg which were deprecated since Py3.2 and Py3.1 respectively.

Unlikely, but maybe the test CI environments are slightly different. But the tests pass all the same.

No, it would only affect developers.


lognaturel (commented):
Looks great to me and I'm in favor of removing reliance on codecov, circleci and appveyor. Doing that in this PR seems fine.

Let's give @yanokwa a few days to review before merging.

lognaturel (changes requested):
Realized the workflow should probably be renamed since PyPI is only involved for releases. Maybe just `pyxform`?

.github/workflows/pypi.yml
```diff
@@ -3,31 +3,126 @@ name: PyPI
 on: [push, pull_request, release]
```
lognaturel, line 3: I suggested renaming the workflow but maybe it makes more sense to have a "verify" workflow that is run on all these actions with jobs `lint` and `test`. And then a separate "release" workflow that runs only on release. That way you don't need the conditional in the `release` job and overall it would feel more event-based.

lindsay-stevens: Thanks, renamed existing workflow PyPI -> `release`, and moved test/lint jobs into new `verify` workflow.

lindsay-stevens:
`./.circleci/config.yml` and `./appveyor.yml` have been deleted so these checks failed. `./.coveragerc` also deleted.

lognaturel (approved):

PR #582, opened 2022-01-27
Remove badges for unused services

lognaturel:
Updates readme to reflect services currently in use.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,18 +2,9 @@
 pyxform v1.7.0
 ===============
 
-|circleci|  |appveyor| |codecov| |python| |black|
+|python| |black|
 
-.. |circleci| image:: https://circleci.com/gh/XLSForm/pyxform.svg?style=shield&circle-token=:circle-token
-    :target: https://circleci.com/gh/XLSForm/pyxform
-
-.. |appveyor| image:: https://ci.appveyor.com/api/projects/status/github/XLSForm/pyxform?branch=master&svg=true
-    :target: https://ci.appveyor.com/project/ukanga/pyxform
-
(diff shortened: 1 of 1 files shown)


lognaturel:
I was hoping that this would give a clean commit we could use to get the build green again since I forgot to disable the AppVeyor and CircleCI webhooks before the last merge. However, AppVeyor tasks are still running for reasons I don't understand. Anyone else know what needs to be done to disable AppVeyor entirely?

lindsay-stevens:
Maybe this was already done, and I can't see the Settings tab for this repo, but there should be a Webhooks section where they can be deleted (example screenshot below).

image

Failing that, the Appveyor config points to https://ci.appveyor.com/project/ukanga/pyxform so maybe @ukanga could delete (or disable) the pyxform project there?

Also there's the profile-level (or XLSForm org-level) OAuth privileges which may still be allowing Appveyor to run checks (example screenshot below)

image

lindsay-stevens (approved):

lognaturel:
Thanks, @lindsay-stevens! I've disabled the webhooks and that worked for CircleCI. I don't have org-level admin access so I can't see the screen you captured. Could you please revoke AppVeyor and CircleCI access?

lindsay-stevens:
@lognaturel oh I meant the example OAuth screenshot is from my personal profile, and my guess is there is something similar for orgs. I am a member of the XLSForm org but probably only an "owner" member can do this sort of thing. I think we will need assistance from @ukanga on this one.

lognaturel:
Aha! @yanokwa was able to remove the application at the org level. Thanks for the tip, @lindsay-stevens.

PR #586, opened 2022-02-10
Add accuracy parameters for geopoint

### LINKED ISSUE #579: Add accuracyThreshold and unacceptableAccuracyThreshold as parameters for geopoint
There's no explicit `pyxform` support for `accuracyThreshold` now though the XLSForm site documents it using the generic body attribute column structure: https://xlsform.org/en/#gps-with-accuracythreshold Adding it to the parameter column as `capture-accuracy` will make it more accessible.

`unacceptableAccuracyThreshold` is new to ODK Collect. It configures a threshold for the worst accuracy that will be labeled as "unacceptable" to help encourage data collectors to wait or improve conditions for a higher-accuracy point.

| XForms attribute              | XLSForm parameter |
|-------------------------------|-------------------|
| accuracyThreshold             | capture-accuracy  |
| unacceptableAccuracyThreshold | warning-accuracy  |

lognaturel:
Closes #579

This is the approach that matches other parameters currently. I feel more and more of a desire to reduce the spaghetti nature of that part of the code but I don't think we should combine that with adding a new parameter.

I had to combine this with `allow-mock-accuracy` which also applies to `geopoint` questions. So there's some risk there. I added test to verify that the new attributes aren't allowed with trace/shape and that all of the point parameters combine well.

https://github.com/XLSForm/xlsform.github.io/issues/222

### DIFF
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -21,7 +21,6 @@ from pyxform.xls2json_backends import csv_to_dict, xls_to_dict, xlsx_to_dict
 if TYPE_CHECKING:
     from typing import Any, Dict, KeysView, List, Optional
 
-
 SMART_QUOTES = {"\u2018": "'", "\u2019": "'", "\u201c": '"', "\u201d": '"'}
 _MSG_SUPPRESS_SPELLING = (
     " If you do not mean to include a sheet, to suppress this message, "
@@ -1287,9 +1286,16 @@ def workbook_to_json(
 
         if question_type in ["geopoint", "geoshape", "geotrace"]:
             new_dict = row.copy()
-            parameters = get_parameters(
-                row.get("parameters", ""), ["allow-mock-accuracy"]
-            )
+
+            if question_type == "geopoint":
+                parameters = get_parameters(
+                    row.get("parameters", ""),
+                    ["allow-mock-accuracy", "capture-accuracy", "warning-accuracy"],
+                )
+            else:
+                parameters = get_parameters(
+                    row.get("parameters", ""), ["allow-mock-accuracy"]
+                )
 
             if "allow-mock-accuracy" in parameters.keys():
                 if parameters["allow-mock-accuracy"] not in ["true", "false"]:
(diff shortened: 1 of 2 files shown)


lognaturel:
The goal would be to get this in a new pyxform release by Tuesday 2/15!

tests/test_allow_mock_accuracy.py
```diff
@@ -117,3 +117,91 @@ def test_foo_fails(self):
...
+    def test_geopoint_parameters_combine(self):
+        self.assertPyxformXform(
+            name="data",
+            md="""
+            | survey |           |             |          |                                                             |
+            |        | type      | name        | label    | parameters                                                  |
+            |        | geopoint  | geopoint    | Geopoint | warning-accuracy=5.5 capture-accuracy=2 allow-mock-accuracy=true |
+            """,
+            xml__xpath_match=[
+                "/h:html/h:body/x:input[@unacceptableAccuracyThreshold='5.5' and @accuracyThreshold='2' and @ref='/data/geopoint']",
```
lognaturel, line 180: @lindsay-stevens I love writing assertions this way! 🎉

lindsay-stevens: Nice! 🎉

lindsay-stevens (approved):

lindsay-stevens:
Looks good! Just a suggestion about the warning message, but it's not really a blocker.

pyxform/xls2json.py
```diff
@@ -1300,6 +1306,29 @@ def workbook_to_json(
...
+            new_dict["control"] = new_dict.get("control", {})
+            if "capture-accuracy" in parameters.keys():
+                try:
+                    float(parameters["capture-accuracy"])
+                    new_dict["control"].update(
+                        {"accuracyThreshold": parameters["capture-accuracy"]}
+                    )
+                except ValueError:
+                    raise PyXFormError(
+                        "Parameter capture-accuracy must have an integer value"
```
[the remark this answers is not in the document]
lognaturel: 🤦🏻‍♀️🤦🏻‍♀️ Thanks for catching that. I did integer initially and didn't fully fix it. Have force pushed this change.

Issue #593, opened 2022-03-09
Add localized label columns to select_.._from_file


lognaturel:
Follow on to #461 

[Latest spec proposal](https://forum.getodk.org/t/proposal-customize-name-and-label-columns-in-selects-from-external-data-in-xlsform/27265/9?u=ln)

We might want to think about how this harmonizes with internal selects and `jr:itext`: https://github.com/XLSForm/pyxform/pull/577/files#r819787633

Issue #590, opened 2022-03-01
Always generate an instance for a list with columns other than name and label


lognaturel:
#### Software and hardware versions 
All versions

#### Problem description
It can be useful to have a queriable dataset defined within a form. Because instances are only generated for selects with choice filters, sometimes that requires using a `true()` filter to force instance creation. If there are columns other than `name` and `label`, those are only of value if they're in an instance.

#### Steps to reproduce the problem
* Create a form with a `select_one` on a list with additional columns but no choice filter
* Write a query on that list
* See the bizarre Validate error: https://forum.getodk.org/t/referencing-values-in-internal-datasets/36874 (from @jkuester)

#### Expected behavior
If a list has columns other than `name` and `label`, an instance should be generated.

#### Other information 
I'd be in favor of always generating instances: https://github.com/XLSForm/pyxform/issues/203
There are weird problems selecting between instances and static selects already anyway: https://github.com/XLSForm/pyxform/issues/387
We've proposed something like a new `internal-data` type https://github.com/XLSForm/pyxform/issues/176 but this case is different -- there's a select referencing the list but it doesn't have a choice filter

lognaturel:
@lindsay-stevens you've been in related code recently. Is there a clean way to add this special case? Or do you think we should perhaps move forward on always generating instances? The idea that there might be some client somewhere that doesn't support secondary instances doesn't feel like a compelling reason to keep the extra code path.

lindsay-stevens:
From a quick look I think we'd be adding at at least another if condition around about [here in xls2json](https://github.com/XLSForm/pyxform/blob/7b542303112f8f17dcc206362ec2e62d01987c7a/pyxform/xls2json.py#L1168-L1188). Most of the work would probably be in thinking about / checking that this behaves in a reasonable way alongside other overlapping features when the condition is met and not met. To keep a lid on that I'm in favour of always generating instances.

I wonder if the spirit of #203 can apply to #584 i.e. might as well always use itext if select choices are never in the XForm body.

lognaturel:
Closing in favor of https://github.com/XLSForm/pyxform/issues/603 which I think we should do sooner than later.

> might as well always use itext if select choices are never in the XForm body.

I think that's a bit different because I have seen single-language forms (and written them myself) that use `label` directly. It's really not as nice to have to use `jr:choice-name` to get the label through an `itext` ref and there are existing forms that would break on update.

PR #594, opened 2022-03-11
Add support for geojson external files

### LINKED ISSUE #591: If external secondary instance has .geojson extension, use `id` as default value ref and `title` as default label ref
The `select_one_from_file` spec is being expanded to include parsing geojson files. If the file specified for a `select_one_from_file` has extension `geojson`, the value ref should default to `id` and the label ref should default to `title`. These should be possible to override in the regular way (https://github.com/XLSForm/pyxform/pull/577).

The external secondary instance declaration should use the `jr://file` endpoint (just like XML files, unlike CSV files).

More on [ODK forum](https://forum.getodk.org/t/ability-to-load-locations-into-the-geowidget-to-guide-data-collection/27406/12?u=ln).

lognaturel:
Closes #591

Just adding some more conditions! It didn't feel like there were a ton of decision to make. 

Regression risk is isolated to external select types. The test coverage there is very high so it's hard to imagine what could go wrong.

https://github.com/XLSForm/xlsform.github.io/issues/224


lindsay-stevens:
Looks good! I suggest a minor change to use data so the rule is in one place, instead of repeating the conditional. Though it will work either way.

For `constants.py`

```python
EXTERNAL_CHOICES_ITEMSET_REF_LABEL = {".xml": "label", ".csv": "label", ".geojson": "title"}
EXTERNAL_CHOICES_ITEMSET_REF_VALUE = {".xml": "name", ".csv": "name", ".geojson": "id"}
```

then in `question.py`

```python
itemset_label_ref = self.parameters.get(
    "label", 
    EXTERNAL_CHOICES_ITEMSET_REF_VALUE[file_extension]
)
itemset_label_ref = self.parameters.get(
    "label", 
    EXTERNAL_CHOICES_ITEMSET_REF_LABEL[file_extension]
)
```

and in `test_external_instances_for_selects.py`

```python
for ext, xp_city, xp_subs in self.xp_test_args:
    expected_value_ref = EXTERNAL_CHOICES_ITEMSET_REF_VALUE[ext]
    expected_label_ref = EXTERNAL_CHOICES_ITEMSET_REF_LABEL[ext]
```

lindsay-stevens (approved):

lindsay-stevens:
The empty string file extensions seem occur for regular choices e.g. `select_one mylist`. It begs the question as to why the `Question` subclasses don't have the original question type, or some indication that the question is a `..from_file` select, rather than needing to split the `itemset` name to detect them. The following would handle the empty string case without the additional dict item:

```python
itemset_value_ref = self.parameters.get(
    "value", 
    EXTERNAL_CHOICES_ITEMSET_REF_VALUE.get(file_extension, "name")
)
itemset_label_ref = self.parameters.get(
    "label", 
    EXTERNAL_CHOICES_ITEMSET_REF_LABEL.get(file_extension, "label")
)
```

or in SQL style

```python
def coalesce(*args):
    return next((a for a in args if a is not None), None)

itemset_value_ref = coalesce(
    self.parameters.get("value"),
    EXTERNAL_CHOICES_ITEMSET_REF_VALUE.get(file_extension)
    "name"
)
```

#593 would likely touch this code again so whatever we end up with might change again soon. So if the above isn't quite right either, I think we should merge without further changes.

lognaturel:
> why the Question subclasses don't have the original question type

Right, I did wonder that briefly. Let's leave it as-is for now and follow up next time we're in this area. I could imagine some tidying up as part of #590, too.

PR #596, opened 2022-03-14
Use read-only openpyxl representation to reduce memory usage

### LINKED ISSUE #595: For some forms, v1.8 uses an unacceptable amount of memory
A 200kb Excel doc was measured using 1.5GB of memory.

The most likely culprit is `openpyxl`, introduced in #575 by @sheppard

From https://openpyxl.readthedocs.io/en/stable/performance.html:
> Memory use is fairly high in comparison with other libraries and applications and is approximately 50 times the original file size, e.g. 2.5 GB for a 50 MB Excel file.

What we're seeing is higher by two orders of magnitude.

- [ ] Look at #571 for possible memory issues
- [x] Try openpyxl [read only mode](https://openpyxl.readthedocs.io/en/stable/optimized.html#read-only-mode.)

lognaturel:
Closes #595

Even though #595 is not critical, this seems like a good change. We don't ever modify an Excel file so there's no point in opening it for writing.

There are slight differences between the writeable and read-only workbook representations I had to handle. The biggest decision I made was not to worry about the difference in dealing with trailing empty rows. That is, if there are empty rows in between some rows with contents, the two representations are the same. If there are empty rows after which there are no more rows with content, the read-only representation includes those empty rows whereas the writeable one does not. One of the equivalency tests between XLS and XLSX picked up on that difference. I tried dropping all empty rows but this would change the output for a rogue `table-list` feature that uses row number to name nodes (!). I verified that empty rows don't affect the output and decided to change the test XLSX instead of changing the code. That seems like the less risky option. I changed the XLSX and XLS files the same way for the `group` and `specify_other` forms. I deleted the rows with formatting but no contents.

There could be some side effects to the different handling of empty rows and columns. I am not thinking of any but that would be good to get a second opinion on.


pyxform/utils.py
```diff
@@ -190,13 +190,12 @@ def xls_sheet_to_csv(workbook_path, csv_path, sheet_name):
 
 
 def xlsx_sheet_to_csv(workbook_path, csv_path, sheet_name):
-    wb = openpyxl.open(workbook_path)
+    wb = openpyxl.open(workbook_path, read_only=True, data_only=True)
     try:
         sheet = wb.get_sheet_by_name(sheet_name)
     except KeyError:
         return False
-    if sheet.max_row < 2:
```
lognaturel, line 198: I believe this was a no-op. I added a test for a case with just the header at https://github.com/XLSForm/pyxform/pull/596/files#diff-56abab35a27949c49dd2950913ccd05332e683da4a81894108a4df6da67fb653 and https://github.com/XLSForm/pyxform/pull/596/files#diff-e5efd53ccd5cb96180d000bb4cd9157844e81d6b367d1253a2d8f23d7341ddcfR619 tests the empty sheet case

tests/test_external_instances_for_selects.py
```diff
@@ -433,7 +448,7 @@ def test_external_other_extension_instances(self):
             |        | select_multiple_from_file neighbourhoods.pdf | neighbourhoods | Neighbourhoods |
             """,  # noqa
             errored=True,
-            error_contains=["should be a choices sheet in this xlsform"],
+            error__contains=["should be a choices sheet in this xlsform"],
```
lognaturel, line 451: A not so great finding: `error_contains` passes no matter what the expected error message is.

lognaturel:
I'm planning to do a `pyxform` release first thing tomorrow (GMT-7) and would like to include this if possible. @lindsay-stevens would be fantastic if you have time for a quick review. If you don't have time or think it's riskier than I believe, let me know and we can release without!

lognaturel:
Sadly I don't understand why Windows tests are failing. 

```
ERROR: Should find that XLSForm conversion produces itemsets.csv from external_choices.
  File "D:\a\pyxform\pyxform\tests\test_external_instances_for_selects.py", line 418, in test_itemset_csv_generated_from_external_choices
    self.assertEqual('"suburb","Footscray","vic","melbourne"\n', rows[-1])
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\contextlib.py", line 119, in __exit__
    next(self.gen)
  File "D:\a\pyxform\pyxform\tests\utils.py", line 74, in get_temp_dir
    shutil.rmtree(temp_dir)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\shutil.py", line 516, in rmtree
    return _rmtree_unsafe(path, onerror)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\shutil.py", line 400, in _rmtree_unsafe
    onerror(os.unlink, fullname, sys.exc_info())
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\shutil.py", line 398, in _rmtree_unsafe
    os.unlink(fullname)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp9ms014r1\\select_one_external.xlsx'
```

lindsay-stevens (commented):
I think there should be a test with the problem case to demonstrates the memory consumption improvement. For example 1) generate the XLSX, 2) record resident memory (see below), 3) run xls2xform_convert, 4) assert resident memory not grown by some amount or percentage.

```python
# Not perfect but good indication
import resource
resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
```

lindsay-stevens:
> Sadly I don't understand why Windows tests are failing.

I haven't tested but it could be that the workbook is still open, as mentioned here: https://openpyxl.readthedocs.io/en/latest/optimized.html

image

pyxform/xls2json_backends.py
```diff
@@ -204,12 +207,15 @@ def xlsx_to_dict_normal_sheet(sheet):
...
+                    value = row[column].value
+                    if isinstance(value, str):
+                        value = value.strip()
 
-                if not is_empty(value):
-                    row_dict[key] = xlsx_value_to_str(value)
+                    if not is_empty(value):
+                        row_dict[key] = xlsx_value_to_str(value)
+                except IndexError:
+                    pass  # rows may not have values for every column
```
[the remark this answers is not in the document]
lognaturel: It depends on the structure of the XLSX document. I assume it's more likely to happen with machine-written files (e.g. the one from `test_itemset_csv_generated_from_external_choices`). I'll add a test explicitly targeting it.

pyxform/xls2json_backends.py
```diff
@@ -175,27 +175,30 @@ def xlsx_to_dict(path_or_file):
...
-            if is_empty(column_header):
-                # Preserve column order (will filter later)
-                column_header_list.append(None)
-            else:
-                if column_header in column_header_list:
-                    raise PyXFormError("Duplicate column header: %s" % column_header)
-                # strip whitespaces from the header
-                clean_header = re.sub(r"( )+", " ", column_header.strip())
-                column_header_list.append(clean_header)
+        try:
```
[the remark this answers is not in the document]
lognaturel: Thank you, that is much better! 🎉

tests/test_external_instances_for_selects.py
```diff
@@ -402,9 +402,6 @@ def test_itemset_csv_generated_from_external_choices(self):
                 xls2xform_convert(
                     xlsform_path=wb_path,
                     xform_path=get_xml_path(wb_path),
-                    validate=True,
-                    pretty_print=False,
-                    enketo=False,
```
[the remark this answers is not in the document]
lognaturel: I mostly wanted to make sure to remove the `validate=True` because it slows the tests a lot. We always run with Validate before release. It seemed to me the others didn't really matter but I'm happy to bring them back.

tests/test_external_instances_for_selects.py
lognaturel: Not a big priority but agreed an issue would be good! 🙏

pyxform/utils.py
lognaturel: I think you were saying I shouldn't have created the `empty_sheets.xlsx` file and instead built it programmatically. I've removed that test because it actually wasn't relevant. I ended up thinking this was about any empty sheet but it's only for `select_one_external` and the `external_choices` sheet.

This code is never reached if the `external_choices` sheet is either empty or has only a header. So I think it's ok not to build a real XLSX file to check those two cases. It seems very strange to me that this re-opens the XLSX again for parsing but I didn't look any deeper.

lognaturel: Ah, right, but the empty sheet case is not supported by md to json. So I'll have the two cases use a real xlsx.

tests/test_external_instances_for_selects.py
```diff
@@ -420,6 +408,57 @@ def test_itemset_csv_generated_from_external_choices(self):
             # Should have excluded column with "empty header" in the last row.
             self.assertEqual('"suburb","Footscray","vic","melbourne"\n', rows[-1])
 
+    def test_empty_external_choices__errors(self):
```
lognaturel, line 411: This test and the next cover what the removed code [here](https://github.com/XLSForm/pyxform/pull/596/files#r826414629) was trying to address.

tests/test_xls2json.py
```diff
@@ -601,6 +603,26 @@ def test_workbook_to_json__optional_sheets_ok(self):
             warnings_count=0,
         )
 
+    def test_xls2xform_convert__e2e_row_with_no_column_value(self):
```
lognaturel, line 606: If you remove the code you commented on [here](https://github.com/XLSForm/pyxform/pull/596/files#r826597130), this test should fail.

tests/test_external_instances_for_selects.py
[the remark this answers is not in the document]
lognaturel: > We always run with Validate before release.

Ah, except that these aren't pyxform test cases so they won't be tested against Validate. I still think that's better because it runs the tests faster. I also happen to know that the entire "fast external itemset" feature is ignored by JavaRosa/Validate.

tests/test_xls2json.py
```diff
@@ -1,11 +1,17 @@
 import os
+import resource
+
+import psutil
```
lognaturel, line 4: `resource` is *NIX-only.

lognaturel:
> it could be that the workbook is still open

Thanks so much for helping me read the docs. That seemed like a likely culprit but either I'm closing in the wrong spot or something else is going on. Unfortunately I'm out of time for this tonight.

Having chatted with @yanokwa and Team Central, I think this fix should go in the pyxform release so I haven't released yet. We don't know how the extra columns get generated and it seems like the kind of thing that others would run into so best to patch it now.

tests/test_xls2json.py
```diff
@@ -617,6 +642,20 @@ def test_xls2xform_convert__e2e_with_settings_misspelling(self):
...
+    ):
+        """Degenerate form with many blank columns"""
+        process = psutil.Process(os.getpid())
+        pre_mem = process.memory_info().rss
+        xls2xform_convert(
+            xlsform_path=os.path.join(example_xls.PATH, "extra_columns.xlsx"),
+            xform_path=os.path.join(test_output.PATH, "extra_columns.xml"),
+        )
+        post_mem = process.memory_info().rss
+        # in v1.8.0, memory usage grew by over 16x
```
lognaturel, line 656: I have run this test in v1.8.0 to get that value.

lindsay-stevens:
I came up with a workaround for the Windows test failures, [PR here](https://github.com/lognaturel/pyxform/pull/1) to your branch. Otherwise, this PR seems good to go.

lindsay-stevens (approved):

lognaturel:
Nice working with you on this, @lindsay-stevens 🤝

lognaturel:
From @lindsay-stevens on the PR to my fork [here](https://github.com/lognaturel/pyxform/pull/1#issuecomment-1070441986):

> Just a note for posterity about my guess for antivirus. Main clue was the Windows "Resource Monitor" tab for Disk activity. This activity list shows active file paths and the PID and image name doing read/write on them. When running the tests, the file was being read by a process with image name "System" and description "NT Kernel & System". Not sure what else the system would be doing with files except to scan them - but whatever it's doing we can't exactly terminate the system process 💥

> The same thing happened when I changed the temp root to pyxform/tests/test_output so it doesn't seem to be related to using the system/user temporary directory.

PR #606, opened 2022-04-14
Set the parameters field/column default type to type dict

ukanga:
Closes https://github.com/XLSForm/pyxform/issues/605

#### Why is this the best possible solution? Were any other approaches considered?

Minimal code change, works. 
See additional context and information on the issue https://github.com/XLSForm/pyxform/issues/605.

#### What are the regression risks?

None

#### Does this change require updates to documentation? If so, please file an issue [here](https://github.com/XLSForm/xlsform.github.io) and include the link below.

No

#### Before submitting this PR, please make sure you have:
- [x] included test cases for core behaviour and edge cases in `tests`
- [x] run `nosetests` and verified all tests pass
- [x] run `black pyxform tests` to format code
- [x] verified that any code or assets from external sources are properly credited in comments

PR #607, opened 2022-04-19
fix: survey json->xml round trip not working for some question types

### LINKED ISSUE #605: Change the parameters field/column default type to type dict
pyxform v1.9

#### Problem description

When you process an XLSForm, in whatever format (markdown, spreadsheet file, e.t.c.), and get the JSON survey form via `survey.to_json()` command. If you create a new survey object using this JSON representation, which is possible via the `SurveyElementBuilder().create_survey_element_from_json(json_survey_dict)`, the `parameters` field for a select one question that results in itemsets will be set to an empty string due to the `str` default type of the parameters field [here ](https://github.com/XLSForm/pyxform/blob/b0ad3a7a8a8a2bf9a3fe262b73208c372eb05c84/pyxform/survey_element.py#L56). This is the case for all XLSForms that do not have a `parameters` values set for any field in the form. As such, the empty str default will be set [here](https://github.com/XLSForm/pyxform/blob/b0ad3a7a8a8a2bf9a3fe262b73208c372eb05c84/pyxform/survey_element.py#L106) and as a result fail with the exception `AttributeError: 'str' object has no attribute 'get'` when generating the XForm XML via `SurveyElementBuilder().create_survey_element_from_json(survey.to_json()).to_xml()`.

```python
 ...
    json_survey = xls2json.parse_file_to_json(xlsform_path, warnings=warnings)
    survey = builder.create_survey_element_from_dict(json_survey)
    SurveyElementBuilder().create_survey_element_from_json(survey.to_json()).to_xml()
  File "/.../pyxform/pyxform/survey.py", line 1117, in to_xml
    self.print_xform_to_file(
  File "/.../pyxform/pyxform/survey.py", line 1082, in print_xform_to_file
    raise error
  File "/.../pyxform/pyxform/survey.py", line 1076, in print_xform_to_file
    file_obj.write(self._to_pretty_xml())
  File "/.../pyxform/pyxform/survey.py", line 866, in _to_pretty_xml
    xml_with_linebreaks = self.xml().toprettyxml(indent="  ")
  File "/.../code/pyxform/pyxform/survey.py", line 254, in xml
    node("h:body", *self.xml_control(), **body_kwargs),
  File "/.../pyxform/pyxform/section.py", line 91, in xml_control
    control = e.xml_control()
  File "/.../pyxform/pyxform/question.py", line 55, in xml_control
    xml_node = self.build_xml()
  File "/.../pyxform/pyxform/question.py", line 210, in build_xml
    itemset_value_ref = self.parameters.get(
AttributeError: 'str' object has no attribute 'get'
```

```python
  # create a survey object from an XLSForm File
  json_survey = xls2json.parse_file_to_json(xlsform_path, warnings=warnings)
  survey = builder.create_survey_element_from_dict(json_survey)
  # test that you can successfully create the XForm XML of the survey object
  survey.to_xml()   # will not raise an exception
 
  # creating survey from the JSON for the same survey even with no modifications will raise the AttributeError exception
  SurveyElementBuilder().create_survey_element_from_json(survey.to_json()).to_xml()  # AttributeError exception will be raised.

```

#### Expected behaviour

The JSON Survey representation should be able to create a Survey object that can generate the XForm XML without any exception being thrown. 

#### Other information 

I noticed that, when processing the XLSForm, or its JSON dict equivalent which is processed by [workbook_to_json()](https://github.com/XLSForm/pyxform/blob/b0ad3a7a8a8a2bf9a3fe262b73208c372eb05c84/pyxform/xls2json.py#L347) always results in the `parameters` field that has a dict empty or with key-value parameters. Hence the recommended fix is making the default value of type `dict` which will not result in an exception. There seem to be no other scenarios, as far as I can tell, where the `parameters` field value is a string when dealing with a `SurveyElement`.

> Why would you want to recreate a Survey object from the "XForm JSON representation"?
 
It is a cheap way to add some extra fields or metadata like setting the XForm version post processing the XLSForm file among other tasks an application could do for the user without the user making those changes themselves and if those changes are relevant to the application and not necessarily the user. Any other option better than this approach for this use case?

lindsay-stevens:
Closes #605

- details of issue noted in new test under js2x_test_import_from_json.py
- xls2json updated to make it easier to be confident that parameters will be a dict (vs. none or a string).
  - moved off params processing to parse/validate step validators package to be consistent with other utils of this nature.
  - updated error message to show all invalid params at once, instead of one at a time (which may have resulted in users to having convert a few times to get a valid params set).

PR #606 also fixes this but I wanted to be a bit more sure that xls2json is really passing in a dict in all cases. Also this PR has a separate test which isolates the issue to the scenario where a question that accepts parameters doesn't have parameters, and that survey is passed in as json and then converted to XML.

Hopefully none! The error message for invalid params is updated, maybe some users expect the old message. But I think it is more user friendly to say what all the issues are at once, if possible.

No

### DIFF
--- a/pyxform/survey_element.py
+++ b/pyxform/survey_element.py
@@ -53,7 +53,7 @@ class SurveyElement(dict):
         "default": str,
         "type": str,
         "appearance": str,
-        "parameters": str,
+        "parameters": dict,
         "intent": str,
         "jr:count": str,
         "bind": dict,
@@ -128,7 +128,7 @@ class SurveyElement(dict):
         else:
             self.add_child(children)
 
-    binding_conversions = {
+    BINDING_CONVERSIONS = {
         "yes": "true()",
         "Yes": "true()",
         "YES": "true()",
@@ -143,16 +143,16 @@ class SurveyElement(dict):
         "FALSE": "false()",
     }
 
-    CONVERTIBLE_BIND_ATTRIBUTES = [
+    CONVERTIBLE_BIND_ATTRIBUTES = (
         "readonly",
         "required",
         "relevant",
         "constraint",
         "calculate",
-    ]
+    )
 
     # Supported media types for attaching to questions
-    SUPPORTED_MEDIA = ["image", "audio", "video"]
+    SUPPORTED_MEDIA = ("image", "audio", "video")
 
     def validate(self):
         if not is_valid_xml_tag(self.name):
@@ -457,10 +457,10 @@ class SurveyElement(dict):
                 # the xls2json side.
                 if (
                     hashable(v)
-                    and v in self.binding_conversions
+                    and v in self.BINDING_CONVERSIONS
                     and k in self.CONVERTIBLE_BIND_ATTRIBUTES
                 ):
-                    v = self.binding_conversions[v]
+                    v = self.BINDING_CONVERSIONS[v]
                 if k == "jr:constraintMsg" and (
                     type(v) is dict or re.search(BRACKETED_TAG_REGEX, v)
                 ):
--- a/pyxform/validators/pyxform/parameters_generic.py
+++ b/pyxform/validators/pyxform/parameters_generic.py
@@ -1,0 +1,43 @@
+from typing import Any, Dict, Sequence
+
+from pyxform.errors import PyXFormError
+
+PARAMETERS_TYPE = Dict[str, Any]
+
+
+def parse(raw_parameters: str) -> PARAMETERS_TYPE:
+    parts = raw_parameters.split(";")
+    if len(parts) == 1:
+        parts = raw_parameters.split(",")
+    if len(parts) == 1:
+        parts = raw_parameters.split()
+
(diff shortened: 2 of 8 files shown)


lognaturel (approved):
Looks good to me. Nice to see some of the `parameters` functionality start to be split off. 

Going to wait to get feedback from @ukanga

ukanga (approved):

ukanga:
Looks good, you took a step further than I did. I prefer this approach as well. Let's tag a release, please.

lognaturel:
@ukanga Forgot the release is not just tag and go yet! Doesn't look like I'll get to it today. Fine by me if you want to merge your own PR and release. Or if you get the PR ready I can tag and release tomorrow.

I've gone back and forth on whether this should be v1.10.0 or v1.9.1. I could go either way. I was leaning towards v1.10.0 because the detection of dynamic defaults is a change in behavior, not just a fix.

Issue #611, opened 2022-06-09
list index out of range when trying to convert file saved with LibreOffice


lognaturel:
[ODK forum thread](https://forum.getodk.org/t/error-list-index-out-of-range/37931).

Manually copying the contents of the file and saving them in a new document worked.

Related to https://github.com/XLSForm/pyxform/issues/604 in that these document compatibility issues come from `openpyxl`.

lindsay-stevens:
From the original thread, [this comment](https://forum.getodk.org/t/error-list-index-out-of-range/37931/2) has an example file [PRUEBA.xlsx](https://github.com/XLSForm/pyxform/files/8975042/PRUEBA.xlsx) which seems corrupted somehow. I can open it with Excel (2010) or LibreOffice (7.2.7.2) but has excessive style / format data. It is 674KB on disk, and approx. 85% of that is from the `./xl/styles.xml` document within the xlsx zip file. After opening the file, there are dozens of hyperlink (`Hipervínculo`) custom formats. Opening the file with openpyxl (with `read_only` and `data_only` modes) takes about 45 seconds on my machine. For this kind of file, perhaps pyxform could have a file read timeout kwarg (in `xls2xform_convert`) to optionally limit resource usage when pyxform is used in a server / service context?

lindsay-stevens:
From the original thread, [this comment](https://forum.getodk.org/t/error-list-index-out-of-range/37931) also has an example file [PRUEBA.xlsx](https://github.com/XLSForm/pyxform/files/8975253/PRUEBA.xlsx) which also seems corrupted. I can open it with Excel (2010) or LibreOffice (7.2.7.2) but has excessive style / format data, as described above. The file won't open with openpyxl, instead an error relating to the workbook style data is thrown, as copied below. For this kind of file, perhaps pyxform could catch the error and in the warning, suggest that the user re-save the file with Excel or try copying the XLSForm data into a new workbook file? Alternatively, perhaps this is a known issue for openpyxl or could be fixed upstream there.

<details><summary>Error traceback</summary>

```
Error
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/local/lib/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/usr/local/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
[...]

PR #612, opened 2022-06-10
avoid processing vast empty areas of buggy / strange workbooks

### LINKED ISSUE #604: Form conversion is slow on form with lots of blank rows and columns
pyxform v1.9

#### Problem description
[UCL_Biomass_Plot_Form.xlsx](https://github.com/XLSForm/pyxform/files/8475575/UCL_Biomass_Plot_Form.xlsx) when converted with pyxform-http (running in Docker on my computer) takes 27 seconds. 

The form has lots of blank columns and rows and seems to be corrupt in some way. 

I re-saved the form as XLSX in Excel 16.59 for Mac and that seems to have cleaned things up. Conversion speed went from to 27 seconds to .43 seconds.

#### Other information 
https://forum.getodk.org/t/the-xlsform-conversion-service-could-not-be-contacted/37315/ is the forum thread with the original issue.

This sounds a lot like https://github.com/XLSForm/pyxform/pull/596, but I have confirmed I'm running 1.9. It's hard to verify directly with xls2xform because my computer is really fast :(

https://github.com/getodk/pyxform-http/blob/master/app/main.py#L40 is the specific pyxform-http function call.

My guess is that the form in question has the wrong dimensions. This script points to that.
```
from openpyxl import load_workbook
wb = load_workbook(filename='UCL_Biomass_Plot_Form.xlsx', read_only=True)
ws = wb['survey']
print(ws.calculate_dimension()) # A1:AMJ1048576
ws.reset_dimensions()
print(ws.calculate_dimension(force=True)) # A1:AMJ91
```
Maybe if we see a survey sheet with more than 50 columns (2x what we'd expect) or 5000 rows, we should reset the dimensions?

https://openpyxl.readthedocs.io/en/2.5.14/optimized.html#worksheet-dimensions

lindsay-stevens:
- somehow it happens that spreadsheets sometimes think they have data for 100's or 1000's of rows/columns even though all cells are empty.
- xls / xlsx processing refactored to stop processing after 20 adjacent empty columns (when getting the headers) or 20 adjacent empty rows (when reading rows). That is to say, the practice of inserting a couple of empty rows or columns for readability / formatting purposes won't ruin the form. It'd only be a problem if users insert patches of 20 or more empty columns / rows amongst the form design area.
- added tests for xls / xlsx to show bad input processed as expected.
- the change to not output empty rows seems to have broken the "flatxlsformtest" case which needs further investigation - seems to just be a question numbering issue and different itext lang order.

Closes #604 
Might also address #611 but haven't checked yet.

It's not clear how / why these buggy spreadsheets come to be. From the associated tickets / forum threads it seems like re-saving with Excel may or may not work. It's possible there is a bug in xlrd or openpyxl to blame, or perhaps these libraries could handle these situations better, but in either case it may be hard to get an upstream fix considering the poor reproduce-ability. The most practical solution seemed to be to have pyxform only process the workbook sheets for as long as there seems to be data in the columns / rows.

As mentioned above there's a broken test which seems at least in part to involve a mismatch from using the row number for a automatically generated question name. But if that's no good then we can still output the empty rows.

It doesn't seem so at this stage.


lognaturel:
Thanks, @lindsay-stevens! I think this is a reasonable approach. Did you also consider @yanokwa's suggestion at the bottom of https://github.com/XLSForm/pyxform/issues/604 to reset the dimensions above a certain reported column count?

lindsay-stevens:
@lognaturel thanks for taking a look! The [source](https://foss.heptapod.net/openpyxl/openpyxl/-/blob/3.0.10/openpyxl/worksheet/_read_only.py#L143-157) for calculate_dimensions iterates through rows until a single break in truthiness, and uses the largest column index from cells indexed in those rows. A cell object can represent formatting info only (no value data). In the example UCL file the re-calculated column dimension at column AMJ (1024th) is from a break in the grey background colour formatting which then extends on forever from column AMK. The relevant XLSForm data goes to column P (16th) only.

So if a user inserts an empty row or column for formatting, but doesn't colour or style it, and the worksheet dimensions are still saved incorrectly somehow, then the recalc might cut off some data. In other words, it'd result in an extra iteration of the worksheet data (2nd time is to read values), and in doing so might iterate too far or not enough. The approach in this PR also guards against accidentally adding irrelevant data or styling way off in at the edges of the worksheet.

Maybe not as much of an issue for XLS (max cols 256 x rows 65536) but there doesn't seem to be an equivalent method in `xlrd`.

yanokwa:
It might be nice to add a warning when do this? Just in case someone has some number of empty rows/columns?

pyxform/xls2json_backends.py
```diff
@@ -37,6 +40,71 @@ def _list_to_dict_list(list_items):
...
+            column_header_list.append(None)
+            if last_col_empty:
+                adjacent_empty_cols += 1
+                # After a run of empty cols, assume we've reached the end of the data.
+                if 20 <= adjacent_empty_cols:
+                    break
+            else:
+                last_col_empty = True
+                adjacent_empty_cols = 0
+        else:
```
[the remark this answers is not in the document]
lognaturel: In fact, can you just keep a single counter instead of also having a boolean? So `empty_adjacent_cols` gets incremented by 1 as you see empty columns and set to 0 as you see non-empty ones.

lognaturel:
Thanks for the additional explanation, @lindsay-stevens, the approach sounds good. I don't feel a strong need for a warning and I imagine it'd be a fair amount of plumbing so doesn't seem worth it.

I'm not sure about the logic and have commented inline. Otherwise the rest of the implementation and the tests look good.

I verified whether it might address #611 and unfortunately it does not. I verified that it does address all 3 forms from the thread in #604. 👍

If at all possible, it would be great to get this released by the end of the week!

pyxform/xls2json_backends.py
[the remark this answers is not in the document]
lindsay-stevens: Thanks! I was sitting on some changes to fix the total vs. adjacent issue but have also included the `empty_adjacent_cols` simplification which makes it a lot cleaner.

lindsay-stevens:
Thanks for the review @lognaturel. I've fixed up the algo so this could be merged. I'll have a look at #611 and open a docs PR for the blank col/row processing behaviour.

pyxform/xls2json_backends.py
```diff
@@ -37,6 +40,75 @@ def _list_to_dict_list(list_items):
...
+def get_excel_column_headers(first_row: Iterator[Optional[str]]) -> List[Optional[str]]:
+    """Get column headers from the first row; stop if there's a run of empty columns."""
+    column_header_list = list()
+    adjacent_empty_cols = 0
+    last_col_empty = False
+    for column_header in first_row:
+        if is_empty(column_header):
+            # Preserve column order (will filter later)
+            column_header_list.append(None)
+            if last_col_empty:
```
lognaturel, line 52: Rows looks good but this still doesn’t look quite right! I don’t think you need last_col_empty at all. You can always increment in this branch and set to 0 in the other.

pyxform/xls2json_backends.py
```diff
@@ -37,6 +40,75 @@ def _list_to_dict_list(list_items):
...
+                trim_trailing_empty_rows = True
+                break
+        else:
+            adjacent_empty_rows = 0
+
+        # There may be some empty rows amongst the XLSForm data. These are included
+        # so that any warning messages that mention row numbers are accurate.
+        result_rows.append(row_dict)
+
+    if trim_trailing_empty_rows:
```
lognaturel, line 107: I don’t think you need trim_trailing_empty_rows either. Instead, it should always be appropriate to trim adjacent_empty_rows. That will also handle the case in which there are eg 2 empty columns before the end.

pyxform/xls2json_backends.py
lindsay-stevens: No worries, updated both cols/rows algos to just use `adjacent_empty_*` instead of flag variables.

pyxform/xls2json_backends.py
lindsay-stevens: Updated to always trim cols/rows, based on `adjacent_empty_*`.

pyxform/xls2json_backends.py
```diff
@@ -37,6 +40,78 @@ def _list_to_dict_list(list_items):
...
+    """Get column headers from the first row; stop if there's a run of empty columns."""
+    max_adjacent_empty = 20
+    column_header_list = list()
+    adjacent_empty_cols = 0
+    for column_header in first_row:
+        if is_empty(column_header):
+            # Preserve column order (will filter later)
+            column_header_list.append(None)
+            # After a run of empty cols, assume we've reached the end of the data.
+            if max_adjacent_empty < adjacent_empty_cols:
```
lognaturel, line 63: OBOB! This will stop after 21 empty columns are identified, the increment should really be before the test. But since 20 is arbitrary anyway, I suppose it doesn't really matter.

pyxform/xls2json_backends.py
```diff
@@ -37,6 +40,78 @@ def _list_to_dict_list(list_items):
...
+            try:
+                cell = row[col_n]
+                if not is_empty(cell.value):
+                    row_dict[key] = cell_func(cell, row_n, key)
+            except IndexError:
+                pass  # rows may not have values for every column
+
+        if 0 == len(row_dict):
+            # After a run of empty rows, assume we've reached the end of the data.
+            if max_adjacent_empty < adjacent_empty_rows:
```
lognaturel, line 102: Same as above, it will only stop after reaching 21 empty rows.

lognaturel (approved):
Thanks! There's a slightly confusing off-by-one-from-intent, I think, but it doesn't affect users so I'm fine to leave it and will let you decide whether it bothers you enough to patch it with a future PR now that I've mentioned it! 😄

pyxform/xls2json_backends.py
lindsay-stevens: Well yes now it has been bothering me! 😄 Will add to a later PR. Thanks

lognaturel: 😁

PR #621, opened 2022-11-03
More likely to be more empty rows than empty columns

### LINKED ISSUE #620: Increase max_adjacent_empty to prevent form breakage
Sometimes, when I'm iterating on forms, I like to use a lot of blank rows to mark my place. And because of https://github.com/XLSForm/pyxform/pull/612, my form doesn't get converted by pyxforms.

max_adjacent_empty of 20 is fine for columns, but for rows, I think we should do 100. 

If it's easy, maybe we should also warn that we stopped processing the form because of lots of empty rows and columns.

yanokwa:
Closes #620

Smallest change possible.

None. We are increasing a limit.

Nope.


lindsay-stevens:
Seems reasonable to me. Let them be S P A C E D!

lindsay-stevens (approved):

lognaturel:
S

P

A

C

E

D

?

😅

PR #635, opened 2023-01-13
Add support for big-image

### LINKED ISSUE #34: big-image and media::big-image column heading for itext tag is not available in XLSForm

lognaturel:
Closes #34

I tried to follow the implementation for other media types exactly. I didn't consider any alternatives.

I considered not adding `big-image` to all of the translation tests but I decided it would feel more consistent to do so.

This is an additive change that doesn't touch much existing logic so regression risks are low. There's some possibility of regression with other media types but they're well tested.

Yes, https://github.com/XLSForm/xlsform.github.io/issues/234

### DIFF
--- a/pyxform/survey_element.py
+++ b/pyxform/survey_element.py
@@ -434,8 +434,10 @@ class SurveyElement(dict):
             raise PyXFormError(msg)
 
         # big-image must combine with image
-        if not ("image" in self.media) and "big-image" in self.media:
-            raise PyXFormError("You must specify an image in order to use big-image.")
+        if "image" not in self.media and "big-image" in self.media:
+            raise PyXFormError(
+                "To use big-image, you must also specify an image for the survey element named {self.name}."
+            )
 
         return result
 
--- a/pyxform/survey_element.py
+++ b/pyxform/survey_element.py
@@ -428,9 +428,14 @@ class SurveyElement(dict):
         msg = "The survey element named '%s' " "has no label or hint." % self.name
         if len(result) == 0:
             raise PyXFormError(msg)
+
         # Guidance hint alone is not OK since they may be hidden by default.
         if not any((self.label, self.media, self.hint)) and self.guidance_hint:
             raise PyXFormError(msg)
+
+        # big-image must combine with image
+        if not ("image" in self.media) and "big-image" in self.media:
+            raise PyXFormError("You must specify an image in order to use big-image.")
 
         return result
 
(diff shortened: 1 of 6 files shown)


pyxform/survey_element.py
```diff
@@ -428,10 +428,15 @@ def xml_label_and_hint(self) -> "List[DetachableElement]":
         msg = "The survey element named '%s' " "has no label or hint." % self.name
         if len(result) == 0:
             raise PyXFormError(msg)
+
         # Guidance hint alone is not OK since they may be hidden by default.
         if not any((self.label, self.media, self.hint)) and self.guidance_hint:
             raise PyXFormError(msg)
 
+        # big-image must combine with image
+        if not ("image" in self.media) and "big-image" in self.media:
```
lindsay-stevens, line 437: Suggest `if "image" not in self.media and "big-image" in self.media`

pyxform/survey_element.py
```diff
@@ -428,10 +428,15 @@ def xml_label_and_hint(self) -> "List[DetachableElement]":
...
+            raise PyXFormError("You must specify an image in order to use big-image.")
```
lindsay-stevens, line 438: Suggest `raise PyXFormError(f"In order to use big-image, you must specify an image for the survey element named {self.name}`

lindsay-stevens (approved):
LGTM. 2 minor suggestions around validation/message.

lindsay-stevens:
One other thought, is it fine that the translation test `test_missing_translation__one_lang_all_cols__warn__default` passes, seeing as the question type is `note` and there's a `big-image` specified?

lognaturel:
> the question type is note and there's a big-image specified

Yes, this is fine and good! `big-image` can be used to provide a bigger image either for a question label or for a select choice. A `note` is a read-only text question and can have the same label types as other question types.

lognaturel:
For posterity/anyone watching who may have opinions, I want to acknowledge that this is the first column with a - in its name, all others use underscore.

I went with it because it matches the underlying XForm spec and is more of a type/variant than the other column headers.

Issue #632, opened 2022-12-07
Incorporate improvements from review of entities implementation


lognaturel:
Great feedback from @lindsay-stevens at https://github.com/XLSForm/pyxform/pull/624#issuecomment-1328804968

lindsay-stevens:
The referenced items are either already implemented or now out of date.

PR #636, opened 2023-01-17
Show error when there are `save_to`s in a repeat, simplify entities tests

### LINKED ISSUE #633: Show error when there are `save_to`s in a repeat
When there are `save_to`s in a repeat and an entity declared in the `entities` sheet, `pyxform` creates an entity declaration at the top level but it should give an unsupported error for now.

lognaturel:
Closes #633
Addresses part of #632

I decided to clean up entities tests here to make it easier to identify what test coverage there currently is.

It took me a little bit to decide how to check whether a row is in a repeat. There was no such functionality at that level (e.g. survey_element has it) and it's only needed in one place so I did it with an inline expression. I think it's easy enough to read and to test from the outside.

This doesn't touch any existing logic. The worst case is probably that if my "in repeat" check is wrong we could be rejecting forms incorrectly.

No.


lognaturel:
> maybe if there was a group inside the repeat? Or the bound item wasn't the first in the repeat?

Good cases to consider. I tried those as an extra sanity check and confirmed they're ok but don't think they're worth adding to the test suite.

lognaturel:
Thank you so much, @lindsay-stevens 🎉

Issue #601, opened 2022-04-01
Forms with a media column result in unhelpful error message.


yanokwa:
If you have a form with a `media` column (not `media::image`) and something in that column, pyxform gives a very unhelpful error.
> Error: 'str' object has no attribute 'items'

See [test.xlsx](https://github.com/XLSForm/pyxform/files/8401168/test.xlsx) for an example.

It would be ideal if we warn that the column heading is likely to be incorrect. Maybe we build on the sheet name checking at #544?

lognaturel:
`media` is not a known column name. It used to just be ignored like other unknown columns but since the translation checks were added, there's an assumption that `media` will be a dictionary.

I wanted to add a helpful error message but then I realized that we don't currently have an example of rejecting based on an issue in the header.

In the mean time, I propose restoring the prior behavior which is to ignore the unknown column name.

lognaturel:
I think https://github.com/getodk/docs/issues/1546 will sufficiently address this.

PR #638, opened 2023-01-17
Ignore media column if it is present

lognaturel:
Makes marginal progress towards https://github.com/XLSForm/pyxform/issues/601

It's not a great solution towards the original issue but I think it's an improvement. We don't have an example of header validation and I don't have time to design a good solution right now.

A user could presumably be expecting an exception there now. That seems fine to break.

No.

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -774,7 +774,7 @@ class Survey(Section):
             if parent and not parent.get("choice_filter"):
                 translation_key = survey_element.get_xpath() + ":label"
                 media_dict = survey_element.get("media")
-                if type(media_dict) is dict:
+                if isinstance(media_dict, dict):
                     _set_up_media_translations(media_dict, translation_key)
 
     def itext(self):
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -774,7 +774,8 @@ class Survey(Section):
             if parent and not parent.get("choice_filter"):
                 translation_key = survey_element.get_xpath() + ":label"
                 media_dict = survey_element.get("media")
-                _set_up_media_translations(media_dict, translation_key)
+                if type(media_dict) is dict:
+                    _set_up_media_translations(media_dict, translation_key)
 
     def itext(self):
         """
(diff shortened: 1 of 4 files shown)


pyxform/survey.py
```diff
@@ -774,7 +774,8 @@ def _set_up_media_translations(media_dict, translation_key):
             if parent and not parent.get("choice_filter"):
                 translation_key = survey_element.get_xpath() + ":label"
                 media_dict = survey_element.get("media")
-                _set_up_media_translations(media_dict, translation_key)
+                if type(media_dict) is dict:
```
lindsay-stevens, line 777: `if isinstance(media_dict, dict)`

pyxform/validators/pyxform/missing_translations_check.py
```diff
@@ -88,7 +88,7 @@ def process_cell(typ, cell):
     if 0 < len(sheet_data):
         # e.g. ("name", "q1"), ("label", {"en": "Hello", "fr": "Bonjour"})
         for column_type, cell_content in sheet_data[0].items():
-            if column_type == constants.MEDIA:
+            if column_type == constants.MEDIA and type(cell_content) is dict:
```
lindsay-stevens, line 91: `if isinstance(cell_content, dict)`

tests/test_sheet_columns.py
```diff
@@ -75,6 +75,17 @@ def test_big_image_invalid_if_no_image(self):
...
 
+    def test_media_column__is_ignored(self):
+        self.assertPyxformXform(
+            name="data",
+            md="""
+            | survey |      |      |       |
+            |        | type | name | media |
+            |        | text | c    | m.png |
+            """,
+            xml__excludes=["m.png"],
```
lindsay-stevens, line 86: xml__xpath_match with a negative assertion, or xml__xpath_count with expected zero?

lindsay-stevens:
> I don't have time to design a good solution right now

Is #601 a common or severe problem? The issue description and this PR don't seem bad to me, so some additional context might be useful for future reference.

lindsay-stevens (approved):

tests/test_sheet_columns.py
lognaturel: Well... I don't even know what to not expect, exactly. 😅 I wanted to make sure that it wasn't defaulting to one of the media types.

lognaturel:
> Is https://github.com/XLSForm/pyxform/issues/601 a common or severe problem?

I don't think so. How about we close #601 after this and I make a doc change: https://github.com/getodk/docs/issues/1546

PR #639, opened 2023-01-20
Prepare v1.12 release

lognaturel:
I did not update Validate because there are no JavaRosa changes that affect pyxform.


yanokwa (approved):
LGTM

PR #644, opened 2023-05-10
Pass through itemset ref case

### LINKED ISSUE #642: Case is lost when specifying custom values for select_one_from_file label and value
pyxform v1.12.0

#### Problem description
It is possible to specify custom elements/columns as label or value for use with `select_one_from_file` as documented [here](https://docs.getodk.org/form-question-types/#customizing-label-and-value). Currently, pyxform sets all parameter keys and values to lowercase. In this case, case specified by the user should be preserved.

In `tests/test_external_instances_for_selects.py`:
```python
    def test_param_value_case_preserved(self):
        """Should find that parameters value/label override the default itemset name/label with case preserved."""
        md = """
        | survey |                                        |         |         |                      |
        |        | type                                   | name    | label   | parameters           |
        |        | select_one_from_file cities{ext}       | city    | City    | value=VAL, label=lBl |
        """
        for ext, xp_city, xp_subs in self.xp_test_args:
            with self.subTest(msg=ext):
                self.assertPyxformXform(
                    name="test",
                    md=md.format(ext=ext),
                    xml__xpath_match=[
                        xp_city.model_external_instance_and_bind(),
                        xp_city.body_itemset_nodeset_and_refs(value="VAL", label="lBl"),
                    ],
                )
```
Form: [geojson-case.xlsx](https://github.com/XLSForm/pyxform/files/11444524/geojson-case.xlsx)
Attachment: [features.geojson.zip](https://github.com/XLSForm/pyxform/files/11444525/features.geojson.zip)

#### Expected behavior
pyxform should pass through the case specified in the form definition.

lognaturel:
Closes #642

I decided to go with an exclude list for case normalization when first parsing parameters. The alternative I considered was to move case normalization to when parameter values are used by `pyxform`. I don't feel strongly between the two. Currently `label` and `value` are the only parameters that get set to user-provided strings. All of the others are numeric or enum values.

I also considered making sure all clients use label and value case-insensitively and leaving `pyxform` as it is. Even if we do choose to go that route, I don't think `pyxform` should be responsible for case normalization.

The only risk I can see would be if some client somewhere counts on `label` and `value` references always being lowercase. However, alternate clients are not common.

No.

### DIFF
--- a/pyxform/validators/pyxform/parameters_generic.py
+++ b/pyxform/validators/pyxform/parameters_generic.py
@@ -4,6 +4,8 @@ from pyxform.errors import PyXFormError
 
 PARAMETERS_TYPE = Dict[str, Any]
 
+CASE_SENSITIVE_VALUES = ["label", "value"]
+
 
 def parse(raw_parameters: str) -> PARAMETERS_TYPE:
     parts = raw_parameters.split(";")
@@ -21,7 +23,7 @@ def parse(raw_parameters: str) -> PARAMETERS_TYPE:
             )
         k, v = param.split("=")[:2]
         key = k.lower().strip()
-        params[key] = v.lower().strip()
+        params[key] = v.strip() if key in CASE_SENSITIVE_VALUES else v.lower().strip()
 
     return params
 
--- a/pyxform/validators/pyxform/parameters_generic.py
+++ b/pyxform/validators/pyxform/parameters_generic.py
@@ -4,6 +4,7 @@ from pyxform.errors import PyXFormError
 
 PARAMETERS_TYPE = Dict[str, Any]
 
+# Label and value are used to match against user-specified files so case should be preserved.
 CASE_SENSITIVE_VALUES = ["label", "value"]
 
 
--- a/tests/test_external_instances_for_selects.py
+++ b/tests/test_external_instances_for_selects.py
@@ -242,6 +242,24 @@ class TestSelectFromFile(PyxformTestCase):
                             error__contains=[msg],
                         )
 
+    def test_param_value_case_preserved(self):
+        """Should find that parameters value/label override the default itemset name/label with case preserved."""
+        md = """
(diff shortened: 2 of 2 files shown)


lognaturel:
@lindsay-stevens although this has been an issue for a while, now that we know about it I think it should be fixed as soon as we can. There's a Central release that's planned for next week and I'd love to include it.

Very interested in seeing what you think of the approach and whether you find it pragmatic or backwards.

tiritea:
Not sure I entirely understand the full context, but...

>The only risk I can see would be if some client somewhere counts on label and value references always being lowercase. 

In my client I'm certainly not making assumptions that select values are supposed to be case-insensitive (eg always lowercase).

FWIW I dont see any mention of case (in)sensitivity in the spec: https://www.w3.org/TR/2003/REC-xforms-20031014/slice8.html#ui-common-choices-value

lognaturel:
> select values

This is about the `value` and `label` **references** for defining itemsets. By default the `value` reference is `name` and the `label` reference is `label`. But those can be customized if e.g. you are using a file from another system that doesn't nave `name` and `label` elements/attributes/columns. In your XLSForm, you can specify something like `value=XYZ_ID label=DisplayName`. But currently the two values (`XYZ_ID` and `DisplayName`) are set to lowercase.

In an XML context it would be very unusual/go against the spec to treat element names as case-insensitive. I think it's would be a bit weird with CSV column names and geojson properties too though probably more common. Regardless, it's clear to me that `pyxform` should pass through the case provided by the user.

tiritea:
> Regardless, it's clear to me that pyxform should pass through the case provided by the user.

+1

lindsay-stevens (approved):
@lognaturel looks good. It would be useful to add short comment in the code, or update the test description, to mention why these parameters should have preserved case - e.g. "to match the case in the external file".

PR #648, opened 2023-08-03
646: Detect instance() expressions in notes and make them into outputs

### LINKED ISSUE #646: Detect instance() expressions in notes and make them into outputs
All versions

#### Problem description
If you put an instance() expression in a note such as `instance('stuff')/root/item[name=${thing}]/label`, it's not recognized by pyxform. The XML output will be something like `instance('stuff')/root/item[name=<output value=" /data/thing "/>]/label`. This means any secondary instance property to display must be stored in a calculate.

Put the expression above in a form. For example: https://docs.google.com/spreadsheets/d/1f1CN4o62T9m5IUIhaW3wS25ga1W9RNl6TFGktAQjMKY/edit#gid=0

#### Expected behavior
instance() expressions are also put in outputs.

Desired XML output: [instance_notes.xml.zip](https://github.com/XLSForm/pyxform/files/11900070/instance_notes.xml.zip)

We also will need to think of where best to document this pattern.

lindsay-stevens:
Closes #646

In a note label, users can show values or metadata for other form items using pyxform reference syntax e.g. ${q1}. If that reference was inside an instance call (e.g. to get the value of a secondary instance label), then the conversion would convert the pyxform token, but it should convert the whole expression.

New implementation works for pyxform references, and (nested) instance expressions with or without a pyxform reference (e.g. in a XPath predicate). It uses a similar approach to the detection of dynamic labels, where an expression lexer looks for certain grammar tokens or sequences. The lexer approach is useful here because a regex to accurately parse instance expressions is either quite complicated or not possible. Particularly considering the wide variety of XPath expressions that some users likely employ now via the workaround or will plan to after this fix.

Maybe at some stage this element of pyxform should be upgraded to use something like `pyparsing`. There is a fair amount of parsing in pyxform which could be consolidated. It may also allow for warning users of potentially invalid expressions. Example XPath parsing code with pyparsing found in projects [arelle](https://github.com/Arelle/Arelle/blob/4ac12d4999c596e7292a338eaa5b53d90a194ba5/arelle/formula/XPathParser.py) and [xpyth_parser](https://github.com/ekaats/xpyth_parser) which both happen to target XPath for XBRL purposes (no significance to XLSForm, just a coincidence).

This slots in to `survey.py` as an extra text processing layer within the existing pyxform reference replacement code, so the risk should be minimal. Obviously if there is a bug in this new code it may prevent form conversion.

It would probably be welcome news on the forum. It seems like it's something that otherwise users would assume should work.

### DIFF
--- a/pyxform/parsing/instance_expression.py
+++ b/pyxform/parsing/instance_expression.py
@@ -1,0 +1,127 @@
+import re
+from typing import TYPE_CHECKING, List, Tuple
+
+from pyxform.utils import BRACKETED_TAG_REGEX, EXPRESSION_LEXER, ExpLexerToken
+
+if TYPE_CHECKING:
+    from pyxform.survey import Survey
+    from pyxform.survey_element import SurveyElement
+
+
+def instance_func_start(token: ExpLexerToken) -> bool:
+    """
+    Determine if the token is the start of an instance expression.
+
+    :param token: The token to examine.
+    :return: If True, the token is the start of an instance expression.
+    """
+    if token is None:
+        return False
+    return token.name == "FUNC_CALL" and token.value == "instance("
+
+
+def find_boundaries(xml_text: str) -> List[Tuple[int, int]]:
+    """
+    Find token boundaries of any instance() expression.
+
+    Presumed:
+    - An instance expression is followed by an XML path expression.
+    - Any token is allowed inside a predicate (e.g. nested paths/preds/funcs).
+    - When not inside a predicate, whitespace terminates a XML path expression.
+    - instance expressions are valid inside predicates of other instance expressions.
+
+    :param xml_text: XML text that may contain an instance expression.
+    :return: Tokens in instance expression, and the string position boundaries.
+    """
+    instance_enter = False
+    path_enter = False
+    pred_enter = False
+    last_token = None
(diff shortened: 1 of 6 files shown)


lindsay-stevens:
An example from the docs ([External XML data](https://xlsform.org/en/#external-xml-data)) that currently doesn't seem be parsed as expected for note output is putting an instance expression inside another function call. Maybe out of scope? Should be possible though.
- input: `count(instance('houses')/house[rooms = current()/../rooms ])`
- output: `<label> count(<output value="instance('houses')/house[rooms = current()/../rooms"/> ]) </label>`

lognaturel (approved):
Nice! 🤩

lognaturel:
> putting an instance expression inside another function call

Good catch. I think considering it out of scope for now is reasonable. We can see whether we use it for our own forms and/or get requests for it. For computed values like that my sense is that it generally makes sense to actually have the value in the form so it's clear what the user saw.

> Maybe at some stage this element of pyxform should be upgraded to use something like pyparsing

Maybe [tree-sitter](https://tree-sitter.github.io/tree-sitter/) would be a good option here? @eyelidlessness is likely to introduce it in Enketo for similar usage and it would be amazing to be able to share a grammar. I think it could be used in Central and Collect as well.

Issue #652, opened 2023-08-29
Introduce list_name as alias for dataset in entities sheet


lognaturel:
Moving forward, ODK Central will use "Entity List" to refer to what has so far been called "Dataset." The ODK XForms spec will continue to use `dataset` and other implementing systems can choose to use `dataset`, `entity list` or another related term.

lindsay-stevens:
Merged into master in #657

PR #654, opened 2023-09-15
Alias list_name to dataset in entities sheet

lognaturel:
Closes https://github.com/XLSForm/pyxform/issues/652

The intent is to match the user-facing language of "Entity List". We discussed some alternative column names but decided that `list_name` has the advantage of being consistent with the `choices` sheet. The meanings are slightly different but very much related.

I went back and forth on the text of error messages. At first I made them generic ("Invalid name") but then I thought that would be too hard to use. I think using "entity list" in the messages is friendlier. People who have existing forms with `dataset` are early adopters so hopefully are in a better position to figure out the issue if they run into the message.

I also added on case-sensitive checking of reserved entity property names.

This area is well-tested and the change is small so I don't think there are any.

https://github.com/getodk/docs/pull/1651

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,12 +15,14 @@ tests/test_output
 build
 dist
 pyxform.egg-info
+pip-wheel-metadata
 
 # By having build ignored above, this ignores documents created by
 # sphinx. Overall, I think this is a good thing.
 
 # a virtual environment
 env
+venv
 
 # ignore pypi manifest
 MANIFEST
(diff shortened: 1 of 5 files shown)


lognaturel:
I've targeted a v1.12.x branch off of the current v1.12.1 release so that we can release this immediately. @lindsay-stevens if you have a chance for a quick look, that'd be great since you have the context but I also recognize I'm catching you Friday afternoon! It should be straightforward for someone else to review if that's better. Thanks!

ktuite (approved):

PR #655, opened 2023-09-16
Prepare for point release off of 1.12.x branch

lognaturel:
We have some changes in the `master` branch that we want to exercise just a little bit more before releasing. We also have some small changes we'd like to release now. We're going to release off of the 1.12.x branch and update the changelog here so it's consistent with the PyPi released version.


yanokwa (approved):
LGTM

PR #656, opened 2023-09-16
Prepare v1.12.2 release

lognaturel:
Prepare v1.12.2 release in v1.12.x branch.

I verified the Validate update by converting a form with a call on the `base64-decode` function which was added since the last update.


yanokwa (approved):
LGTM

PR #659, opened 2023-09-21
Add support for specifying a custom camera app for image questions

### LINKED ISSUE #651: Add support for specifying a custom camera app for image questions
In support of https://github.com/getodk/collect/issues/5621

In the `parameters` column for `image` questions, introduce an `app` key for fields of type `image`. The value of this key should be put in a body attribute with name `intent`. 

Example: `app=com.jeyluta.timestampcamerafree`

This is very similar to the existing `intent` body attribute used to specify an external app that returns multiple values: https://github.com/XLSForm/pyxform/blob/129abe98c97025cb5b978125e5d2ba5c0c0f74f7/pyxform/section.py#L191, https://docs.getodk.org/launch-apps-from-collect/#external-apps-to-populate-multiple-fields

The value is an Android application id. My understanding is that these must always use Java package name conventions so we could do some validation there, e.g. using [regex]( https://stackoverflow.com/questions/29783092/regexp-to-match-java-package-name ). I'd like someone else to confirm that the structure is guaranteed as part of deciding whether not we validate.

grzesiek2010:
Closes #651 

This is consistent with how we handle other parameters.

This is not a risky change. It's isolated and touches only image questions.

Yes https://github.com/XLSForm/xlsform.github.io/issues/247


pyxform/xls2json.py
```diff
@@ -1324,6 +1330,20 @@ def workbook_to_json(
                     (ROW_FORMAT_STRING % row_number)
                     + " Use the max-pixels parameter to speed up submission sending and save storage space. Learn more: https://xlsform.org/#image"
                 )
+
+            if "app" in parameters.keys():
+                appearance = row.get("control", {}).get("appearance")
+                if appearance is None or appearance == "annotate":
```
lindsay-stevens, line 1336: The regex seems not quite right, e.g. package name segments after the first shouldn't start with numbers `abc.9.abc`, and they could consist of one character `org.a.a`, etc. [Flutter's validation](https://github.com/pq/flutter-intellij/blob/985d084ff742515dacedc6099954299ec57e1ec4/src/io/flutter/utils/AndroidUtils.java#L1) seems reasonably straightforward to translate. Using that would help the validation to be accurate to Android requirements, and it allows returning specific error messages. Maybe not worth checking reserved words though.

The function should live in a validator package module e.g. pyxform.validators.pyxform.android_package_name.py

<details>
<summary>This should more or less cover it</summary>

```python
def validate_android_package_name(name: str) -> bool:
    pre = "Invalid Android package name - "
    has_sep = False
    front = True
    if name is None or 0 == len(name):
        raise PyXFormError(f"{pre}package name is missing.")
    c = None
    last = None
    for c in name:
        if ("a" <= c <= "z") or ("A" <= c <= "Z"):
            front = False
            last = c
            continue
        if ("0" <= c <= "9") or c == "_":
            if not front:
                last = c
                continue
            else:
                if c == "_":
                    raise PyXFormError(
                        f"{pre}the character '_' cannot be the first character in a package segment."
                    )
                else:
                    raise PyXFormError(
                        f"{pre}a digit cannot be the first character in a package segment."
                    )
        if c == ".":
            if last == ".":
                raise PyXFormError(f"{pre}package segments must be of non-zero length.")
            has_sep = True
            front = True
            last = c
            continue
        PyXFormError(
            f"{pre}the character '{c}' is not allowed."
        )
    if c is not None and c == ".":
        raise PyXFormError(f"{pre}the package name cannot end in a '.' separator.")
    if not has_sep:
        raise PyXFormError(f"{pre}the package must have at least one '.' separator.")
    return True

```
</details>

lindsay-stevens: We should also include the row number in the error message, like on line 1330

tests/test_image_app_parameter.py
```diff
@@ -0,0 +1,84 @@
+# -*- coding: utf-8 -*-
+"""
+Test image app parameter.
+"""
+from tests.pyxform_test_case import PyxformTestCase
+
+
+class ImageAppParameterTest(PyxformTestCase):
```
lindsay-stevens, line 8: Python convention is to put `Test` at the front like `TestClassName`

tests/test_image_app_parameter.py
```diff
@@ -0,0 +1,84 @@
...
+            | survey |        |          |       |                                     |            |
+            |        | type   | name     | label | parameters                          | appearance |
+            |        | image  | my_image | Image | app=com.jeyluta.timestampcamerafree | annotate   |
+            """,
+            xml__xpath_match=[
+                "/h:html/h:body/x:upload[@intent='com.jeyluta.timestampcamerafree' and @mediatype='image/*' and @ref='/data/my_image']"
+            ],
+        )
+
+    def test_ignoring_android_package_name_in_signature_image(self):
```
lindsay-stevens, line 47: Since these last 3 are looking at images with non-annotate appearance, they could be compacted as follows (and covering the no-annotation case). Also it seems like either there should not be an `@intent` attribute, or it should not contain the `app` parameter value (not sure which is expected).

```python
    def test_ignoring_android_package_name_in_image_with_non_annotate_appearance(self):
        appearances = ("", "signature", "draw", "new-front")
        md = """
        | survey |        |          |       |                                     |            |
        |        | type   | name     | label | parameters                          | appearance |
        |        | image  | my_image | Image | app=com.jeyluta.timestampcamerafree | {case}     |
        """
        for case in appearances:
            with self.subTest(msg=case):
                self.assertPyxformXform(
                    name="data",
                    md=md.format(case=case),
                    xml__xpath_match=[
                        "/h:html/h:body/x:upload[not(@intent) and @mediatype='image/*' and @ref='/data/my_image']"
                    ],
                )
```

grzesiek2010: Nice! Done.

tests/test_image_app_parameter.py
grzesiek2010: Ok, fixed.

pyxform/xls2json.py
grzesiek2010: > We should also include the row number in the error message, [as shown here](https://github.com/XLSForm/pyxform/blob/bce47941b1bd73db85c3aa50dc32f93e278c0fd2/pyxform/xls2json.py#L1323)

Done.

> The regex seems not quite right

Right it's too simple.

Now as I think about it I'm not sure if we should validate those package names that carefully because we can easily miss something. Maybe it should be just a simple regex that verifies if the string is android package like and if not add a warning not an error? What do you think @lognaturel @lindsay-stevens

lindsay-stevens: Raising an error is typically used for "we can't continue form conversion" (bad/missing/inconsistent data) or "we're sure your form isn't going to work how you seem to want" (format/spec issues). Adding a warning is typically used for "it'll work but this probably isn't what you wanted to do" (missing languages, labels, headers) or "that's not recommended" (deprecated/not ideal feature use).

So if the Android package name validation is accurate I think it would qualify to be an error.

grzesiek2010: Yeah that makes sense. I was just wondering if it make sense to be super careful here if package name convention is not something created by us. I'm not against so if @lognaturel is not against too I think we can add it for custom camera apps now and later even use the same validation mechanism in other types of external widgets/groups.

lognaturel: Matching Flutter's validation and making it an error sounds good to me! I'm sure that if someone wants to use this feature and finds that their form is rejected when it shouldn't be we'll hear about it.

grzesiek2010: Done.

pyxform/xls2json.py
```diff
@@ -1324,6 +1331,21 @@ def workbook_to_json(
...
+                    app_package_name = str(parameters["app"])
+                    validation_result = validate_android_package_name(app_package_name)
+                    if validation_result is None:
+                        new_dict["control"] = new_dict.get("control", {})
+                        new_dict["control"].update({"intent": app_package_name})
+                    else:
+                        aaa = (ROW_FORMAT_STRING % row_number) + " " + validation_result
```
lindsay-stevens, line 1344: This line seems like a typo

grzesiek2010: Right, fixed.

grzesiek2010:
@lindsay-stevens I used `nose2` instead of `nose` (as mentioned in the README) to run tests and everything was fine (the same in `pyCharm`). I did that because I had problems with using `nose` which is no longer maintained.
It looks like the old `nose` does not recognize [str | None:](https://github.com/XLSForm/pyxform/pull/659/files#diff-e332b6986e7163c47f64958346d9623fc9018d61c015dec5f8455c83198c0fe4R4) I'm I right?

lindsay-stevens:
I recommend using the dev requirements as-is and using a supported Python version. In addition to the readme, the GitHub workflows show how to set up a dev environment from scratch.

The union operator for types was added in Python 3.10 ([docs](https://docs.python.org/3.10/library/typing.html#typing.Union)) as an alternative to `Union[X, Y]`. For the `Union[str, None]` case, it can (perhaps more clearly) be written as `Optional[str, None]`. Not related to the `nose` package.

grzesiek2010:
> The union operator for types was added in Python 3.10 ([docs](https://docs.python.org/3.10/library/typing.html#typing.Union)) as an alternative to Union[X, Y]. For the Union[str, None] case, it can (perhaps more clearly) be written as Optional[str, None]. Not related to the nose package.

Right. It's fixed.

lindsay-stevens:
@grzesiek2010 not sure if you're waiting on me but there are a few comments from the last review to address. Thanks

grzesiek2010:
> @grzesiek2010 not sure if you're waiting on me but there are a few comments from the last review to address. Thanks

ups what exactly? I think I've addressed all of them.

pyxform/xls2json.py
```diff
@@ -1324,6 +1331,20 @@ def workbook_to_json(
...
+            if "app" in parameters.keys():
+                appearance = row.get("control", {}).get("appearance")
+                if appearance is None or appearance == "annotate":
+                    app_package_name = str(parameters["app"])
+                    validation_result = validate_android_package_name(app_package_name)
+                    if validation_result is None:
+                        new_dict["control"] = new_dict.get("control", {})
+                        new_dict["control"].update({"intent": app_package_name})
+                    else:
+                        raise PyXFormError(
```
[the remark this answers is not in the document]
lindsay-stevens: @grzesiek2010

pyxform/validators/pyxform/android_package_name.py
```diff
@@ -0,0 +1,31 @@
...
+    if any(segment == "" for segment in segments):
+        return f"{prefix}package segments must be of non-zero length."
+
+    if any(segment.startswith("_") for segment in segments):
+        return f"{prefix}the character '_' cannot be the first character in a package name segment."
+
+    if any(segment[0].isdigit() for segment in segments):
+        return f"{prefix}a digit cannot be the first character in a package name segment."
+
+    pattern = re.compile(r"[^a-zA-Z0-9._]")
```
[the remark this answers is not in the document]
lindsay-stevens: @grzesiek2010

tests/test_image_app_parameter.py
```diff
@@ -0,0 +1,81 @@
+# -*- coding: utf-8 -*-
+"""
+Test image app parameter.
+"""
+from tests.pyxform_test_case import PyxformTestCase
+
+
+class TestImageAppParameter(PyxformTestCase):
```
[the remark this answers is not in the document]
lindsay-stevens: @grzesiek2010

lindsay-stevens:
@grzesiek2010 I thought you could see the comments because you responded to the one about the Union operator. I have tagged you in the remaining ones. Thanks

pyxform/xls2json.py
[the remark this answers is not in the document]
grzesiek2010: Done.

pyxform/validators/pyxform/android_package_name.py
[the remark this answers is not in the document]
grzesiek2010: > Not a huge deal but please compile the regex at module level (see for example in xls2json.py).

Done

> On one hand the way you've rewritten this validation code is clearer to read. But the benefit of the Flutter way was that the error told the user exactly which character was not allowed. How can we make it clear to the user what is permitted? Putting the regex in the error message might be fine since it's a somewhat advanced feature.

How about something like this: `Parameter 'app' has an invalid Android package name - the package name contains characters that are not allowed. Package names can only include letters (a-z, A-Z), numbers (0-9), dots (.), and underscores (_).`

tests/test_image_app_parameter.py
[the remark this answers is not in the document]
grzesiek2010: I already have such tests in https://github.com/XLSForm/pyxform/pull/659/files#diff-c18c0587ea650f0a9927f6f6d73373d241b4e6db91ef9f6a3ba35a887f6b233aR5 isn't in enough?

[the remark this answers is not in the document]
lindsay-stevens: They are great unit tests but they exclude any steps before the data reaches that function - e.g. parsing. Do we know that an empty or blank string will make it to the validation func? This is a reason why most tests in pyxform use PyxformTestCase to feed in a XLSForm (mentioned in [readme](https://github.com/XLSForm/pyxform#writing-tests)). Should be straightforward to refactor into repeat style like in a few of the TestImageParameters tests.

pyxform/validators/pyxform/android_package_name.py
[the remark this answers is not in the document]
lindsay-stevens: 👍  would be good to include this sort of thing in the docs entry

tests/test_image_app_parameter.py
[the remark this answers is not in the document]
grzesiek2010: Do you mean as part of https://github.com/getodk/docs/issues/1652?

pyxform/validators/pyxform/android_package_name.py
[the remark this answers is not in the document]
grzesiek2010: Do you mean as part of https://github.com/getodk/docs/issues/1652?

tests/test_image_app_parameter.py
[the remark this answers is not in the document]
grzesiek2010: Done.

pyxform/validators/pyxform/android_package_name.py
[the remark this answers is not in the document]
lindsay-stevens: Yes

lindsay-stevens (approved):

lindsay-stevens:
Thanks @grzesiek2010 ! @lognaturel I think the PR is ready to merge - do you want this in master or some other branch?

lognaturel:
We’re going to master now! Can’t wait to get all the good stuff there released. 🚀

Issue #662, opened 2023-10-24
Add update action for entities


lognaturel:
As described in https://forum.getodk.org/t/update-an-entity-from-a-form-submission-with-server-side-conflict-detection-and-resolution/43272

PR #664, opened 2023-11-04
Add support for entity updates

lognaturel:
Closes https://github.com/XLSForm/pyxform/issues/662

The complexity here comes from the fact that the spec allows for implicit actions. Specifically:
* if there's just a `dataset` and `label` specified, it's implied that submissions should always create entities
* if an `entity_id` is specified, it's implied that submissions should always update entities

That makes the implementation more complex but I continue to think it's worth it because it's more intuitive than having to explicitly specify `true()` as create or update conditions.

Given that complexity, I decided to pass through the XLSForm values in the json representation (`entities_parsing`) and then to handle all the cases when building the XML representation (`entity_declaration`).

I used a truth table to make sure I handled and tested all the combinations of `entity_id`, `create_if` and `update_if`:

```
    # id    create  update  result
    # 1     0       0       always update
    # 1     0       1       update based on condition
    # 1     1       0       error, id only acceptable when updating
    # 1     1       1       include conditions for create and update, user's responsibility to make sure they're exclusive
    # 0     0       0       always create
    # 0     0       1       error, need id to update
    # 0     1       0       create based on condition
    # 0     1       1       error, need id to update
```

Hopefully that also helps with review.

I did find it challenging to balance writing narrow, specific tests and feeling confident that all aspects of the form were taken care of with each combination of options.

There's some possible regression risk around entity creation. Hopefully the tests guard against it! The bigger risk is that I have a logic error in the new code.

Part of https://github.com/getodk/docs/issues/1678

### DIFF
--- a/pyxform/entities/entities_parsing.py
+++ b/pyxform/entities/entities_parsing.py
@@ -23,10 +23,10 @@ def get_entity_declaration(
     entity_row = entities_sheet[0]
 
     dataset_name = get_validated_dataset_name(entity_row)
-    entity_id = entity_row["entity_id"] if "entity_id" in entity_row else None
-    create_condition = entity_row["create_if"] if "create_if" in entity_row else None
-    update_condition = entity_row["update_if"] if "update_if" in entity_row else None
-    entity_label = entity_row["label"] if "label" in entity_row else None
+    entity_id = entity_row.get("entity_id", None)
+    create_condition = entity_row.get("create_if", None)
+    update_condition = entity_row.get("update_if", None)
+    entity_label = entity_row.get("label", None)
 
     if update_condition and not (entity_id):
         raise PyXFormError(
--- a/pyxform/entities/entities_parsing.py
+++ b/pyxform/entities/entities_parsing.py
@@ -28,17 +28,17 @@ def get_entity_declaration(
     update_condition = entity_row.get("update_if", None)
     entity_label = entity_row.get("label", None)
 
-    if update_condition and not (entity_id):
+    if update_condition and not entity_id:
         raise PyXFormError(
             "The entities sheet is missing the entity_id column which is required when updating entities."
         )
 
-    if entity_id and create_condition and not (update_condition):
+    if entity_id and create_condition and not update_condition:
         raise PyXFormError(
             "The entities sheet can't specify an entity creation condition and an entity_id without also including an update condition."
         )
 
-    if not (entity_id) and not (entity_label):
+    if not entity_id and not entity_label:
         raise PyXFormError(
             "The entities sheet is missing the label column which is required when creating entities."
         )
--- a/pyxform/entities/entities_parsing.py
+++ b/pyxform/entities/entities_parsing.py
@@ -20,7 +20,43 @@ def get_entity_declaration(
             "Currently, you can only declare a single entity per form. Please make sure your entities sheet only declares one entity."
         )
 
-    entity = entities_sheet[0]
+    entity_row = entities_sheet[0]
+
+    dataset_name = get_validated_dataset_name(entity_row)
+    entity_id = entity_row["entity_id"] if "entity_id" in entity_row else None
+    create_condition = entity_row["create_if"] if "create_if" in entity_row else None
+    update_condition = entity_row["update_if"] if "update_if" in entity_row else None
+    entity_label = entity_row["label"] if "label" in entity_row else None
+
+    if update_condition and not (entity_id):
+        raise PyXFormError(
+            "The entities sheet is missing the entity_id column which is required when updating entities."
+        )
+
+    if entity_id and create_condition and not (update_condition):
+        raise PyXFormError(
+            "The entities sheet can't specify an entity creation condition and an entity_id without also including an update condition."
+        )
+
+    if not (entity_id) and not (entity_label):
+        raise PyXFormError(
+            "The entities sheet is missing the label column which is required when creating entities."
+        )
+
+    return {
+        "name": "entity",
+        "type": "entity",
+        "parameters": {
+            "dataset": dataset_name,
+            "entity_id": entity_id,
+            "create": create_condition,
+            "update": update_condition,
+            "label": entity_label,
+        },
+    }
+
+
+def get_validated_dataset_name(entity):
     dataset = entity["dataset"]
 
     if dataset.startswith(constants.ENTITIES_RESERVED_PREFIX):
@@ -41,20 +77,7 @@ def get_entity_declaration(
             f"Invalid entity list name: '{dataset}'. Names must begin with a letter, colon, or underscore. Other characters can include numbers or dashes."
         )
 
-    if not ("label" in entity):
-        raise PyXFormError("The entities sheet is missing the required label column.")
-
-    creation_condition = entity["create_if"] if "create_if" in entity else "1"
-
-    return {
-        "name": "entity",
-        "type": "entity",
-        "parameters": {
-            "dataset": dataset,
-            "create": creation_condition,
-            "label": entity["label"],
-        },
-    }
+    return dataset
 
 
 def validate_entity_saveto(
--- a/pyxform/entities/entity_declaration.py
+++ b/pyxform/entities/entity_declaration.py
@@ -5,6 +5,23 @@ from pyxform.utils import node
 
 
 class EntityDeclaration(SurveyElement):
+    """
+    An entity declaration includes an entity instance node with optional label child, some attributes, and corresponding bindings.
+
+    The ODK XForms Entities specification can be found at https://getodk.github.io/xforms-spec/entities
+
+    XLSForm uses a combination of the entity_id, create_if and update_if columns to determine what entity action is intended:
+        id    create  update  result
+        1     0       0       always update
+        1     0       1       update based on condition
+        1     1       0       error, id only acceptable when updating
+        1     1       1       include conditions for create and update, user's responsibility to make sure they're exclusive
+        0     0       0       always create
+        0     0       1       error, need id to update
+        0     1       0       create based on condition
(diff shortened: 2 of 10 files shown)


tests/test_entities_create.py
```diff
@@ -26,6 +26,9 @@ def test_basic_entity_creation_building_blocks(self):
                 '/h:html/h:head/x:model/x:bind[@nodeset = "/data/meta/entity/label" and @type = "string" and @readonly = "true()" and @calculate = "a"]',
                 '/h:html/h:head/x:model[@entities:entities-version = "2022.1.0"]',
             ],
+            xml__xpath_count=[
+                ("/h:html/h:head/x:model/x:instance/x:data/x:meta/x:entity/@update", 0),
```
lindsay-stevens, line 30: In case you originally want to use only xml__xpath_match, then asserting the absence of the attribute would be:
```
/h:html/h:head/x:model/x:instance/x:data/x:meta/x:entity[not(@update)]
```

tests/test_entities_update.py
```diff
@@ -0,0 +1,172 @@
...
+            | entities |         |                      |           |
+            |          | dataset | update_if            | entity_id |
+            |          | trees   | string-length(a) > 3 | ${id}     |
+            """,
+            xml__xpath_match=[
+                '/h:html/h:head/x:model/x:bind[@nodeset = "/data/meta/entity/@update" and @calculate = "string-length(a) > 3"]',
+                '/h:html/h:head/x:model/x:instance/x:data/x:meta/x:entity[@update = "1"]',
+                '/h:html/h:head/x:model/x:bind[@nodeset = "/data/meta/entity/@id" and @type = "string" and @readonly = "true()" and @calculate = " /data/id "]',
+                '/h:html/h:head/x:model/x:instance/x:data/x:meta/x:entity[@baseVersion = ""]',
+                '/h:html/h:head/x:model/x:bind[@nodeset = "/data/meta/entity/@baseVersion" and @type = "string" and @readonly = "true()" and @calculate = "instance(\'trees\')/root/item[name= /data/id ]/__version"]',
```
lindsay-stevens, line 108: Due to single quotes, this case would need to use `xml__xpath_exact` for the full instance expression. [See docstring here](https://github.com/XLSForm/pyxform/blob/aeedfca54652cab2efb5f6f7de4d242325d22261/tests/test_dynamic_default.py#L62) and [this test for example](https://github.com/XLSForm/pyxform/blob/aeedfca54652cab2efb5f6f7de4d242325d22261/tests/test_dynamic_default.py#L753). Or it could test for a partial match with contains().

pyxform/entities/entities_parsing.py
```diff
@@ -20,7 +20,43 @@ def get_entity_declaration(
             "Currently, you can only declare a single entity per form. Please make sure your entities sheet only declares one entity."
         )
 
-    entity = entities_sheet[0]
+    entity_row = entities_sheet[0]
+
+    dataset_name = get_validated_dataset_name(entity_row)
+    entity_id = entity_row["entity_id"] if "entity_id" in entity_row else None
```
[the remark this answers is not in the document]
lognaturel: Thank you! 😊

tests/test_entities_create.py
lognaturel: I didn't think of that! What I kind of wanted was `xpath_excludes` but I'm pretty happy with this -- I find it helpful to separate contains and excludes. Sounds like you're ok with leaving it as-is?

tests/test_entities_update.py
lognaturel: I don't understand this! It looks like the case is being correctly verified. If I change anything in the test expression, it shows up as a test failure and as it's written now it passes. The docstring talks about `&quot;` which is the entity for *double* quotes. Are you sure there's an issue with single quotes?

I used single quotes specifically to avoid issues with double quotes.

lindsay-stevens: My bad - I had a closer look at those dynamic tests and yeah the problem is with double quotes. It's the `Case`s that have double quotes in function calls that don't work. For some reason I thought it was the cases with `instance()` calls and/or single quotes. Anyway, in this case if you want to you could use triple quotes `"""` for the xpath string to avoid the need for a backslash escape on the single quote.

tests/test_entities_update.py
```diff
@@ -0,0 +1,172 @@
...
+            xml__xpath_match=[
+                "/h:html/h:head/x:model/x:instance/x:data/x:meta/x:entity",
+                '/h:html/h:head/x:model/x:instance/x:data/x:meta/x:entity[@dataset = "trees"]',
+                # defaults to always updating if an entity_id is specified
+                '/h:html/h:head/x:model/x:instance/x:data/x:meta/x:entity[@update = "1"]',
+                '/h:html/h:head/x:model/x:instance/x:data/x:meta/x:entity[@id = ""]',
+                '/h:html/h:head/x:model/x:bind[@nodeset = "/data/meta/entity/@id" and @type = "string" and @readonly = "true()" and @calculate = " /data/id "]',
+                '/h:html/h:head/x:model/x:instance/x:data/x:meta/x:entity[@baseVersion = ""]',
+                '/h:html/h:head/x:model/x:bind[@nodeset = "/data/meta/entity/@baseVersion" and @type = "string" and @readonly = "true()" and @calculate = "instance(\'trees\')/root/item[name= /data/id ]/__version"]',
+                '/h:html/h:head/x:model[@entities:entities-version = "2022.1.0"]',
```
lindsay-stevens, line 27: The value `2022.1.0` is coming from `constants.py`. Seeing as this update is to conform to an Entities spec change, will this need to be incremented to the new spec version date?

pyxform/entities/entities_parsing.py
```diff
@@ -20,7 +20,43 @@ def get_entity_declaration(
...
+    create_condition = entity_row.get("create_if", None)
+    update_condition = entity_row.get("update_if", None)
+    entity_label = entity_row.get("label", None)
+
+    if update_condition and not (entity_id):
+        raise PyXFormError(
+            "The entities sheet is missing the entity_id column which is required when updating entities."
+        )
+
+    if entity_id and create_condition and not (update_condition):
```
lindsay-stevens, line 36: It's in a few places in this changeset but I will just tag this instance - I found these redundant parentheses confusing. If the value is `None` then it's `not None` and otherwise it's a truthiness test on the contents of the parentheses e.g. `not ""`. But at first reading it seemed like it was testing `not tuple()` or similar.

pyxform/entities/entity_declaration.py
```diff
@@ -8,43 +8,82 @@ class EntityDeclaration(SurveyElement):
     def xml_instance(self, **kwargs):
```
lindsay-stevens, line 8: For the `EntityDeclaration` class I get a Pycharm warning that it hasn't implemented all base class abstract methods, namely `xml_control`. Could you paste something like the following to make it clear it's expected that not control should be emitted for an Entity?

```python
    def xml_control(self):
        raise NotImplementedError()
```

pyxform/entities/entities_parsing.py
```diff
@@ -20,7 +20,43 @@ def get_entity_declaration(
             "Currently, you can only declare a single entity per form. Please make sure your entities sheet only declares one entity."
         )
 
-    entity = entities_sheet[0]
+    entity_row = entities_sheet[0]
```
lindsay-stevens, line 23: I think a copy of the truth table from the PR description would be useful to put in the docstring for `get_entity_declaration`, or for `EntityDeclaration` class (with one referencing the other for "more details"). I found that truth table exceptionally helpful

lindsay-stevens (approved):
I stepped through the spec, truth table, and code and it seems to all check out - so, great! I approve. I added a few minor suggestion comments.

I think it's worth considering calling the validation logic from `EntityDeclaration` (during XML processing) either in addition to or instead of from `workbook_to_json`. A key benefit to checking during the latter is that it's possible to tell the user where they made a mistake in the XLSForm, but users can only have 1 entity per form so that is kind of redundant. Also by calling the validation during the XML pipeline it can apply to data coming in via the JSON entrypoint to pyxform (e.g. `builder.py`).

pyxform/entities/entities_parsing.py
lognaturel: 😬 I don't know, I feel very uncomfortable with a unary boolean operator that's a word AND that works on all types! I went through and fixed some other instances I'd littered throughout.

tests/test_entities_update.py
lognaturel: THANK YOU. I'd somehow convinced myself this wasn't necessary because the spec change is additive only. You're absolutely right. 🙏

lognaturel:
Thanks so much for the thorough review, all the great suggestions, and particularly for catching that the spec change should be versioned. 🙏 

> I think it's worth considering calling the validation logic from EntityDeclaration (during XML processing) either in addition to or instead of from workbook_to_json

This is an interesting suggestion that I'd rather chew on and address in a follow-up PR.

> A key benefit to checking during the latter is that it's possible to tell the user where they made a mistake in the XLSForm, but users can only have 1 entity per form so that is kind of redundant. 

The current messages use a lot of spreadsheet language too.

> by calling the validation during the XML pipeline it can apply to data coming in via the JSON entrypoint

Do we know that it's for sure being used?

lindsay-stevens:
Thanks! To summarise the related slack thread, the behaviour in this PR is that if a XLSForm uses entity updates then it gets the newer spec version `2023.1.0` and otherwise it gets the older spec version `2022.1.0`. This is to allow forms not using the new features to be accepted by older versions of Central. If a form uses entity updates then older Central wouldn't know what to do with them so rejecting the form based on the spec version avoids users setting up forms that don't work as expected.

Issue #592, opened 2022-03-04
Create a framework for adding parameters


lognaturel:
Some time ago, we introduced a convention of using a `parameters` column for specifying attributes that only apply to a particular question type as key/value pairs. This allows us to do specialized validation on those parameters without adding columns to forms.

That pattern has served us well and we're likely to keep adding to it. Checking for and validating parameters is now a bunch of spaghetti code. Ideally that functionality could be pulled out for easier extensibility and maintenance. 

One high-level concept would be to introduce a json spec for specifying parameters. That might make it possible to define what types they apply to and what function should be used to validate their values. Some types of things that might be tricky to represent: 
* the `value` and `ref` params only apply to `select_*_from_file`, not other selects
* the `capture-accuracy` param only applies to `geopoint` and not to `geoshape` and `geotrace`. Some other params apply to all three.

lindsay-stevens:
Relevant https://github.com/XLSForm/pyxform/issues/651#issuecomment-1745392057

jkuester:
Would you consider just passing though any parameters that do not match the pre-defined set of parameters for a particular question type? Basically, I am interested in being able to set a custom param on a question and then have that parameter get set as an attribute in the xml.  This would provide implementations of ODK the flexibility to parameterize custom form features without needing to fork Pyxform (or upstream parameters that are specific to a particular platform).  

In [the CHT](https://github.com/medic/cht-core) ecosystem, we have been stuck on a fork of Pyxform because there is some additional data we need to load from the xlsx file to support custom form functionality. I am hoping to find a "supported" way to do this without having to customize Pyxform.  

An example of what we are running into is https://github.com/medic/cht-core/issues/8681.  We have a Countdown timer widget (displays a simple countdown timer in the form) and we want to make it just a custom _appearance_ for the `trigger` question type (so that the timer is displayed to the user and `OK` value gets set as the question value once the timer runs out). The trouble is that we need a way for a form to customize the _duration_ of the timer. I was hoping to be able to do this via a parameter....  Definitively open to other ideas on how to address this challenge!  (and let me know if I should have logged this as a separate issue).  Thanks!

lognaturel:
Hi @jkuester!

Is there a reason you haven’t proposed it for inclusion in the ODK spec? You can do that at https://forum.getodk.org/c/ideas/9 if it seems generally useful. 

Alternately, have you considered the extension points listed at https://xlsform.org/en/#advanced-use-and-extensibility ?

[...]

PR #667, opened 2023-11-21
Add rows to parameters column for text type 

### LINKED ISSUE #616: Add rows to parameters column for text type
It's been part of the ODK XForms spec forever but is not conveniently exposed.

https://forum.getodk.org/t/odk-collect-text-entry-options-limited-by-single-line-design/37942/6?u=ln

grzesiek2010:
Closes #616 

The solution is consistent with how we handle other parameters in other question types. In My previous pull request, I added a new parameter for image questions (see https://github.com/XLSForm/pyxform/pull/659) and this one solves the problem in the same way. I haven't considered any other approaches. I know there is a plan to [Create a framework for adding parameters](https://github.com/XLSForm/pyxform/issues/592) and then the solution would be different but for now we should follow the existing patterns.

This is not a risky change. It's isolated and touches only reading parameters used in text questions.

Yes: https://github.com/XLSForm/xlsform.github.io/issues/254

### DIFF
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -1306,6 +1306,21 @@ def workbook_to_json(
             parent_children_array.append(new_dict)
             continue
 
+        if question_type == "text":
+            new_dict = row.copy()
+
+            if "rows" in parameters.keys():
+                try:
+                    int(parameters["rows"])
+                except ValueError:
+                    raise PyXFormError("Parameter rows must have an integer value.")
+
+                new_dict["control"] = new_dict.get("control", {})
+                new_dict["control"].update({"rows": parameters["rows"]})
+
+            parent_children_array.append(new_dict)
+            continue
+
         if question_type == "photo":
             new_dict = row.copy()
 
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -1308,6 +1308,7 @@ def workbook_to_json(
 
         if question_type == "text":
             new_dict = row.copy()
+            parameters_generic.validate(parameters=parameters, allowed=("rows",))
 
             if "rows" in parameters.keys():
                 try:
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -1314,7 +1314,10 @@ def workbook_to_json(
                 try:
                     int(parameters["rows"])
                 except ValueError:
-                    raise PyXFormError("Parameter rows must have an integer value.")
+                    raise PyXFormError(
+                        (ROW_FORMAT_STRING % row_number)
+                        + " Parameter rows must have an integer value."
+                    )
 
                 new_dict["control"] = new_dict.get("control", {})
                 new_dict["control"].update({"rows": parameters["rows"]})
--- a/tests/test_rows.py
+++ b/tests/test_rows.py
@@ -47,5 +47,7 @@ class TestRows(PyxformTestCase):
                     name="data",
                     md=md.format(case=case),
                     errored=True,
-                    error__contains=["Parameter rows must have an integer value."],
+                    error__contains=[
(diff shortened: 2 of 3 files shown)


pyxform/xls2json.py
```diff
@@ -1306,6 +1306,21 @@ def workbook_to_json(
             parent_children_array.append(new_dict)
             continue
 
+        if question_type == "text":
+            new_dict = row.copy()
+
+            if "rows" in parameters.keys():
```
lindsay-stevens, line 1312: Please validate the allowed parameters - see usage of `parameters_generic.validate(` for geopoint for example.

pyxform/xls2json.py
```diff
@@ -1306,6 +1306,21 @@ def workbook_to_json(
...
+                try:
+                    int(parameters["rows"])
+                except ValueError:
+                    raise PyXFormError("Parameter rows must have an integer value.")
```
lindsay-stevens, line 1316: Please include the row number in the error message.

tests/test_rows.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
```
lindsay-stevens, line 1: Please prefix the file name e.g. `test_parameters_rows.py` so it's clearer that this is about a parameter called 'rows' and easier to find.

tests/test_rows.py
```diff
@@ -0,0 +1,51 @@
...
+"""
+Test text rows parameter.
+"""
+from tests.pyxform_test_case import PyxformTestCase
+
+
+class TestRows(PyxformTestCase):
```
lindsay-stevens, line 8: What happens if you provide both `body::rows` and the rows parameter - either with the same value or different values? Does one of them "win" (if so, which)? Should an error be raised if both are present?

lindsay-stevens (changes requested):

lindsay-stevens:
> Nothing to discuss here.

I mean, OK, but that's a bit flippant. You could say whether or not you followed the existing patterns for parameters (and why), whether or not you searched for other issues that might overlap with the linked issue, whether or not you tested the XForm output with Collect and/or Enketo, etc.

I often use PR descriptions, issues, and commit messages to do code archaeology to understand why things are a certain way, so adding some detail is worth the effort even if it seems obvious now.

pyxform/xls2json.py
grzesiek2010: Done.

pyxform/xls2json.py
grzesiek2010: Done.

tests/test_rows.py
grzesiek2010: Done.

grzesiek2010:
> I mean, OK, but that's a bit flippant. You could say whether or not you followed the existing patterns for parameters (and why), whether or not you searched for other issues that might overlap with the linked issue, whether or not you tested the XForm output with Collect and/or Enketo, etc.

> I often use PR descriptions, issues, and commit messages to do code archaeology to understand why things are a certain way, so adding some detail is worth the effort even if it seems obvious now.

Yeah, that makes sense. Even if something is obvious adding a short description might be valuable for other devs.

tests/test_rows.py
grzesiek2010: The value specified in the parameters wins. I've added a test to check it https://github.com/XLSForm/pyxform/pull/667/files#diff-abdea13ad59db09df2c93db5a3d79d441fa45fbc8615a597ccf7d970cb253c32R22
I think that's ok since it's the new and recommended way of setting rows and body::rows should be deprecated and maybe even removed at some point.

lindsay-stevens (approved):

lindsay-stevens:
@grzesiek2010 thanks!

PR #671, opened 2023-11-28
#669: reject form with unknown columns in entities sheet

### LINKED ISSUE #669: Reject form with unknown columns in entities sheet
Currently, unknown columns are ignored on all sheets. This is generally convenient for the `survey` sheet because it makes it straightforward to add commentary or turn certain features off. In the case of the entities sheet (and maybe also the settings sheet), it feels like those patterns are less likely to be used.

Entity-related functionality is generally going to be added by defining new columns. For example, we recently introduced `update_if`. Because extra columns are ignored, using a form with `update_if` with an old version of `pyxform` does convert the form, it just ignores `update_if`. It would be more helpful if those forms were rejected instead. It would also be more inline with the [versioning scheme for the entities spec](https://getodk.github.io/xforms-spec/client-audit-logs) and the fact that other consumers of the spec reject versions of the spec newer than they're aware of.

lindsay-stevens:
Closes #669

This adds to the existing entities validation in the parsing module. The approach is to create an enum with the allowed column names, and if any extra columns are found, raise an error naming the unexpected columns. I went through and updated internal string usages to reference the constants module to ensure that the keys are consistent (this typing could be done with a class but `workbook_to_json` emits a dict).

Users may now get an error if they have already created forms where they have used annotation columns like those allowed in other XLSForm sheets. The Entities sheet is now "strict" in the sense that it only allows certain columns now.

Probably not, the documentation would already say which columns are required / allowed.

### DIFF
--- a/pyxform/constants.py
+++ b/pyxform/constants.py
@@ -8,6 +8,7 @@ are easier to find.
 """
 # TODO: Replace matching strings in the json2xforms code (builder.py,
 # survey.py, survey_element.py, question.py) with these constants
+from pyxform.util.enum import StrEnum
 
 TYPE = "type"
 TITLE = "title"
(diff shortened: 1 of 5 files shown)


lognaturel (approved):
YES!

PR #673, opened 2023-11-30
660: nicer error about choices sheet when extension omitted in select_.._from_file

### LINKED ISSUE #660: Misleading error about choices sheet when extension omitted in select_one_from_file
pyxform v1.12.2

#### Problem description
If `select_one_from_file` is given an argument without an extension, we get `There should be a choices sheet in this xlsform. Please ensure that the choices sheet has the mandatory columns 'list_name', 'name', and 'label'.`

Convert a form with a `select_one_from_file` with an argument without an extension like [Bad select one from file.xlsx](https://github.com/XLSForm/pyxform/files/12752833/Bad.select.one.from.file.xlsx)

#### Expected behavior
The error should be about specifying the filename extension (e.g. `Filename for select_one_from_file "things" should include .csv, .xml or .geojson extension`). I think it's reasonable to assume that if an extension isn't specified, the form designer made a mistake (it's rare to save files without extensions).

lindsay-stevens:
Closes #660

Previously, an error about a missing sheet name may be thrown, e.g.  if the form didn't contain any regular choices. This change explicitly checks for file extension, in order to show a nicer error message. The existing constants, aliases, and validation pattern are used.

I think this is just an enhancement, since the old behaviour was confusing, and in both cases an error is raised.

No it's already documented in the template, just an easy mistake to make

image

### DIFF
--- a/pyxform/aliases.py
+++ b/pyxform/aliases.py
@@ -18,6 +18,12 @@ control = {
     "loop": constants.LOOP,
     "looped group": constants.REPEAT,
 }
+select_from_file = {
+    "select_one_from_file": constants.SELECT_ONE,
+    "select_multiple_from_file": constants.SELECT_ALL_THAT_APPLY,
+    "select one from file": constants.SELECT_ONE,
+    "select multiple from file": constants.SELECT_ALL_THAT_APPLY,
+}
 select = {
     "add select one prompt using": constants.SELECT_ONE,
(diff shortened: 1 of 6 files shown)


lognaturel (approved):

lognaturel:
Yay, @lindsay-stevens out here making my life so much better! (And hopefully that of other users, too 😊)

lognaturel:
Just got this error again, such a big improvement!

PR #674, opened 2023-12-01
576: fix _count suffix name clash with repeats targeting another item

### LINKED ISSUE #576: Use of _count in variable name results in duplicate survey element error
pyxform v1.7.0

1. Upload [circuit_first.xlsx](https://github.com/XLSForm/pyxform/files/7658076/circuit_first.xlsx) to https://getodk.org/xlsform/.
   Screen Shot 2021-12-05 at 20 59 38
2. You will get this error: `Error: There are more than one survey elements named 'circuit_count' (case-insensitive) in the section named 'data'.`
3. If you replace `circuit_count` with `circuit_countx`, the form converts with no problems.
   Screen Shot 2021-12-05 at 21 04 08

lindsay-stevens:
Closes #576

Given a form with an item named "a_count" that holds the desired repeat count, and a repeat called "a", when pyxform injects an internal item suffixed with "_count" to hold the repeat count for "a", then pyxform emits a duplicate survey element error (2x "a_count"). This PR adds new condition to existing logic, where if the expression parses to a single pyxform reference, then the item injection is skipped.

As described in #435 (case 1), when the repeat count is just a reference to another item, then that item could be used directly, instead of injecting an item with a calculate item, thereby causing a name clash and thus the error. I tried also satisfying #435 case 3 by allowing single number expressions, but a few test failed with "something broke the parser!" which I think would be due to https://github.com/getodk/javarosa/issues/399

For future reference, the section of code in `xls2json.py` around repeats is a bit confusing but the dict structure of json["control"]["jr:count"] is a result of processing in dealias_and_group_headers (repeat_count -> control::jr:count -> dict).

ODK Validate seems OK with this, but I have not tested it with Collect or Enketo. If they or any other XForms tool expects there to always be an injected instance item to hold the count, then I expect there would be an error.

No, it's an internal form structure issue, which addresses a user-facing bug.

### DIFF
--- a/pyxform/parsing/expression.py
+++ b/pyxform/parsing/expression.py
@@ -1,0 +1,14 @@
+from typing import Iterable
+
+from pyxform.utils import parse_expression
+
+
+def is_single_token_expression(expression: str, token_types: Iterable[str]) -> bool:
+    """
+    Does the expression contain single token of one of the provided token types?
+    """
+    tokens, _ = parse_expression(text=expression.strip())
+    if 1 == len(tokens) and tokens[0].name in token_types:
+        return True
+    else:
+        return False
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -22,6 +22,7 @@ from pyxform.entities.entities_parsing import (
     validate_entity_saveto,
 )
 from pyxform.errors import PyXFormError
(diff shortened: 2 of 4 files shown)


lognaturel (approved):
🎉

PR #680, opened 2024-01-09
678: itextId node not added to instance items for choice list names with dashes in multilingual forms

### LINKED ISSUE #678: itextId node not added to instance items for choice list names with dashes in multilingual forms
pyxform v2.0.0

#### Problem description
If a choice list has a dash in its name and the form has multi-lingual labels, `itextId` nodes are not generated for the items in the instance. This means labels don't get displayed.

1. Convert a form like [this one](https://docs.google.com/spreadsheets/d/14X7QMCBH8G9VZhoG6tvOUPw9fycbNWm1ipr6XijtgHk/edit#gid=122424091) which has choice lists with dashes and multi-language choice labels
2. See that the generated instance does not include `itextId` nodes:
```
            <instance id="labels-only">
                <root>
                    <item>
                        <name>choice1</name>
                    </item>
                    <item>
                        <name>choice2</name>
                    </item>
                </root>
            </instance>
```

#### Expected behavior

The generated instance should include `itextId` nodes, just like it does if the list name has underscores:
```
            <instance id="labels_only">
                <root>
                    <item>
                        <itextId>labels_only-0</itextId>
                        <name>choice1</name>
                    </item>
                    <item>
                        <itextId>labels_only-1</itextId>
                        <name>choice2</name>
                    </item>
                </root>
            </instance>
```

#### Other information 
Image and audio choices don't seem affected, only text ones

lindsay-stevens:
Closes #678

In survey.py, _generate_static_instances has a check for multi_lang, which involves matching the list_name to the translations dict. That split the list_name on the itextId separator (a dash) but a dash can also appear in the list_name, so it was broken when a list_name contained a dash.

The fix is to still split the string, but chop off the last group (the number) to do the list_name match. Another way could be to not use a string internally (e.g. a tuple) to avoid parsing issues, then output a string when necessary.

This fixes a regression.

No

### DIFF
--- a/pyxform/survey.py
+++ b/pyxform/survey.py
@@ -319,7 +319,7 @@ class Survey(Section):
                 for items in self._translations.values()
                 for k, v in items.items()
                 if v.get(constants.TYPE, "") == constants.CHOICE
-                and k.split("-")[0] == list_name
+                and "-".join(k.split("-")[:-1]) == list_name
             )
             if 0 < len(choices):
                 multi_language = True
--- a/tests/test_translations.py
+++ b/tests/test_translations.py
@@ -1737,7 +1737,7 @@ class TestTranslationsChoices(PyxformTestCase):
         )
 
     def test_specify_other__choice_filter(self):
-        """Should raise an error since these featuers are unsupported together."""
+        """Should raise an error since these features are unsupported together."""
         md = """
         | survey  |                        |       |            |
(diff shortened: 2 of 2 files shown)


lognaturel (approved):
Oh, of course! Too bad we didn't catch that but great that it's such a straightforward fix. 🎉

PR #681, opened 2024-01-10
150: don't use regex to clean up XML following pretty-printing

### LINKED ISSUE #150: Don't use regex to clean up XML
`Survey._to_pretty_xml` pretty-prints XML output using a combination of `minidom` and regular expressions.

`minidom` formats the XML a bit, but also introduces some strange whitespace, which the regular expressions then attempt to clean up.  While this is _reasonably_ successful, it's not perfect.  Simpler would be to use a different XML formatter to `minidom` which actually works out of the box.

The in-line TODO suggests checking out PyXML as per http://ronrothman.com/public/leftbraned/xml-dom-minidom-toprettyxml-and-silly-whitespace/#best-solution:
```
from xml.dom.ext import PrettyPrint
from StringIO import StringIO

def toprettyxml_fixed (node, encoding='utf-8'):
    tmpStream = StringIO()
    PrettyPrint(node, stream=tmpStream, encoding=encoding)
    return tmpStream.getvalue()
```

lindsay-stevens:
Closes #150

Default xml.minidom pretty-printer treats each node in a mixed-content element as a separate node. As a result it adds newlines and indents for each text node and output element, which is ugly and wrong. Then survey.py cleans that up with some regex.

This change copies and customises the default `writexml` implementation from minidom, and skips newline/indent for text nodes. It also includes conditional whitespace to match the previous processing, in case Collect or Enketo expected it to be exactly that way.

Added a test to specifically enumerate various output/text mixture permutations that appear in a variety of existing tests.

I think not outputting the undesirable whitespace in the first place is better than cleaning it up afterwards. Copying `writexml` adds a bit of complexity to the code base but I think the trade off vs the regex is worth it. Being able to do this without a new dependency is preferable. It may also allow more flexibility around users inputting whitespace in their question labels, without pyxform mangling it.

While the added test should provide coverage, it's possible that some combination of text and references doesn't produce exactly the same white-spaced output as before, and assuming Collect or Enketo are sensitive to that it might change the output. But that risk seems pretty small. I'm not sure that the whitespace is actually significant in the first place.

No


lognaturel (approved):
Looks good! Pretty printing isn't done by default so it's probably not very common but good to get rid of some regex.

Feel free to merge after addressing the conflict.

PR #682, opened 2024-01-10
585: add test for allowed duplicate choices with translations

### LINKED ISSUE #585: itext identifiers prevent duplicate choices from displaying distinct labels
pyxform v1.7.0, Python v3.7.4

#### Problem description
When a form has multiple languages, the labels are stored whithin the `<itext>` section and referred to by ID.  The ID pattern uses the `name` of the choice as part of the ID string.  If choices have duplicate values, only one of the duplicate choice's label is stored as all duplicate choices would have the same ID.

Example form: [multilang_duplicatechoice.xlsx](https://github.com/XLSForm/pyxform/files/7981589/multilang_duplicatechoice.xlsx)

type | name | label::English (en) | label::Korean (ko)
---- | ---- | ---- | ----
select_one dupelist | q1 | Duplicate choices | ko__Duplicate Choices__ko

list_name | name | label::English (en) | label::Korean (ko)
-- | -- | -- | --
dupelist | 1 | Pass | 패스
dupelist | 0 | Fail | 실패
dupelist | -1 | Skipped | 건너뛴
dupelist | -1 | Not Applicable | 해당 없음

The resulting XML only has 3 choice entries in the itext section XML:
```xml
<?xml version="1.0" encoding="utf-8"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:esri="https://esri.com/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
	<h:head>
		<h:title>Survey title not set</h:title>
		<model odk:xforms-version="1.0.0">
			<itext>
				<translation lang="Korean (ko)">
					<text id="/data/q1:label">
						<value>ko__Duplicate choices__ko</value>
					</text>
					<text id="/data/q1/1:label">
						<value>패스</value>
					</text>
					<text id="/data/q1/0:label">
						<value>실패</value>
					</text>
					<text id="/data/q1/-1:label">
						<value>해당 없음</value>
					</text>
				</translation>
				<translation default="true()" lang="English (en)">
					<text id="/data/q1:label">
						<value>Duplicate choices</value>
					</text>
					<text id="/data/q1/1:label">
						<value>Pass</value>
					</text>
					<text id="/data/q1/0:label">
						<value>Fail</value>
					</text>
					<text id="/data/q1/-1:label">
						<value>Not Applicable</value>
					</text>
				</translation>
			</itext>
			<instance>
				<data id="multilang_duplicatechoice">
					<q1/>
					<meta>
						<instanceID/>
					</meta>
				</data>
			</instance>
			<bind nodeset="/data/q1" type="string"/>
			<bind jr:preload="uid" nodeset="/data/meta/instanceID" readonly="true()" type="string"/>
		</model>
	</h:head>
	<h:body>
		<select1 ref="/data/q1">
			<label ref="jr:itext('/data/q1:label')"/>
			<item>
				<label ref="jr:itext('/data/q1/1:label')"/>
				<value>1</value>
			</item>
			<item>
				<label ref="jr:itext('/data/q1/0:label')"/>
				<value>0</value>
			</item>
			<item>
				<label ref="jr:itext('/data/q1/-1:label')"/>
				<value>-1</value>
			</item>
			<item>
				<label ref="jr:itext('/data/q1/-1:label')"/>
				<value>-1</value>
			</item>
		</select1>
	</h:body>
</h:html>
```

#### Expected behavior
Ideally, each choice will display as written in the XLSX file.  

#### Other information 
Given that the existing ID pattern is attempting to create a unique ID per string (as 2 questions with the same string will have different IDs since the question properties are used to generate the ID), perhaps the ID could be switched to a GUID?

lindsay-stevens:
Closes #585

Prior to always generating secondary instances, this scenario would have not output the duplicate choices, even if allowed. Now, itext are identified by choice position so it's not a problem. This test adds to the existing suite by using translations with the duplicates.

Since no code changes are needed, this test is just to improve coverage and verify the fix.

None

No

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -126,6 +126,39 @@ class FieldsTests(PyxformTestCase):
             ],
         )
 
+    def test_duplicate_choices_with_allow_choice_duplicates_setting_and_translations(
+        self,
+    ):
+        md = """
+            | survey  |                 |      |       |
+            |         | type            | name | label::en | label::ko |
+            |         | select_one list | S1   | s1        | 질문 1     |
+            | choices |                 |      |                |
+            |         | list name       | name | label::en      | label::ko |
+            |         | list            | a    | Pass           | 패스       |
+            |         | list            | b    | Fail           | 실패       |
+            |         | list            | c    | Skipped        | 건너뛴     |
+            |         | list            | c    | Not Applicable | 해당 없음  |
+            | settings |                |                         |
+            |          | id_string      | allow_choice_duplicates |
+            |          | Duplicates     | Yes                     |
+            """
+        self.assertPyxformXform(
+            md=md,
+            xml__xpath_match=[
+                xpc.model_itext_choice_text_label_by_pos(
+                    "en",
+                    "list",
+                    ("Pass", "Fail", "Skipped", "Not Applicable"),
+                ),
+                xpc.model_itext_choice_text_label_by_pos(
+                    "ko",
+                    "list",
+                    ("패스", "실패", "건너뛴", "해당 없음"),
+                ),
+            ],
+        )
+
     def test_choice_list_without_duplicates_is_successful(self):
         md = """
             | survey  |                 |      |       |


lognaturel (approved):
Nice Korean!

PR #684, opened 2024-01-16
Add 'Other' label text for all languages defined by form

### LINKED ISSUE #683: Or_other choice label is - in multi language form
A user reports that they are seeing - instead of “Other” as the choice label when using or_other. Their form is multi-language.

I have not tried to reproduce yet and wanted to file immediately in case you have cycles before I get to it, @lindsay-stevens.

lognaturel:
Closes #683

It leads to all languages having a choice label of "Other" which is more similar to the behavior before pyxform 2 and will help with migration. A couple of implementation notes/questions are highlighted inline.

I also snuck in removing an error in case a select has both a choice filter and `or_other`. Now all selects are built on secondary instances so the choice filter case is not special.

It's limited to or_other. I can imagine new issues around that behavior.

No.


pyxform/builder.py
```diff
@@ -268,9 +268,29 @@ def _create_section_from_dict(self, d):
...
-                    and OR_OTHER_CHOICE not in itemset_choices
+                    and not any(
+                        OR_OTHER_CHOICE[const.NAME] in choice[const.NAME]
+                        for choice in itemset_choices
+                    )
                 ):
-                    itemset_choices.append(OR_OTHER_CHOICE)
+                    if isinstance(itemset_choices[0][const.LABEL], Dict):
+                        # Get all langs defined across all existing choices in this list and give them all a label of "Other"
+                        langs = set().union(
```
lognaturel, line 278: I think this is the only way to detect a multi-language choice list at this point, right?

This looks particularly tricky to understand with the black formatting, I think. It may not be hugely performant but I don't think that matters in this context.

tests/test_translations.py
```diff
@@ -1521,26 +1520,92 @@ def test_missing_translation__two_lang__warn__default(self):
...
+                xpc.model_itext_choice_text_label_by_pos(
+                    "fr", "with_us", ("l1a-fr", "l1b-fr")
+                ),
+                xpc.model_itext_choice_text_label_by_pos(
+                    "fr", "with-dash", ("l2a-fr", "l2b-fr")
+                ),
+            ],
+        )
+
+    # region or_other
```
lognaturel, line 1555: We haven't used regions yet. I find it helpful for navigation but don't feel strongly about it if we'd rather not introduce a new pattern.

pyxform/builder.py
```diff
@@ -205,8 +205,32 @@ def _add_other_option_to_multiple_choice_question(d: Dict[str, Any]) -> None:
...
+                SurveyElementBuilder._get_translated_other_label(choice_list)
+            )
+
+    @staticmethod
+    def _get_translated_other_label(choice_list) -> Dict[str, Any]:
+        translated_other_choice = OR_OTHER_CHOICE.copy()
+
+        if isinstance(choice_list[0][const.LABEL], Dict):
+            # Get all langs defined across all existing choices in this list and give them all a label of "Other"
+            langs = set().union(
```
lognaturel, line 221: I think this is the only way to detect a multi-language choice list at this point in the code, right?

This looks particularly tricky to understand with the black formatting, I think. It may not be hugely performant but I don't think that matters in this context.

PR #685, opened 2024-01-17
dev: update dependencies, testing, packaging, and linting/formatting

lindsay-stevens:
Overall this PR modernises pyxform and lints a swathe of code. Most of the commits are progressive application of linter rules, which hopefully makes it clearer what is being done and why - refer to `pyproject.toml` to see what rules were enabled in that commit.

Changes summary:

- update prod and dev/test dependencies to latest versions for Python 3.8
- update testing runner from `nose` 1.3.7 (2015) to built-in `unittest`, which can do everything required by pyxform. This also revealed some testing issues which are fixed:
  - ResourceWarnings due to input XLS/X files not being closed properly. Presumably `nose` was suppressing warnings so these were not visible until switching to unittest.
  - BadChoicesSheetHeaders test expected warnings that either don't exist or apply anymore. This test was not being run by `nose` for some unknown reason.
  - AttributeColumnsTest test expected output was not updated for secondary instances, and the fix to missing itext media. This test was not being run by `nose` for some unknown reason.
- update packaging to use `flit` since `setuptools` is no longer recommended for python. Another option is `hatch` but it looked like it might be a bit overkill and/or introduce OS dependency requirements just for package building.
- update linting to use `ruff`, which has linter/formatter rules that encompass all the existing tools (black, flake8, isort) and many more. An initial batch of extra linting rule sets are enabled and applied.
- remove unused and unmaintained `sphinx` based `docs` stub folder - not published anywhere and broken anyway

These changes support upgrading pyxform supported Python versions, since the tools replaced in this PR are either out of date already or will not be supported in future.

A lot of code in pyxform is/was crufty due to it's origins in Python 2, and as a natural result of having dozens of committers over many years. The increased linting moves the code towards a more coherent style that is more consistent with modern Python code. This should make contributions easier in future.

The update of AttributeColumnsTest seems to have revealed a bug introduced by #380 for custom attributes and repeats (will backref this ticket). Nobody seems to have complained about this though so perhaps it's a low priority to fix.

The release publishing GitHub action may not be 100% right since I'm not sure there's a nice way to test that.

The code should be functionally equivalent for users. Where applicable, the `README` file has been updated.

- [x] run ~`nosetests`~ `unittest` and verified all tests pass
- [x] run ~`black pyxform tests`~ `ruff check pyxform tests` & `ruff format pyxform tests` to lint and format code


pyxform/builder.py
```diff
@@ -60,24 +76,6 @@ def copy_json_dict(json_dict):
 
 class SurveyElementBuilder:
     # we use this CLASSES dict to create questions from dictionaries
```
lognaturel, line 78: Delete comment

pyxform/constants.py
```diff
@@ -150,3 +147,25 @@ class EntityColumns(StrEnum):
     "prefix the sheet name with an underscore. For example 'setting' "
     "becomes '_setting'."
 )
+
+BINDING_CONVERSIONS = {
```
lognaturel, line 151: Should this be in aliases?

pyproject.toml
```diff
@@ -57,7 +57,7 @@ extend-select = [
 # Potentially useful rule sets to enable in future:
 select = [
   "B",  # flake8-bugbear
-#  "C4",  # flake8-comprehensions
+  "C4",  # flake8-comprehensions
```
lognaturel, line 55: This one is nice! Lots of improved readability.

lognaturel (approved):
This is really nice, thank you! I've put a couple of small notes inline and happy for you to merge after you've considered them.

pyproject.toml
```diff
@@ -1,11 +1,90 @@
...
+  {name = "github.com/xlsform", email = "info@xlsform.org"},
+]
+description = "A Python package to create XForms for ODK Collect."
+readme = "README.rst"
+requires-python = ">=3.7"
+dependencies = [
+  "xlrd==2.0.1",        # Read XLS files
+  "openpyxl==3.1.2",    # Read XLSX files
+  "defusedxml==0.7.1",  # Parse XML
+]
```
amjadtaleb, line 14:
```suggestion
dependencies = [
  "xlrd >= 2.0.1, < 2.1",      # Read XLS files
  "openpyxl >= 3.0, < 4",      # Read XLSX files
  "defusedxml >= 0.7, < 0.8",  # Parse XML
]
```

lindsay-stevens: Thanks, but pyxform is not tested against a range of dependency versions so this would be a risky change. Currently the install requirements are pinned to exact versions so it is not a new specification.

pyxform/constants.py
lindsay-stevens: Fixed in 64903fd

pyxform/builder.py
lindsay-stevens: Fixed in 64903fd

lindsay-stevens:
@lognaturel thanks for the review! I'll wait for #694 to be decided before merging.

lognaturel:
Let's go ahead and merge! I think we're going to let #694 simmer for a bit. We can always release off a special release branch if needed.

PR #690, opened 2024-01-26
689: translations not generated for choice list used with search()

### LINKED ISSUE #689: Translations not generated for choice list used with search()
pyxform v2.0.1

#### Problem description
https://forum.getodk.org/t/text-is-not-localizable-for-default-locale/44745

Convert attached form using `--skip_validate`, notice that an inline select is generated:
```
			<select1 appearance="autocomplete search('registered_participants')" ref="/data/Select_benefic/search_idcode">
				<label ref="jr:itext('/data/Select_benefic/search_idcode:label')"/>
				<item>
					<label ref="jr:itext('/data/Select_benefic/search_idcode/foo:label')"/>
					<value>foo</value>
				</item>
				<item>
					<label ref="jr:itext('/data/Select_benefic/search_idcode/id_:label')"/>
					<value>id_</value>
				</item>
			</select1>
```

But there's no corresponding translation.

#### Expected behavior

With v1.12.1, we also get
```
					<text id="/data/Select_benefic/search_idcode/id_:label">
						<value>fullname</value>
					</text>
```

#### Other information 
I briefly thought it could be something else:
- List name of `name`
- Single-item list
- Underscore in item name

But it's definitely related to the `search()`-specific code path since an inline select is still generated.

We should continue to generate an inline select in this special case, we just need the translation to also be generated just to satisfy Validate.

lindsay-stevens:
- existing test case search_and_select didn't cover the multi-language case, which requires that itext is emitted for the inline selects.
- approach and details described in Survey._redirect_is_search_itext.

Closes #689

### DIFF
--- a/tests/test_translations.py
+++ b/tests/test_translations.py
@@ -1553,6 +1553,75 @@ class TestTranslationsChoices(PyxformTestCase):
         )
 
 
+class TestTranslationsSearchAppearance(PyxformTestCase):
+    """Translations behaviour with the search() appearance."""
+
+    def test_shared_choice_list(self):
+        """Should include translation for search() items, sharing the choice list"""
+        md = """
+        | survey  |               |       |            |           |                    |
+        |         | type          | name  | label::en  | label::fr | appearance         |
+        |         | select_one c1 | q1    | Question 1 | Chose 1   | search('my_file')  |
+        |         | select_one c2 | q2    | Question 2 | Chose 2   |                    |
+        | choices |               |       |            |           |
+        |         | list_name     | name  | label::en  | label::fr |
+        |         | c1            | na    | la-e       | la-f      |
+        |         | c1            | nb    | lb-e       | lb-f      |
+        |         | c2            | na    | la-e       | la-f      |
+        """
+        self.assertPyxformXform(
+            md=md,
+            run_odk_validate=True,
+            xml__xpath_match=[
+                "/h:html/h:body/x:select1/x:item[./x:value/text()='na']",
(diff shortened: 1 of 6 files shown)


tests/test_translations.py
```diff
@@ -1553,6 +1553,75 @@ def test_choice_name_containing_dash_output_itext(self):
...
+    def test_shared_choice_list(self):
+        """Should include translation for search() items, sharing the choice list"""
+        md = """
+        | survey  |               |       |            |           |                    |
+        |         | type          | name  | label::en  | label::fr | appearance         |
+        |         | select_one c1 | q1    | Question 1 | Chose 1   | search('my_file')  |
+        |         | select_one c2 | q2    | Question 2 | Chose 2   |                    |
+        | choices |               |       |            |           |
+        |         | list_name     | name  | label::en  | label::fr |
+        |         | c1            | na    | la-e       | la-f      |
```
lognaturel, line 1568: The item list for a select with `search()` isn't actually a true item list, it's used to configure the columns used in the attached file. A more realistic example would be the one I had in the test in the issue: a single element with a name of something like id, a label::en of something like label_en and a label::fr of label_fr. It doesn't particularly matter for testing purposes but could be confusing to come back to.

lognaturel:
Thanks for this. This makes me kind of wish we weren't as obsessed with keeping old forms working but I continue to think it's important for our users.

Forgive me -- I'm going to go ahead and merge even though you marked this as draft! I have one suggestion for the tests that I'll do a follow-up PR for. I'd really like to put this release behind us and I think this is a very pragmatic approach and is easy to figure out with the comment.

lognaturel (approved):

PR #691, opened 2024-01-26
Make tests have more realistic search() usage

lognaturel:
Follow-up to #690

This doesn't make any substantive change so I considered not doing it but I do think it's likely to help us in the future.

`search(param.csv)` uses a 1-element inline list to configure how it queries `param.csv`. It's only implemented in Collect and doesn't have a formal spec. I believe it gives an error if the supplied list has more than one element.

Test-only change so none.

No.


lognaturel:
Replaced by https://github.com/XLSForm/pyxform/pull/693

PR #693, opened 2024-01-29
689: translations not generated for choice list used with search() [tests]

lindsay-stevens:
Follow up to #690 to finish the test cases.

Covers use cases in relation to choice / translation behaviour per the [XLSForm docs](https://xlsform.org/en/#dynamic-selects-from-pre-loaded-data). I found the numbered list there a bit abstract, so for reference [this docs page](https://docs.surveycto.com/02-designing-forms/04-sample-forms/12.search-and-select.html) from SurveyCTO has good illustrations of the advanced usages.

None, it's just tests.

Not directly, but I noticed a dead link on the XLSForms docs page. Not sure if the file is gone or it's a permissions issue:

> For an example, see the [advanced search-and-select sample form here](https://drive.google.com/drive/folders/1Wx09ZFOJuiWmy0XKKtpxzGsZHfxeTO9c?usp=sharing).


lognaturel (approved):
Fantastic, thank you!

PR #697, opened 2024-02-20
694: search and pulldata secondary instance conflict

### LINKED ISSUE #694: Secondary instance generated for search() configuration likely to conflict with pulldata secondary instance
pyxform v2.0.2

#### Problem description
The fix for https://github.com/XLSForm/pyxform/pull/690 led to secondary instances being generated for the choice list used to configure `search()`. That can lead to an error if the name of that list is the same as the name of the CSV. Unfortunately, that's the pattern documented at https://xlsform.org/en/#dynamic-selects-from-pre-loaded-data so I imagine it's common.

```
    def test_search_and_pulldata(self):
        md = """
        | survey  |                   |         |             |                    |             |
        |         | type              | name    | label       | appearance         | calculation |
        |         | select_one fruits | fruit   | Question 1  | search('fruits')   |             |
        |         | calculate         | calc    |             |                    | pulldata('fruits', 'this', 'that', ${fruit}) |
        | choices |               |       |        |
        |         | list_name     | name  | label  |
        |         | fruits        | na    | la     |
        """
        self.assertPyxformXform(
            md=md,
            xml__xpath_match=[
                "/h:html/h:body/x:select1/x:item[./x:value/text()='na']"
            ],
        )
```

See the error:
```
ERROR: 'The same instance id will be generated for different external instance source URIs. Please check the form. Instance name: 'fruits', Existing type: 'pulldata', Existing URI: 'jr://file-csv/fruits.csv', Duplicate type: 'choice', Duplicate URI: 'None', Duplicate context: 'survey'.'
```

#### Expected behavior
No error.

#### Other information 
Could we do something hacky like add a uuid to the secondary instance name if it's hard not to produce it? I don't think it's used for anything.

lindsay-stevens:
Closes #694

The previous solution for search + translations was pretty close, but created the possibility of conflicts in the secondary instances. This PR removes the secondary instance that would have been emitted. It also raises an error two search feature combinations that are not supported, and even if they were, would be broken by this PR's strategy (see tests for examples, where `errored=True`). These combinations are: 1) search appearance on a select_from_file, 2) search appearance choice list used in other selects.

An alternative detailed in #694 was to merge compatible secondary instances, but @lognaturel clarified offline that this wouldn't work because the `src` attribute is how Collect/Enketo identify external instances. So having choices in a secondary instances with a non-blank `src` wouldn't work.

If users had forms which included the above 2 scenarios which now raise an error, they would need to fix up their form. If the error seems to harsh maybe a warning could be emitted, but ODK Validate raises an ugly error for these cases anyway (due to missing instance).

I tidied up and included some settings tests I wrote for recent forum issues. These settings could be included in the XLSForm reference template.


lognaturel (approved):
Hurray, that turned out not to be too bad and some nice cleanup with it too! 🎉 🙏

lognaturel:
>  1) search appearance on a select_from_file, 2) search appearance choice list used in other selects.

I don't think either of these has reasonable expected behavior and errors are an improvement. 👍

tests/test_settings.py
```diff
@@ -1,12 +1,220 @@
...
+from tests.xpath_helpers.questions import xpq
 
 
 class TestSettings(PyxformTestCase):
-    def test_instance_xmlns__is_set__custom_xmlns(self):
+    def test_title(self):
+        """Should find the title set in the XForm."""
+        md = """
+        | settings |
+        |          | title   |
```
lognaturel, line 11: We document this as the `form_title` alias.

tests/test_settings.py
```diff
@@ -1,12 +1,220 @@
...
+        self.assertPyxformXform(
+            md=md,
+            xml__xpath_match=["/h:html/h:head/h:title[.='My Form']"],
+        )
+
+    def test_id_string(self):
+        """Should find the instance id set in the XForm."""
+        md = """
+        | settings |
+        |          | id_string |
```
lognaturel, line 26: We document this using the `form_id` alias.

lognaturel:
> These settings could be included in the XLSForm reference template

Filed settings docs issue at https://github.com/XLSForm/xlsform.github.io/issues/266

tests/test_settings.py
lindsay-stevens: Updated test to match documented name, added a note to do the same for future tests

tests/test_settings.py
lindsay-stevens: Updated test to match documented name, added a note to do the same for future tests

pyxform/survey.py
```diff
@@ -780,6 +797,21 @@ def get_choices():
                         }
                     )
 
+        for q_name, list_name in search_lists:
+            choice_refs = [f"'{q}'" for q, c in non_search_lists if c == list_name]
+            if len(choice_refs) > 0:
+                choice_refs_str = ", ".join(choice_refs)
+                msg = (
+                    f"Question '{q_name}' uses the 'search' appearance, and its select "
```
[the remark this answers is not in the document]
lindsay-stevens: Updated error message to refer to `search()`, noted the difference in the test module, and renamed the test module from `test_search_appearance` to `test_search_function`.

pyxform/survey.py
```diff
@@ -680,9 +679,20 @@ def _redirect_is_search_itext(element: SurveyElement) -> None:
         except (KeyError, TypeError):
             is_search = False
         if is_search:
+            file_id, ext = os.path.splitext(element[constants.ITEMSET])
+            if ext in EXTERNAL_INSTANCE_EXTENSIONS:
+                msg = (
+                    f"Question '{element[constants.NAME]}' is a select from file type, "
+                    "with a 'search' appearance. This combination is not supported. "
+                    "Remove the 'search' appearance, or change the select type."
```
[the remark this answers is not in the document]
lindsay-stevens: Updated error message to refer to `search()`, noted the difference in the test module, and renamed the test module from `test_search_appearance` to `test_search_function`.

lindsay-stevens:
Thanks! `search()` vs `search` is a bit confusing. I've gone with `search()` for the errors and clarified it in the code + tests as well.

Issue #661, opened 2023-10-04
Update supported Python versions


lindsay-stevens:
Current supported versions of Python 3.7, 3.8 and 3.9. Version 3.7 went EOL 2023-07-27. Version 3.8 is good until 2024-10-14. Version 3.12 was released on 2023-10-02.

So support for 3.7 should be dropped, and support for versions 3.10 and 3.11 should be added. In practical terms this means updating the [GitHub actions workflow here](https://github.com/XLSForm/pyxform/blob/bce47941b1bd73db85c3aa50dc32f93e278c0fd2/.github/workflows/verify.yml#L44). Worth seeing if 3.12 works, but if it doesn't then it's probably so new that few will be using it yet.

Also update the README accordingly.

PR #699, opened 2024-03-04
Add Dockerfile and compose config for testing

spwoodcock:
#### Included

- I didn't want to install Java on my system, so created a quick docker setup for running the tests.
- I added to the `contrib` dir to make it clear that this is not maintained by the authors, but thought it could be useful to some (perhaps included in the docs?).
- The included dockerfile installs all dev dependencies, but could be easily extended to only install production dependencies in a stage, then used for a repository image to distribute pyxform: e.g. `ghcr.io/xlsform/pyxform`.
- I used this setup for testing in PR #698

- A did base this on Python 3.10, while the repo only states 3.7-3.9 are supported.
- I use pyxform without issues in a project based on Python 3.10.
- Python 3.7 is EOL, 3.8 will be EOL in about 0.5yr and 3.9 EOL in about 1.5yr.

- None

- Could possibly be included in the docs about testing, if useful.
- Simply run `docker compose -f contrib/images/docker-compose.yml run --rm pyxform` to run the tests.
- Or `docker compose -f contrib/images/docker-compose.yml run --rm pyxform -k test_xls2form_convert_via_api` for a specific test.
- Or `docker run -i ghcr.io/xlsform/pyxform:ci --help` to use the `xlsform` CLI tool.

### DIFF
--- a/contrib/images/Dockerfile
+++ b/contrib/images/Dockerfile
@@ -1,0 +1,32 @@
+ARG PYTHON_IMG_TAG
+FROM python:${PYTHON_IMG_TAG}-slim-bookworm as base
+
+
+FROM base as build
+RUN set -ex \
+    && apt-get update \
+    && DEBIAN_FRONTEND=noninteractive apt-get install \
+    -y --no-install-recommends \
+        "build-essential" \
+    && rm -rf /var/lib/apt/lists/*
+WORKDIR /code
+COPY ../../ ./
+# Install with dev dependencies for tests
+RUN pip install .[dev] \
+    --user --no-warn-script-location --no-cache-dir
+
+
+FROM base as runtime
+ENV PATH="/root/.local/bin:$PATH"
+RUN set -ex \
+    && apt-get update \
+    && DEBIAN_FRONTEND=noninteractive apt-get install \
+    -y --no-install-recommends \
+        "openjdk-17-jdk" \
+    && rm -rf /var/lib/apt/lists/*
+# Copy Python deps from build to runtime
+COPY --from=build \
+    /root/.local \
+    /root/.local
+WORKDIR /code
+ENTRYPOINT ["xlsform"]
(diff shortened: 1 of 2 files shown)


lindsay-stevens:
Please have a look at [my comment here](https://github.com/XLSForm/pyxform/pull/698#issuecomment-1985344325) on contributing to pyxform.

The current pyxform development processes do not use Docker by default. For a potential new contributor there are substantially less pitfalls involved in installing a JRE, compared to standing up Docker and/or Compose, let alone understanding how to use these effectively in a development context. We would need to at least mention in the `README` how to use them. It may be more appropriate to publish Docker files in a separate repository, as [has been done here](https://github.com/seadowg/pyxform-docker/tree/master) by another ODK team member.

In relation to supported python versions, there is an [issue here](https://github.com/XLSForm/pyxform/issues/661) to address that and it's at the top of the "Next" milestone [here](https://github.com/XLSForm/pyxform/milestone/3) - meaning it will be worked on very soon.

spwoodcock:
I won't dispute that installing a JRE is pretty easy, but would argue that installing Docker and Compose are not that much more complex. In a couple of commands you can have things working.

The config is aimed at developers, specifically for testing.
What if I am a Java developer requiring multiple runtimes for different projects?
Docker makes this type of setup easy.

Either way, would it be worth linking the docker configuration in the docs here?
I had no idea it existed & would have happily contributed to that repo!
(although note within that repo it would only be a Dockerfile specifically for **using** pyxform, not for **running tests**).

lognaturel:
> installing Docker and Compose are not that much more complex

A few years ago we had a lot more Dockerfiles for dev but we scaled it back because we found we often forgot to update them and that we preferred not having to manage a Docker daemon on our machines and make sure we had enough memory and disk space for other things. The latter has gotten somewhat better with Orbstack but I think we're all still feeling a little burned and use more native tools for dev environment management ([`nix`](https://nixos.org/), [`jenv`](https://www.jenv.be/), etc). This is definitely a team culture kind of thing.

How about putting your dockerfile in a repo and we'll link to it from the readme? I have a proposed structure for this that I'll put a PR up for in a moment.

PR #709, opened 2024-06-03
707: Detect instance expressions with double quotes

### LINKED ISSUE #707: Detect instance expressions with double quotes
https://github.com/XLSForm/pyxform/pull/648 detects instance() expressions in labels. I believe it only does so if the instance name is in single quotes. Instance names can be in single or double quotes.

Forum post: https://forum.getodk.org/t/conditionally-change-text-style-an-entity-issue/46907/5?u=ln

lindsay-stevens:
Closes #707 

Like other functions, argument value(s) (instance name) should be allowed to be wrapped in single or double quotes. Previously only singles quotes worked.

Uses existing XML generation function. Not sure why there was an f-string used before, maybe some kind of conflict with something else that has since been resolved.

It's more consistent with the rest of pyxform now, but maybe some users don't want XML escapes here. As shown in the updated tests, other characters that will be escaped now are `<>&`.

No

### DIFF
--- a/pyxform/parsing/instance_expression.py
+++ b/pyxform/parsing/instance_expression.py
@@ -1,7 +1,7 @@
 import re
 from typing import TYPE_CHECKING
 
-from pyxform.utils import BRACKETED_TAG_REGEX, EXPRESSION_LEXER, ExpLexerToken
+from pyxform.utils import BRACKETED_TAG_REGEX, EXPRESSION_LEXER, ExpLexerToken, node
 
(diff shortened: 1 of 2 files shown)


lognaturel (approved):
A quick form I used to complement the tests: https://docs.google.com/spreadsheets/d/13Ow7UXMt5wjMXawa8XrJcoxGhnvwk91xHovFS8K0IVI/edit#gid=0

A small change with a big impact!

PR #712, opened 2024-06-25
698: library API for non-path input, accept markdown and dict input

### LINKED ISSUE #599: Move markdown to pyxform functionality out of tests and into main package
⚠️ See [this comment](https://github.com/XLSForm/pyxform/issues/599#issuecomment-1081514288) for how we've decided to proceed.

pyxform v1.7.0 - v1.9.0, Python 3.8.13

#### Problem description

I have been using the functionality within [pyxform_test_case](https://github.com/XLSForm/pyxform/blob/master/tests/pyxform_test_case.py) in writing tests using markdown to minimise or eliminate the use of Excel files within the codebase. I believe the change introduced by the PR https://github.com/XLSForm/pyxform/pull/562 eliminated the inclusion of the package during distribution and as such the module is not available in third party code bases that rely on it.

 ```python
Python 3.8.13 (main, Mar 24 2022, 04:09:24) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
>>> from pyxform.tests.pyxform_test_case import PyxformMarkdown
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'pyxform.tests'
```

#### Expected behaviour

Could we make this module available under `pyxform.test_utls` or any relevant path and allow the continued use by third party code bases for writing tests? The goal is the same as was for this package to reduce the use of binary XLSForm files.

### PR DESCRIPTION
**Pull request by lindsay-stevens**
Closes #698
Closes #599

- add: xls2xform.convert for library users to call pyxform without needing to use files (accepts bytes/bufferedreader (from `open()`)/strings)
  - accepts markdown input since this is widely used by pyxform
  - accepts dict to avoid needing to use internal funcs that may change (also you could more easily go from pandas/database -> dict -> pyxform -> xform; the dict use case could be improved by adding a schema or something).
- chg: avoid writing to files unless validate=True (for ODK Validate)
  - also avoid assuming any files were written, e.g. missing_ok=True
- chg: move xls/x_sheet_to_csv, sheet_to_csv from utils.py to xls2json_backends.py because they are backends for csv input.
- chg: move md_to_dict from test directory into xls2json_backends.py
- chg: refactor pyxform_test_case.py to use xls2xform.convert only, instead of internal funcs associated with md_to_dict, so that the existing tests check API stability e.g. file types, dict input, etc.
- chg: survey.py print_xform_to_file returns xml, so now survey.to_xml() doesn't create xml twice (once itself, once via print_xform_to_file).
  - similarly, xls2xform would read the input XLSForm a 2nd time if there were external choices. Now it uses the data from the first reading.

Differences to #698: as shown in the new tests  in `test_xls2xform.py`, there are a few reasonable call patterns that don't return BytesIO; and adding a new function allows more easily tailoring it to the library use case.

Was not initially planning on the markdown/dict changes but considering a) #599 has been open for a while, b) the recent review of pyxform-related projects showed there is a need for alternative input formats, or at least more flexibility, c) I figured that it would be best to support markdown/dict now so that the API was designed for that, rather than needing to potentially refactor for it later. I think the main value in pyxform is in workbook_to_json -> survey.py -> xform, so by adding the option of providing a dict, library users can do whatever deserialisation is necessary and pass the data to pyxform.

There are some functions moved around as mentioned above, so at upgrade time, library users would need to change import locations or switch to using `xls2xform.convert`. Functionality should be the same.

I think there's a good amount of info in the docstrings and tests, but a readme section could be added to make it easier to take in, if required.

### CODE REVIEW

**Comment by lindsay-stevens** [comment]
FYI @spwoodcock

**Comment by lindsay-stevens** [comment]
Forum thread for community input: https://forum.getodk.org/t/for-developers-that-use-pyxform-new-library-api-proposal/47748

File: tests/test_form_name.py [hunk 1]
```diff
@@ -7,33 +7,17 @@
 
 class TestFormName(PyxformTestCase):
     def test_default_to_data_when_no_name(self):
-        """
-        Test no form_name will default to survey name to 'data'.
-        """
-        survey = self.md_to_pyxform_survey(
-            """
+        """Should default to form_name of 'test_name', and form id of 'data'."""
```
lognaturel, line 10: This seems backwards. I'd expect the default root node name to be `data`. The form id default I'd expect to be the test name but that's less important.

tests/xform_test_case/test_bugs.py
```diff
@@ -11,148 +12,30 @@
...
-    def test_conversion(self):
-        filename = "repeat_date_test.xls"
-        self.get_file_path(filename)
-        expected_output_path = os.path.join(
-            test_expected_output.PATH, self.root_filename + ".xml"
+    def test_conversion_raises(self):
+        """Should find that conversion results in an error being raised by pyxform."""
+        cases = (
+            ("group_name_test.xls", "[row : 3] Question or group with no name."),
+            # (
```
lognaturel, line 27: Is this intentionally commented out?

pyxform/constants.py
```diff
@@ -34,6 +34,7 @@
 SUBMISSION_URL = "submission_url"
 AUTO_SEND = "auto_send"
 AUTO_DELETE = "auto_delete"
+DEFAULT_FORM_NAME = "data"
```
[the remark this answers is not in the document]
lognaturel: Could we use this default for tests as well, maybe? It feels weird to me to have `test_name` be the default there (I indicated this inline with an example as well). This is not critical and could certainly be a separate issue to consider. I think it's worth mentioning now since several tests now use the default when they didn't previously.

tests/test_form_name.py
lognaturel: Looks like this is what it's been previously for some reason, it's just being exposed here.

lognaturel (approved):
I had a bit of trouble getting through this one! Maybe I should have ignored all the test changes initially. `convert` looks good. I almost recommended some parameter name changes like `enketo` -> `enketo_validate` but matching the existing `xls2xform_convert` feels right after all.

Couple of tiny questions/suggestions inline but I don't think they need to block merge.

tests/xform_test_case/test_bugs.py
lindsay-stevens: No - fixed in e2de762. Thanks

pyxform/constants.py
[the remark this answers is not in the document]
lindsay-stevens: It is a bit weird. The `test_name` comes from the parameter `form_name` which when using file input is set from the file name stem e.g. `myform.xls` -> `myform`, or the default `data` (same process for `form_id`, and `form_title` follows the `form_id`).

All (or most) of the more recent XPath tests use a form_name of `test_name` so I thought it would be good either that name or the default "data", except for cases like `test_form_name.py` etc that specifically test using a non-default name.

The current diff has 9 hits for `/test_name` whereas the full test directory has 236. Also, 130 uses of `assertPyxformXform` explicitly set `name="data"`. Previous to this PR the `pyxform_test_case.py` function `_autoname_inputs` function set `name="test_name"`, `title="test_title"` and `id_string="test_id"`. Something I like about `/test_name` is that if the default form_id is also `data` it makes code search and find and replace a little easier, but equally I could do `/data` vs `@id="data"`.

If you would prefer, I could do a follow-up PR to update the tests to all use the pyxform defaults only (unless testing a feature to use something different)?

tests/test_form_name.py
lindsay-stevens: Noted, thanks. Resolving this thread in favour of the other one for discussion of this topic

pyxform/constants.py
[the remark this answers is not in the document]
lognaturel: Yes, let's consider it as a separate low priority issue! Maybe we end up leaving it as-is. Your point about "data" being hard to search for makes sense.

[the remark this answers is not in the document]
lindsay-stevens: Follow up ticket here: https://github.com/XLSForm/pyxform/issues/715

lognaturel:
@lindsay-stevens `pyxform-http` used `sheet_to_csv` from `utils`: https://github.com/getodk/xlsform-online/pull/81/files Was that intended to be a public API or not?

lindsay-stevens:
@lognaturel it's still public, just moved to be with the other backend functions.
> chg: move xls/x_sheet_to_csv, sheet_to_csv from utils.py to xls2json_backends.py because they are backends for csv input.

This PR set the scene for making parts of pyxform non-public (by offering a library API that does what most users would want to do), but didn't make changes in that regard.

lognaturel:
> @lognaturel it's still public, just moved to be with the other backend functions.

That's a breaking change, then, and should be released as a major version bump, right? I believe that otherwise the changes are non-breaking so maybe it's worth preserving the prior APIs (calling the new functions) so it's non-breaking? Or we bend the semver rules because we think usage as a library is rare?

lindsay-stevens:
Agree it's a breaking change, and agree it's probably not going to affect many people (possibly no-one?) - the solution is to update the import path (the function bodies are unchanged). Equally, incrementing the version is OK.

PR #719, opened 2024-09-05
713: support offline entities spec v2024.1.0 via opt-in setting

### LINKED ISSUE #713: Add support for offline Entity create and update
Spec discussion at https://forum.getodk.org/t/form-spec-proposal-support-entity-updates-from-multiple-offline-clients/47890

We are going to stage this. In the next version of `pyxform`, users will need to explicitly opt in to produce forms that use the v2024.1.0 spec to use offline Entities to allow for some piloting of the functionality across all components. The opt-in flag will be removed in the following version and all Entity-related forms will be produced following the v2024.1.0 spec.

Entity-related forms can create or update Entities or do both. There is existing logic in `pyxform` to determine which of these a form definition specifies.

## Forms that can create Entities
For forms that can create Entities, the only change needed in `pyxform` is to set the `entities-version` attribute on the `model` node to `2024.1.0`.

## Forms that can update Entities
For forms that can update Entities, the `entities-version` attribute on the `model` node will need to be set to `2024.1.0`.

Additionally:
* Add `trunkVersion` and `branchId` attributes to the `entity` element
* Add the following `binds`:
```
<bind nodeset="/data/meta/entity/@trunkVersion" type="string" calculate="instance('trees')/root/item[name=/data/tree]/__trunkVersion" />
<bind nodeset="/data/meta/entity/@branchId" type="string" calculate="instance('trees')/root/item[name=/data/tree]/__branchId" />
```

## Opt in
Initially, the changes described above should be opt-in so we also need to introduce a setting column named `offline` on the `entities` sheet. It should be optional and its default value should be `no`/`false`. If its value is set to `yes`, the changes above should be applied. Otherwise, Entity forms should be generated as they are today.

lindsay-stevens:
Closes #713

Implements #713: opt-in via a new entities setting `offline` with a value `yes`. If not opted-in, retain all current behaviour. If opted-in, for the create case, use the new spec version. If opted-in, for the update case, use the new spec version and emit trunkVersion/branchId properties and bindings.

Deviations to the requirements:

1. Bindings readonly property. I used the existing function in `entity_declaration.py`, `EntityDeclaration._get_bind_node()`, which sets the attribute `readonly="true()"`, which presumably is correct since this function is used for the existing `baseVersion` property with a similar structure.
2. Opt-in behaviour. Like other yes/no settings, the exact value `yes` is accepted, or it's aliases e.g. `Yes`, `true`, etc.
3. Opt-out behaviour. A user can opt out either by omitting the `offline` column, or by providing a value other than an alias for `yes` (such as `no`, `bananas`, or anything else). This seemed more consistent with the requirements than raising an error if the column is present but the value is not an alias for `yes`, although an error may make sense too.

Low risk, should be backwards compatible as per the tests suite.

In future when the new spec version is the default, presumably the `offline` setting will be removed, meaning that from then on forms will have to remove that setting or else get an error about an unexpected column. This would be a minor point of friction but beneficial in that the setting would do nothing (all forms would use spec `2024.1.0`) which may be confusing it were kept/allowed.

Updates to the entities spec and ODK docs are already in progress. The new `offline` setting is intended to be kept for only a few months, so I'm not sure if that's worth an update to the XLSForm template, especially if the intended audience during that period is insiders/advanced users. The xlsform.org page doesn't mention entities at all.

### DIFF
--- a/pyxform/constants.py
+++ b/pyxform/constants.py
@@ -114,7 +114,8 @@ CURRENT_XFORMS_VERSION = "1.0.0"
 
 # The ODK entities spec version that generated forms comply to
 ENTITIES_CREATE_VERSION = "2022.1.0"
-CURRENT_ENTITIES_VERSION = "2023.1.0"
+ENTITIES_UPDATE_VERSION = "2023.1.0"
+ENTITIES_OFFLINE_VERSION = "2024.1.0"
 ENTITY = "entity"
 ENTITY_FEATURES = "entity_features"
 ENTITIES_RESERVED_PREFIX = "__"
@@ -126,6 +127,7 @@ class EntityColumns(StrEnum):
     CREATE_IF = "create_if"
     UPDATE_IF = "update_if"
     LABEL = "label"
+    OFFLINE = "offline"
 
 
 DEPRECATED_DEVICE_ID_METADATA_FIELDS = ["subscriberid", "simserial"]
--- a/pyxform/entities/entities_parsing.py
+++ b/pyxform/entities/entities_parsing.py
@@ -1,6 +1,7 @@
 from typing import Any
 
 from pyxform import constants as const
+from pyxform.aliases import yes_no
 from pyxform.errors import PyXFormError
 from pyxform.xlsparseutils import find_sheet_misspellings, is_valid_xml_tag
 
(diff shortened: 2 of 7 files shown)


lognaturel (approved):
Thanks for all the nice refinements and for calling them out explicitly in the PR description! I agree with all the decisions you made. 👍

PR #720, opened 2024-09-05
Prepare v2.1.0 release

lognaturel:
Prepares for release

I went back and forth on whether this should be 3.0 or 2.1 because there is a change in required Python versions and a small breaking API change. I went with 2.1 because we're dropping support for Python versions that are very close to end of life and we have a `requires-python` directive. Other people have gone back and forth on this and come to various conclusions (e.g. [here](https://www.reddit.com/r/Python/comments/107g04w/would_you_increment_semver_when_removing_support/)). There is technically a small API change but it's unlikely to have been in use and is easy to respond to.

I included dependency updates here because I think they're low risk and nice to keep on top of. I also ran `ruff` and made changes accordingly.

There are deprecation warnings about `datetime.utcnow()` that are annoying when running tests. I tried to quickly address these but got some errors in the validator updater when I did the switch there. I think we should come back to it but I figured I'd leave the easy updates in here.

Feels low-risk. Probably the highest risk is around the Validate update since that's jumping a lot of JR versions. We also haven't run with the new dependency versions for very long so that could be risky.

https://github.com/getodk/docs/issues/1813
https://github.com/getodk/docs/issues/1795


lognaturel:
I am stuck trying to figure out why linting is failing! It passes for me locally in Python 3.12 and 3.10. I've verified that I'm using `ruff` 0.6.4 and that the task does a pip install.

@lindsay-stevens I'd be grateful for any ideas you may have! Please push to this branch if you want.

lindsay-stevens (approved):

lindsay-stevens:
The lint problem was that ruff didn't like the space added between the two import lines in `test_android_package_name.py`.

For future reference, although the latest openpyxl is 3.1.5, this update is to 3.1.3 because since 3.1.4, openpyxl pinned lxml to 5.1.0. Their problem in 5.1.0 ([here](https://foss.heptapod.net/openpyxl/openpyxl/-/issues/2146)) is marked as fixed in 5.1.1, but the pin isn't removed until openpyxl 3.2 which is not released yet. Also pyxform is already using 5.2.2 and this update moves to 5.3.0.

lindsay-stevens:
For anyone wondering why ODK_Validate.jar more than doubled in size (from ~5MB to ~13MB), most of it (~5MB) seems to be because the underlying [javarosa](https://github.com/getodk/javarosa/pull/738) library recently added a dependency on `bouncycastle`.

lognaturel:
> The lint problem was that ruff didn't like the space added between the two import lines in test_android_package_name.py.

Wild, when I run ruff locally on `master` now, it fails and wants to put the space back. I wonder whether I might have some kind of global config interacting with it somehow. I don't use `ruff` anywhere other than here and for `pyodk`.

> why ODK_Validate.jar more than doubled in size (from ~5MB to ~13MB)

Oh dear, I didn't think to check on size! That is highly unexpected, I will look into it.

> maybe it's possible to make some javarosa dependencies optional, so validate can build without them?

Yes, that should be possible.

lindsay-stevens:
Maybe an alias/history thing - pyodk is `ruff==0.3.4` and pyxform is `ruff==0.6.4`. I've had that happen before - a bunch of unexpected ruff changes but it was because I was running the ruff version from the other project.

lognaturel:
I'll try reinstalling all dependencies if I run into it again and let you know if I need more troubleshooting help.

I have reduced Validate size down to 3.9 MB. I'll put in a PR to update that before we release for real.

PR #723, opened 2024-09-09
Upgrade Validate to reduce file size

lognaturel:
Addresses comments at https://github.com/XLSForm/pyxform/pull/720#issuecomment-2333307738

Updates changelog in preparation for release

Verified this using https://docs.google.com/spreadsheets/d/1FOlADSaTTUOhsbeQDcU5HZrJfzseVYwBMdTIz1Skmkw/edit?gid=1068911091#gid=1068911091

This reduces Validate's size down to 3.9 MB. I think that's a good balance between benefit and time spent for now. We can come back to https://github.com/getodk/validate/issues/92

Most likely would be with the `extract-signed` function but I think the form I tried is enough.

No.


lindsay-stevens (approved):

lindsay-stevens:
Amazing!

Issue #716, opened 2024-09-03
New Feature: expose xforms-value-changed event with odk:setgeopoint action


tiritea:
#### Problem description
Desire to expose the existing ODK Collect and Enketo XForm action that permits saving a background geopoint - aka odk:setgeopoint action - in response to an xforms-value-changed event.

This functionality is documented in ODK XForms spec here, which includes an example of the desired output XML: https://getodk.github.io/xforms-spec/#setting-a-static-value-when-a-nodes-value-changes

#### Expected behavior
Per Forum discussion, the consensus appears to be adding a new XLSForm question type, perhaps called either `bg-geopoint` or `background-geopoint` (similar to how start-geopoint is used to identify a odk-instance-first-load event-triggered odk:setgeopoint action). This new question type would REQUIRE a non-null ref trigger entry in the form, much like existing setvalue-based triggers, to indicate to source question that triggers the event. Like setvalue, this trigger would be required. Unlike a regular setvalue trigger, however, the desired XForm action would be odk:setgeopoint (not setvalue) and there does NOT require an associated calculation. Indeed, the calculation column SHOULD (or MUST?) be null for this new question type, otherwise it might be unclear what the consequence of the trigger will be

#### Other information 
There is a lengthy discussion on ODK Forum here: https://forum.getodk.org/t/spec-proposal-expose-xforms-value-changed-event-with-odk-setgeopoint-action-in-xlsform/48655

tiritea:
Per ODK XForm spec, the desired output is:

```
<input ref="/data/my_text">
    <odk:setgeopoint event="xforms-value-changed" ref="/data/my_current_location" />
</input>

```
This is envisioned to appear in XLSForm as something like:

```
type                | name                | trigger | calculation
--------------------+---------------------+---------+------------
background-geopoint | my_current_location | my_text |
```

tiritea:
Per Forum discussion, there are unresolved issue around both the existing setvalue-based xforms-value-changed events - that apply equally to setgeopoint-based xforms-value-changed events - regarding the relative nesting of the trigger and target in the XML hierarchy, specifically when repeat groups are involved. These issues may result in the addition of new checks to pyxform for existing triggers which also need to be applied to this _new_ question type, in order to flag to the user semantic form issues where the specified trigger and target are 'incompatible'.

tinok:
@tiritea Should the unresolved issues be moved to a separate issue to keep this feature extension narrower?

tiritea:
[...]

PR #726, opened 2024-10-02
Add `background-geopoint` question type which exposes `xforms-value-changed` event with `odk:setgeopoint` action

RuthShryock:
Closes https://github.com/XLSForm/pyxform/issues/716

This solution introduces a new `background-geopoint` question type that triggers the `odk:setgeopoint` action when the `xforms-value-changed` event occurs. The design and implementation were based on discussions from this [forum post](https://forum.getodk.org/t/spec-proposal-expose-xforms-value-changed-event-with-odk-setgeopoint-action-in-xlsform/48655). The solution followed the proposal and was straightforward to implement based on the details in the issue.

Because this introduces a new question type and modifies how `setvalue` triggers are handled, there is a risk of breaking existing workflows that rely on the current `setvalue` behavior. I refactored some functions to handle both `setvalue` and the new `setgeopoint` triggers, however, this needs to be tested thoroughly to ensure that the changes don’t break any of the existing `setvalue` functionality.

#### Does this change require updates to documentation? 
Yes: https://github.com/XLSForm/xlsform.github.io/issues/277

- [X] run `python -m unittest` and verified all tests pass
- [X] run `ruff format pyxform tests` and `ruff check pyxform tests` to lint code

### DIFF
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -34,10 +34,17 @@ class Question(SurveyElement):
         # question type dictionary.
         if self.type not in QUESTION_TYPE_DICT:
             raise PyXFormError(f"Unknown question type '{self.type}'.")
+        # ensure that background-geopoint questions have non-null triggers that correspond to an exsisting questions
         if self.type == "background-geopoint":
             if not self.trigger:
                 raise PyXFormError(
                     f"background-geopoint question '{self.name}' must have a non-null trigger."
+                )
+            trigger_cleaned = self.trigger.strip("${}")
+            if not self.get_root().question_exists(trigger_cleaned):
+                raise PyXFormError(
+                    f"Trigger '{trigger_cleaned}' for background-geopoint question '{self.name}' "
+                    "does not correspond to an existing question."
                 )
 
     def xml_instance(self, **kwargs):
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -34,6 +34,11 @@ class Question(SurveyElement):
         # question type dictionary.
         if self.type not in QUESTION_TYPE_DICT:
             raise PyXFormError(f"Unknown question type '{self.type}'.")
+        if self.type == "background-geopoint":
+            if not self.trigger:
+                raise PyXFormError(
+                    f"background-geopoint question '{self.name}' must have a non-null trigger."
+                )
 
     def xml_instance(self, **kwargs):
         survey = self.get_root()
@@ -50,7 +55,15 @@ class Question(SurveyElement):
         if self.type == "calculate" or (
             ("calculate" in self.bind or self.trigger) and not (self.label or self.hint)
         ):
-            nested_setvalues = self.get_root().get_setvalues_for_question_name(self.name)
+            if self.type == "background-geopoint":
+                if "calculate" in self.bind:
+                    raise PyXFormError(
+                        f"'{self.name}' is triggered by a geopoint action, so the calculation must be null."
+                    )
+
+            nested_setvalues = self.get_root().get_trigger_values_for_question_name(
+                self.name, "setvalue"
+            )
             if nested_setvalues:
                 for setvalue in nested_setvalues:
                     msg = (
@@ -64,29 +77,36 @@ class Question(SurveyElement):
         xml_node = self.build_xml()
 
         if xml_node:
-            self.nest_setvalues(xml_node)
+            # Get nested setvalue and setgeopoint items
+            setvalue_items = self.get_root().get_trigger_values_for_question_name(
+                self.name, "setvalue"
+            )
+            setgeopoint_items = self.get_root().get_trigger_values_for_question_name(
+                self.name, "setgeopoint"
+            )
+
+            # Only call nest_set_nodes if the respective nested items list is not empty
+            if setvalue_items:
+                self.nest_set_nodes(xml_node, "setvalue", setvalue_items)
+            if setgeopoint_items:
+                self.nest_set_nodes(xml_node, "odk:setgeopoint", setgeopoint_items)
 
         return xml_node
 
-    def nest_setvalues(self, xml_node):
-        nested_setvalues = self.get_root().get_setvalues_for_question_name(self.name)
-
-        if nested_setvalues:
-            for setvalue in nested_setvalues:
-                setvalue_attrs = {
+    def nest_set_nodes(self, xml_node, tag, nested_items):
+        if nested_items:
+            for item in nested_items:
+                node_attrs = {
                     "ref": self.get_root()
-                    .insert_xpaths(f"${{{setvalue[0]}}}", self.get_root())
+                    .insert_xpaths(f"${{{item[0]}}}", self.get_root())
                     .strip(),
                     "event": "xforms-value-changed",
                 }
-                if not setvalue[1] == "":
-                    setvalue_attrs["value"] = self.get_root().insert_xpaths(
-                        setvalue[1], self
-                    )
+                if item[1]:
+                    node_attrs["value"] = self.get_root().insert_xpaths(item[1], self)
 
-                setvalue_node = node("setvalue", **setvalue_attrs)
-
-                xml_node.appendChild(setvalue_node)
+                set_node = node(tag, **node_attrs)
+                xml_node.appendChild(set_node)
 
     def build_xml(self):
         return None
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -34,12 +34,8 @@ class Question(SurveyElement):
         # question type dictionary.
         if self.type not in QUESTION_TYPE_DICT:
             raise PyXFormError(f"Unknown question type '{self.type}'.")
-        # ensure that background-geopoint questions have non-null triggers that correspond to an exsisting questions
+        # ensure that background-geopoint questions have triggers that correspond to an existing questions
         if self.type == "background-geopoint":
-            if not self.trigger:
-                raise PyXFormError(
-                    f"background-geopoint question '{self.name}' must have a non-null trigger."
-                )
             trigger_cleaned = self.trigger.strip("${}")
             if not self.get_root().question_exists(trigger_cleaned):
                 raise PyXFormError(
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -34,13 +34,6 @@ class Question(SurveyElement):
         # question type dictionary.
         if self.type not in QUESTION_TYPE_DICT:
             raise PyXFormError(f"Unknown question type '{self.type}'.")
-        # ensure that background-geopoint questions have triggers that correspond to an existing questions
-        if self.type == "background-geopoint":
-            trigger_cleaned = self.trigger.strip("${}")
-            if not self.get_root().question_exists(trigger_cleaned):
-                raise PyXFormError(
-                    f"background-geopoint question '{self.name}' must have a trigger corresponding to an existing question."
-                )
 
     def xml_instance(self, **kwargs):
         survey = self.get_root()
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -43,8 +43,7 @@ class Question(SurveyElement):
             trigger_cleaned = self.trigger.strip("${}")
             if not self.get_root().question_exists(trigger_cleaned):
                 raise PyXFormError(
-                    f"Trigger '{trigger_cleaned}' for background-geopoint question '{self.name}' "
-                    "does not correspond to an existing question."
+                    f"background-geopoint question '{self.name}' must have a trigger corresponding to an existing question."
                 )
 
     def xml_instance(self, **kwargs):
@@ -65,7 +64,7 @@ class Question(SurveyElement):
             if self.type == "background-geopoint":
                 if "calculate" in self.bind:
                     raise PyXFormError(
-                        f"'{self.name}' is triggered by a geopoint action, so the calculation must be null."
+                        f"'{self.name}' is triggered by a geopoint action, please remove the calculation from this question."
                     )
 
             nested_setvalues = self.get_root().get_trigger_values_for_question_name(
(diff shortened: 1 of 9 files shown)


tests/test_background_geopoint.py
```diff
@@ -0,0 +1,71 @@
...
+        self.assertPyxformXform(
+            name="data",
+            md="""
+            | survey |                    |             |                |
+            |        | type               | name        | label          | trigger   |
+            |        | integer            | temp        | Enter the current temperature |        |
+            |        | background-geopoint| temp_geo    |                | ${temp}   |
+            |        | note               | show_temp_geo | location: ${temp_geo} |        |
+            """,
+            xml__contains=[
```
lindsay-stevens, line 17: Please use `xml__xpath_match` rather then contains, [as described here](https://github.com/XLSForm/pyxform?tab=readme-ov-file#writing-tests).

pyxform/question.py
```diff
@@ -34,6 +34,18 @@ def validate(self):
         # question type dictionary.
         if self.type not in QUESTION_TYPE_DICT:
             raise PyXFormError(f"Unknown question type '{self.type}'.")
+        # ensure that background-geopoint questions have non-null triggers that correspond to an exsisting questions
```
lindsay-stevens, line 37: If possible please do question / form structure validation earlier in processing, under [`workbook_to_json`](https://github.com/XLSForm/pyxform/blob/e1ed2826da4eeeec7e909a466ba0c63e35ca0e80/pyxform/xls2json.py#L394). This allows us to include the XLSForm row reference in the error message. Unless the validation code is trivial, it should be split out to a [validator module here](https://github.com/XLSForm/pyxform/blob/master/pyxform/validators/pyxform/__init__.py).

pyxform/question.py
```diff
@@ -34,6 +34,18 @@ def validate(self):
...
+        if self.type == "background-geopoint":
+            if not self.trigger:
+                raise PyXFormError(
+                    f"background-geopoint question '{self.name}' must have a non-null trigger."
+                )
+            trigger_cleaned = self.trigger.strip("${}")
+            if not self.get_root().question_exists(trigger_cleaned):
+                raise PyXFormError(
```
lindsay-stevens, line 45: Kind of picky but I'd phrase this like the above error - leading with "background-geopoint question..." then stating the problem

pyxform/survey.py
```diff
@@ -297,8 +298,22 @@ def xml(self):
...
+            "setgeopoint": self.setgeopoint_by_triggering_ref,
+        }
+
+        return trigger_map.get(trigger_type, {}).get(f"${{{question_name}}}")
+
+    def question_exists(self, question_name: str) -> bool:
+        """
+        Check if a question with the given name exists in the survey.
+        """
+        for element in self.iter_descendants():
```
lindsay-stevens, line 313: Iterating the survey structure can be a pretty heavy operation for big forms. Then if a form has lots of background-geopoints questions, the operation is repeated many times. There are other parts of the processing pipeline that unavoidably iterate the form like this - can this check be consolidated with those checks?

tests/test_background_geopoint.py
```diff
@@ -0,0 +1,71 @@
...
+            |        | type               | name        | label          | trigger     | calculation |
+            |        | integer            | temp        | Enter the current temperature |        |             |
+            |        | background-geopoint| temp_geo    |                | ${temp}     | 5 * temp |
+            |        | note               | show_temp_geo | location: ${temp_geo} |        |             |
+            """,
+            errored=True,
+            error__contains=[
+                "'temp_geo' is triggered by a geopoint action, so the calculation must be null."
+            ],
+        )
```
lindsay-stevens, line 71: Please add test case scenarios to cover possible regressions:
- use in combination with other question types that trigger `setvalue` output
- use in groups and repeats: permuations thereof [discussed here](https://forum.getodk.org/t/spec-proposal-expose-xforms-value-changed-event-with-odk-setgeopoint-action-in-xlsform/48655)

lindsay-stevens:
Thanks for this PR - looks like it will do what was agreed earlier. My comments are mainly to polish it: increase test coverage, optimise error messages, try to avoid performance regressions / tech debt.

tests/test_background_geopoint.py
tiritea: > use in combination with other question types that trigger setvalue output

Agree. A triggering question that perform both a setvalue and odk:setgeopoint (obviously against different refs) would be prudent.

> use in groups and repeats: permuations thereof discussed here

I think its probably worth adding a testcase for a trigger and target both being inside a repeat [aside: does a regular setvalue trigger have such a testcase?].

However, the various different permutations of trigger vs target location, as discussed [elsewhere](https://github.com/XLSForm/pyxform/issues/716#issuecomment-2341730539), remains somewhat an unresolved issue _in the base XForm functionality itself_ (!). So its probably a bit premature to add a bunch of definitive tests upstream in pyxform till the precise behavior of what is/not a valid usage of xforms-value-changed actions with respect to the source control (ie trigger) and its referenced target element is formally specified somewhere [certainly the current behavior of the Collect vs Enketo clients differs today in this regard]. Once done, suitable testcases for _both_ setvalue (aka regular XLSForm 'triggers') and odk-setgeopoint (aka background-geopoint) can be added to pyxform to cover.

lindsay-stevens: I'd prefer to have tests that document pyxform behaviour which is "wrong" (by some definition, or until a spec is formalised, or hey it might be right already), than to release a feature with undefined behaviour which can randomly change (because there's no test to show what it does now and to keep it some way or the other). In future if/when the feature needs to be changed, such tests also help reduce the period of head-scratching about what it was supposed to do before, are we going to break anyone's workflows that rely on how it worked before, etc.

tiritea: The correct behavior of background-geopoint in relation to groups and repeats is basically exactly the same as it is (or should be for) setvalue triggers; or more specifically, they're both (just) xlsform-value-changed event actions on a control that set a value on an identified target ref (which is currently unconstrained...). AFAIK the finer details of what permutations hasn't been prescribed in ODK XForm spec, although this been raised and discussed: https://forum.getodk.org/t/spec-proposal-expose-xforms-value-changed-event-with-odk-setgeopoint-action-in-xlsform/48655/15

What it clear is that a trigger and target both inside the same repeat group is a valid permutation, one that is broken in Enketo but works in Collect (see https://forum.getodk.org/t/incorrect-setvalue-trigger-behavior-in-enketo-when-used-inside-repeats/49058). So a testcase for this is certainly justified.

But for the other permutatons, what is the prescribed behavior for xforms-value-changed setvalue triggers, so that @RuthShryock can code a corresponding testcase for odk:setgeopoint? @lognaturel 

Aside: I expect whatever the answer we will probably need to add appropriate pyxform tests for setvalue too.

lognaturel: > what is the prescribed behavior for xforms-value-changed setvalue triggers

Can we do what @lindsay-stevens suggested and write tests that assert the current behavior, whatever that is? 

That would involve writing the form structures described at https://forum.getodk.org/t/spec-proposal-expose-xforms-value-changed-event-with-odk-setgeopoint-action-in-xlsform/48655/18, see what pyxform converts to today, call that the expected behavior for now without thinking too much about it. That will at least give us a safety net as we follow up with future improvements in this area so we can be confident about what we do and don't change.

lognaturel:
Thanks and welcome, @RuthShryock!

tests/test_background_geopoint.py
tiritea: > write tests that assert the current behavior, whatever that is?

I'll go thru the setvalue trigger permutations that @lindsay-stevens lists [here](https://forum.getodk.org/t/spec-proposal-expose-xforms-value-changed-event-with-odk-setgeopoint-action-in-xlsform/48655/15), write a test XLSForm for each, and post the output produced by pyxform in a followup post shortly.

tiritea: > Can we do what @lindsay-stevens suggested and write tests that assert the current behavior, **whatever that is?**

The following series of XLSForms and their resulting XForms implement the various permutations listed:

1. trigger in non-repeating group A, target in non-repeating group A
2. trigger in non-repeating group A, target in non-repeating group B
3. trigger in repeating group A, target in repeating group A
4. trigger in repeating group A, target in repeating group B
5. trigger in repeating group A, target in non-repeating group B
6. trigger in non-repeating group A, target in repeating group B

I also added a #0 test for the trigger and target not being in any group or repeat. I assume there's a basic test for setvalue triggers which should already cover this, but I've explicitly listed it here so @RuthShryock  knows to include an equivalent test for a basic setgeopoint trigger.

0. trigger not in a group, target not in a group

[TestTrigger1.xlsx](https://github.com/user-attachments/files/17320473/TestTrigger1.xlsx)
[TestTrigger3.xlsx](https://github.com/user-attachments/files/17320474/TestTrigger3.xlsx)
[TestTrigger2.xlsx](https://github.com/user-attachments/files/17320475/TestTrigger2.xlsx)
[TestTrigger0.xlsx](https://github.com/user-attachments/files/17320476/TestTrigger0.xlsx)
[TestTrigger4.xlsx](https://github.com/user-attachments/files/17320477/TestTrigger4.xlsx)
[TestTrigger5.xlsx](https://github.com/user-attachments/files/17320478/TestTrigger5.xlsx)
[TestTrigger6.xlsx](https://github.com/user-attachments/files/17320479/TestTrigger6.xlsx)

I've run all these XLSForms thru https://getodk.org/xlsform/ to obtain resulting XForms. The XML files for each are in attached zip:
[Archive.zip](https://github.com/user-attachments/files/17320533/Archive.zip)

All the form permutations are successful translated, none of them currently throw any pyxform errrors, and the resulting XForms load and can be previewed in Enketo.

tiritea: In the above, if you replace the existing XForm trigger action `<setvalue...>` with `<odk:setgeopoint...>`, and remove the `<value=...>` (aka the setvalue trigger calculation) from each, you should pretty much have what the desired background-geopoint XML should look like. ie

```
<input ref="/data/temp">
    <label>Enter the current temperature</label>
    <setvalue ref="/data/temp_geo" event="xforms-value-changed" value="&quot;1.0 2.0 3.0 4.0&quot;"/>
</input>
```

becomes

```
<input ref="/data/temp">
    <label>Enter the current temperature</label>
    <odk:setgeopoint ref="/data/temp_geo" event="xforms-value-changed"/>
</input>

```

RuthShryock: @tiritea Thank you so much for the test cases! Will be adding more tests soon :)

tiritea: Looks like you have all the permutations covered, thanks.

FYI the 'show_XXX' notes arent strictly necessary to tst the core functionality, but there's probably no particular harm in having them (and it means if somebody wanted they could copy the XLSForm defintion from the test code and run it manually under, say, Enketo)

lognaturel:
@lindsay-stevens I've only taken a quick look and it's looking good. Could you please do a final review, especially looking at the perf concerns you had brought up? 

Maybe if there are smaller things pending we can file an follow-on issue as we sometimes do. Thanks!

tests/test_background_geopoint.py
lindsay-stevens: Test assertions were changed to xml__xpath_match in https://github.com/XLSForm/pyxform/pull/726/commits/3b7bde3ad6a38c9002c4b0e3ea08e08473f62cd9

pyxform/question.py
lindsay-stevens: Moved validations to workbook_to_json and added validator module in https://github.com/XLSForm/pyxform/pull/726/commits/705174b45d260fe83427310f16e00fa5ea28745a

pyxform/question.py
lindsay-stevens: Updated error messages to show the XLSForm row reference and suggest how to address the error https://github.com/XLSForm/pyxform/pull/726/commits/705174b45d260fe83427310f16e00fa5ea28745a

pyxform/question.py
```diff
@@ -50,7 +62,15 @@ def xml_control(self):
         if self.type == "calculate" or (
             ("calculate" in self.bind or self.trigger) and not (self.label or self.hint)
         ):
-            nested_setvalues = self.get_root().get_setvalues_for_question_name(self.name)
+            if self.type == "background-geopoint":
+                if "calculate" in self.bind:
+                    raise PyXFormError(
+                        f"'{self.name}' is triggered by a geopoint action, so the calculation must be null."
```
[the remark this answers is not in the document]
lindsay-stevens: Updated error messages to show the XLSForm row reference and suggest how to address the error https://github.com/XLSForm/pyxform/commit/705174b45d260fe83427310f16e00fa5ea28745a

tests/test_background_geopoint.py
lindsay-stevens: The test cases seem to cover the relevant permutations and regressions. I removed the `note`s in https://github.com/XLSForm/pyxform/pull/726/commits/705174b45d260fe83427310f16e00fa5ea28745a since they weren't relevant to the feature being tested or any specific edge cases of interest.

tests/test_background_geopoint.py
```diff
@@ -0,0 +1,71 @@
+from tests.pyxform_test_case import PyxformTestCase
+
+
+class BackgroundGeopointTest(PyxformTestCase):
+    """Test background-geopoint question type."""
+
+    def test_background_geopoint(self):
+        self.assertPyxformXform(
+            name="data",
+            md="""
```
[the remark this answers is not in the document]
lindsay-stevens: All is ***Aligned*** in https://github.com/XLSForm/pyxform/pull/726/commits/705174b45d260fe83427310f16e00fa5ea28745a

lindsay-stevens (approved):

lindsay-stevens:
@lognaturel I added a commit to address the remaining review feedback. Could you please take a look and merge/rebase if it looks good to you? Thanks

lognaturel (approved):

lognaturel:
Thanks, everyone!

PR #732, opened 2024-10-24
725: always produce entity forms with the offline entities spec version

### LINKED ISSUE #725: Always produce Entity forms with the offline entities version
https://github.com/XLSForm/pyxform/pull/719 introduced support for offline Entities on an opt-in basis to allow for integrated testing of all the components.

For the next pyxform release, this should be the only behavior and the opt-in setting should be removed.

lindsay-stevens:
Closes #725

- Removed the older create/update spec versions which are now unused.
- Although it looks a bit useless now, the "entity_features" code in xls2json.py and survey.py was retained since it seems likely that this will be needed in the near future as the entities spec evolves.
- Incorporated the previous opt-in test expectations for create/update entities into the respective "basic building blocks" tests.

- Pyxform will only emit entities version `2024.1.0` so if these are provided to an older version of Central (or another spec-compliant server) then they will be rejected.
- The `offline` column is now undefined for the `entities` sheet, so like other undefined columns it will now raise an error. Any forms that added the `offline` opt-in column will now need to remove it during their next form version update.

Filed an issue to add the proposed offline version description to the spec: https://github.com/getodk/xforms-spec/issues/319

### DIFF
--- a/pyxform/constants.py
+++ b/pyxform/constants.py
@@ -113,8 +113,6 @@ EXTERNAL_INSTANCES = {"calculate", "constraint", "readonly", "required", "releva
 CURRENT_XFORMS_VERSION = "1.0.0"
 
 # The ODK entities spec version that generated forms comply to
-ENTITIES_CREATE_VERSION = "2022.1.0"
-ENTITIES_UPDATE_VERSION = "2023.1.0"
 ENTITIES_OFFLINE_VERSION = "2024.1.0"
 ENTITY = "entity"
 ENTITY_FEATURES = "entity_features"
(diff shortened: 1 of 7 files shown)


lognaturel:
I've given this a quick look and didn't see any issues. I'm going to let it marinate and give it some more thought.

Let's hold off on merging a bit because when we do it will go on the staging server and could cause some confusion. Entity forms converted after this is merged will not work with Collect v2024.2 and prior.

lognaturel (approved):

lognaturel:
Looks ready to go to me! Please fix conflict and merge when you get a chance @lindsay-stevens 🙏

lognaturel:
I was planning to fix the merge conflict myself so I could do a release but I just realized that I don't have permissions to do so! @lindsay-stevens if you could do today, that would be very helpful. If you're not able to I'll merge these commits from another PR at the end of my day!

PR #734, opened 2024-10-29
724: raise an error if a pyxform reference is malformed

### LINKED ISSUE #724: Error if there's a ${ with a space or newline before the next }
A missing } causes particularly bad errors in Collect because it uses ${} notation for placeholders in outputs.

https://forum.getodk.org/t/incomplete-field-reference-in-note-returns-unclear-error/49537

lindsay-stevens:
Closes #724

Changes:

- pyxform_reference.py
  - new module to check pyxform reference syntax
  - unit test generates 2874 unique usage permutations
- xls2json.py
  - remove extra iteration of sheet data in replace_smart_quotes by combining it with clean_text_values and calling that instead
- expression.py
  - move expression parser from utils to here
  - update usages of lexer to reference a cached wrapper func instead
- use compiled regex patterns directly instead of importing re as well

The approach was to try and consistently apply pyxform reference syntax validation anywhere that references may be used in a XLSForm, ideally where the sheet data was already being iterated over, in order to minimise performance impact. Almost all sheet data was being passed through `remove_smart_quotes` and/or `clean_text_values` so I consolidated them with the new check under `clean_text_values`.

May be validating or cleaning text that wasn't being validated or cleaned before, but if an error is raised then the value is probably invalid.

No. Although on xlsform.org the concept of a pyxform reference is not addressed at all except in passing and I'm not sure even the ODK docs give that concept a name?

### DIFF
--- a/pyxform/parsing/expression.py
+++ b/pyxform/parsing/expression.py
@@ -1,13 +1,108 @@
+import re
 from collections.abc import Iterable
+from functools import lru_cache
+from typing import NamedTuple
 
-from pyxform.utils import parse_expression
+
+def get_expression_lexer() -> re.Scanner:
+    """
+    Get a expression lexer (scanner) for parsing.
+    """
+    # ncname regex adapted from eulxml https://github.com/emory-libraries/eulxml/blob/2e1a9f71ffd1fd455bd8326ec82125e333b352e0/eulxml/xpath/lexrules.py
+    # (C) 2010,2011 Emory University Libraries [Apache v2.0 License]
+    # They in turn adapted it from https://www.w3.org/TR/REC-xml/#NT-NameStartChar
+    # and https://www.w3.org/TR/REC-xml-names/#NT-NCName
+    namestartchar = (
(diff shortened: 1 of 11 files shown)


pyxform/parsing/expression.py
```diff
@@ -1,13 +1,108 @@
...
+
+    def get_tokenizer(name):
+        def tokenizer(scan, value):
+            return ExpLexerToken(name, value, scan.match.start(), scan.match.end())
+
+        return tokenizer
+
+    lexicon = [(v, get_tokenizer(k)) for k, v in lexer_rules.items()]
+    # re.Scanner is undocumented but has been around since at least 2003
+    # https://mail.python.org/pipermail/python-dev/2003-April/035075.html
```
lognaturel, line 72: "a masterpiece of work to get inspiration from, but not as a tool to give out to anybody"
-- https://mail.python.org/pipermail/python-dev/2003-April/035070.html

Feels like annoying fence-sitting! https://bugs.python.org/issue5337

lognaturel (approved):
Beautiful, thank you!!

Issue #534, opened 2021-04-22
Require labels for items in choice lists


lognaturel:
Currently, pyxform does not require labels for items in choice lists. 

For dynamic selects, I can see an argument for them not being required because users could be building a list they only want to query and won't use in a select. However, I think that's pretty rare and for those advanced users, it's not a big deal to just add a label. In v1.4, if a user did build a dynamic select with a choice missing a label **and used it in a select**, there'd be a Validate error. In v1.5, there's a crash (narrowly addressed by #535).

For static selects, I don't see any use for a choice without a label. Currently there's no validation on this at all.

CC @MartijnR @pbowen-oc

```
    def test_choices_without_labels__for_static_selects__allowed(self):
        """
        Test choices without labels for static selects. Validate will NOT fail.
        """
        self.assertPyxformXform(
            md = """
            | survey   |                    |      |       |
            |          | type               | name | label |
            |          | select_one choices | a    | A     |
            | choices  |                    |      |       |
            |          | list_name          | name | label |
            |          | choices            | 1    |       |
            |          | choices            | 2    |       |
            """,
            xml__contains=[
                '<value>1</value>'
            ],
        )

    def test_choices_without_labels__for_dynamic_selects__allowed_by_pyxform(self):
        """
        Test choices without labels for dynamic selects. Validate will fail.
        """
        self.assertPyxformXform(
            md = """
            | survey   |                    |      |       |               |
            |          | type               | name | label | choice_filter |
            |          | select_one choices | a    | A     | true()        |
            | choices  |                    |      |       |
            |          | list_name          | name | label |
            |          | choices            | 1    |       |
            |          | choices            | 2    |       |
            """,
            run_odk_validate=False,
            xml__contains=[
                '<instance id="choices">',
                '<item>',
                '<name>1</name>'
            ],
        )
```

tiritea:
what about image-only selections?

https://docs.getodk.org/form-question-types/#select-from-image-widget
or
https://docs.getodk.org/form-question-types/#select-one-from-map-widget

There's probably no particular harm in requiring the form designer to *always* add a text label to everything - even if the appearance may hide them.  But making them mandatory may result in the inability to prevent any text being displayed should they desire a purely pictorial interface for the selection (for whatever reason...); eg I dont think you can otherwise hide the label being shown for a map.

lindsay-stevens:
[...]

PR #736, opened 2024-10-31
647: include all choice lists in output

### LINKED ISSUE #630: Forbid colon as first character in XML identifiers
Colons are currently allowed as the first character of XML identifiers: https://github.com/XLSForm/pyxform/blob/master/pyxform/xlsparseutils.py#L8

This is consistent with https://www.w3.org/TR/REC-xml/#NT-NameStartChar but doesn't really make sense in real usage as described at https://stackoverflow.com/a/40446164/137744

We should probably forbid colons in the identifiers at all since we don't really support namespacing of body elements (https://github.com/XLSForm/pyxform/issues/105#issuecomment-1305977323).

### LINKED ISSUE #647: Choice lists are only included in the output if they are referenced in a select.
pyxform v1.12.0

#### Problem description
Currently, `pyxform` only includes choice lists used in selects in its output.

Convert a form that has multiple choice lists but only one of them used in a select ([example](https://docs.google.com/spreadsheets/d/1KdvDGGUGzmcQ2HRjJUARceVHLMS8YIPXy6gngS1kAvY/edit#gid=0), remove the calculate).

#### Expected behavior
My ideal would be to include lists that are referenced in an `instance` expression (see example above) and to output a warning when a choice list is ignored. I do think it's reasonable to omit choice lists that really aren't used -- I've seen folks who have a standard choices sheet with many choices that they use. The problem now is that there's a way to use choice lists that's not detected by pyxform.

lindsay-stevens:
Closes #647
Closes #630

- Allows users to include reference data for expressions via choices without needing a hidden dummy question. Includes all choices rather than parsing expressions to find references, because it seems unlikely that both a XLSForm would include completely irrelevant choices and have so many such choices that the XForm document size is a problem.
- Improves choices validation by including the row number in the error message and using a consistent structure
- Improves xml tag validation by using the parsing module rather than a separate regex. This was initially a tidy-up thing but I realised it also mostly addressed #630. Forbidding colons altogether seemed uncertain there so they're still allowed in this PR.

- If users have XLSForms with giant reference choice lists, and only a subset of which are used in the form, and the use case for that XForm has slow internet connectivity, or the choices list is just too big for Collect/Webforms, then it's a problem. It can be avoided easily, by removing choice lists that are not being used, or cut/paste them to a different sheet like "reference choices" which pyxform would ignore.
- If users were expecting the exact error messages about choices as before, well hopefully they are momentarily delighted by these new ones.
- If users have question names starting with a colon they will need to change them.

I don't think so. The problem scenario for the choices sheet is probably not relevant to most users, and I don't think the dummy question workaround was documented (except on the forum?) but for those who wanted it to work and had to go looking for the workaround, well now it just works. As for the leading colon in the name, xlsform.org says what characters names can contain and colon isn't one of them. 


pyxform/validators/pyxform/choices.py
```diff
@@ -0,0 +1,59 @@
...
+from pyxform.errors import PyXFormError
+
+INVALID_NAME = (
+    "[row : {row}] On the 'choices' sheet, the 'name' value is invalid. "
+    "Choices must have a name. "
+    "Learn more: https://xlsform.org/en/#setting-up-your-worksheets"
+)
+INVALID_LABEL = (
+    "[row : {row}] On the 'choices' sheet, the 'label' value is invalid. "
+    "Choices should have a label. "
```
lognaturel, line 11: Must, right?

lognaturel (approved):
Much delight will be had, indeed!

One wording piece I want to check my understanding of. Otherwise looks great.

lognaturel:
I agree this can close #630 and we can continue allowing : in the middle of an identifier.

tiritea:
>Much delight will be had

Hidden selects Begone! :-)

+1

pyxform/validators/pyxform/choices.py
lindsay-stevens: The reason for this check being a "should" / warning is to match existing behaviour - historically labels aren't required, but that might change per #534

lognaturel: Ohhh, thank you for introducing me to my past self. :D

PR #742, opened 2024-12-09
740: Improve complex forms performance - follow up - IANA test/update, move comment

lindsay-stevens:
Follow up on #740 feedback.

Addresses [this comment](https://github.com/XLSForm/pyxform/pull/740#discussion_r1870298934):

> Would be good to add a quick test with a longer subtag to test_language_warnings.

And [this comment](https://github.com/XLSForm/pyxform/pull/740#discussion_r1868563491):

> This comment looks like it's in the wrong place

Also updates results in performance test cases.

Detailed notes in commit message and docstrings / inline comments.

For the updater script, considered having it also pull from the URL rather than save manually, but 1) I don't expect that this tag list will require regular updates (last updated 6 years ago) for it to be worth streamlining it further, and 2) by the time another update is done or needed the relevant URL might be different. The main value of the script is having a documented and reproducible way to generate the code lists.

None

No

### DIFF
--- a/pyxform/parsing/expression.py
+++ b/pyxform/parsing/expression.py
@@ -74,7 +74,6 @@ def get_expression_lexer(name_only: bool = False) -> re.Scanner:
     return re.Scanner(lexicon)
 
 
-# Scanner takes a few 100ms to compile so use this shared instance.
 class ExpLexerToken:
     __slots__ = ("name", "value", "start", "end")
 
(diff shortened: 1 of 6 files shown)


lognaturel (approved):
Nice to see the perf improvements and good call on adding a reproducible way to build the IANA tag lists.

PR #743, opened 2024-12-10
More performance improvements

lindsay-stevens:
A few more general ideas following #740

See commit messages for details of each change. Overall a modest but worthwhile improvement. I think to get much more will require addressing inefficiencies in XML generation, which may require overriding more of of minidom's Element class and perhaps Attr as well. Also the relative repeats recursion issue remains.

Should be minimal risk, the only tests changes were to update the performance test stats.

No

### DIFF
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -3,7 +3,7 @@ XForm Survey element classes for different question types.
 """
 
 import os.path
-from collections.abc import Iterable
+from collections.abc import Callable, Generator, Iterable
 from itertools import chain
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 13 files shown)


pyxform/question.py
```diff
@@ -391,6 +391,22 @@ def validate(self):
             for child in self.children:
                 child.validate()
 
+    def iter_descendants(
```
lognaturel, line 388: Maybe this implementation could go in some sort of shared place? Everyone's favorite, a util file? I'm worried about these identical implementations drifting apart...

lognaturel (approved):
Nice! 🏃‍♀️

pyxform/question.py
lindsay-stevens: Agree 3x copy/pastas = probably time for a shared func. I didn't do that here because of the points noted in #745. In which case the only copy of this method override for `iter_descendants` would be MultipleChoice.

Issue #749, opened 2025-01-27
Choices have reference to the last parent in the event two or more questions use the same choices


ukanga:
The generated XML is as expected but if you traverse the Survey object and request a choice parent on the first occurrence, it will point to the last question. In the example below, the questions `fruita` and `fruitz` use the same choices. With the current implementation, `fruita.children[0].parent == fruitz` instead of `fruita`.

```markdown
| survey |
|        | type                   | name   | label   |
|        | begin group            | a      | Group A |
|        | select one fruits      | fruita | Fruit A |
|        | select one fruity      | fruity | Fruit Y |
|        | end group              |        |         |
|        | begin group            | b      | Group B |
|        | select one fruits      | fruitz | Fruit Z |
|        | select_multiple fruity | fruitb | Fruit B |
|        | end group              |        |         |
| choices |
|         | list name | name   | label  |
|         | fruits    | orange | Orange |
|         | fruits    | mango  | Mango  |
|         | fruity    | orange | Orange |
|         | fruity    | mango  | Mango  |
        
```

PR #750, opened 2025-01-27
fix: ensure choices/option parent is set correctly

ukanga:
Closes https://github.com/XLSForm/pyxform/issues/749

Restores the previous behaviour where traversing choices always point to the parent they are assigned to. 
The deep copy is at the last possible place when the choices are assigned to a question before the parent is set.

No

### DIFF
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -2,6 +2,7 @@
 XForm Survey element classes for different question types.
 """
 
+import copy
 import os.path
 from collections.abc import Callable, Generator, Iterable
 from itertools import chain
@@ -372,7 +373,7 @@ class MultipleChoiceQuestion(Question):
         kw_children = kwargs.pop(constants.CHILDREN, None)
         choices = coalesce(kw_choices, kw_children)
         if isinstance(choices, tuple) and isinstance(next(iter(choices)), Option):
-            self.children = choices
+            self.children = copy.deepcopy(choices)
         elif choices:
             self.children = tuple(
                 Option(**c) for c in combine_lists(kw_choices, kw_children)
--- a/tests/test_builder.py
+++ b/tests/test_builder.py
@@ -8,9 +8,14 @@ from pathlib import Path
 from unittest import TestCase
 
 import defusedxml.ElementTree as ETree
-from pyxform import InputQuestion, Survey
-from pyxform.builder import SurveyElementBuilder, create_survey_from_xls
+from pyxform import InputQuestion, MultipleChoiceQuestion, Survey
+from pyxform.builder import (
+    SurveyElementBuilder,
+    create_survey_element_from_dict,
+    create_survey_from_xls,
+)
 from pyxform.errors import PyXFormError
+from pyxform.question import Option
 from pyxform.xls2json import print_pyobj_to_json
 
 from tests import utils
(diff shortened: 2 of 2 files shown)


lognaturel:
Thanks! Can you please check the import order for the linter?

lindsay-stevens:
Sorry but I'm strongly against this change. Many performance and correctness issues have been due to dict-like behaviour in pyxform and this change propagates that assumption, making it more difficult to transition to classes. Unnecessarily performing copies or deepcopies of dicts, either actual dicts (e.g. the `builder` module) or dict-like objects (e.g. `SurveyElement`s), is the kind of thing I've worked hard to systematically eliminate over the past few months. In relation to issue #749 this is about internals not exposed in `xls2xform`, and it doesn't affect XForm conversion correctness.

Anyway, navigating SurveyElements via parent/child for questions and choices will not be compatible usage soon. I identified this parent reference issue a few weeks ago and it is addressed in #746 (mainly commit 6927767) by using an Itemset class and changing the SurveyElement structures to remove the parent-child relationship for Questions and Choices. In other words, a choice list doesn't really "belong" to a Question in any sense that is meaningful to generating the XForm: a choice belongs to an Itemset and an Itemset belongs to Survey; and one or more Questions can reference each Itemset. In the context of the generated XForm, this corresponds to the change from choices being copied into the question and always being in the `body` with the `select`, to the current behaviour of model-level secondary instance elements for choices with references to those instances from `select`s (except for some weirdness to accommodate the deprecated `search` appearance). A similar duplication/unnecessary reference should soon be removed from xls2json `add_choices_info_to_question` which has been there for about 18 months for the backwards compatibility of external code, but slows down every XForm conversion and complicates the internal code.

External code that wants to use these pyxform internals should reference the choices on the `Survey.choices` dict using the `MultipleChoiceQuestion.itemset` name as the key, instead of assuming a copy of the choices exists in `MultipleChoiceQuestion.children` or `.choices`; and do the equivalent if using the internal dict structure emitted from `workbook_to_json`, or any `SurveyElement.to_json_dict`.

ukanga:
@lindsay-stevens Can we then remove the parent reference that currently exists with choices? The reference exists, hence the assumption that it is broken, but if the expected behaviour is that it should not be aware of the parent, then we should make it so.

ukanga:
@lindsay-stevens, how do you access the (rectangular) XLSForm formats? Assuming you have an Excel file representing the XLSForm, you plan not to reread it once you process it but would want to persist a representation of it and make some modifications.

I mainly use pyxform as a library, and such representation is quite helpful to the functionality we have built using this library. Losing such changes will result in a need for a tremendous rewrite of an 11-year-old codebase.

lindsay-stevens:
By rectangular XLSForm formats I'm referring to XLSX, XLS, CSV, and MD. Is this input data not retained by your app? Why cannot the modifications be done in-memory on the `SurveyElement` structures rather than their dict copies? What kinds of modifications are being made? Where and how is the intermediate representation persisted? How is the representation conversion resumed e.g. where in the pipeline is this modified data fed back in to pyxform?

I don't know about your use case but I hope you can appreciate that it will be nearly impossible to non-disruptively maintain pyxform against unknown requirements placed on its internal data structures and behaviour. I think the solution will involve discussion about what exactly your app is doing, what it's needs and assumptions are, and how that might fit into an explicitly supported use case for pyxform. In the meantime `to_json_dict` still works but if they are to be retained I think the above should be resolved.

ukanga:
The output of `survey.to_json()` is persisted in the DB. Later on, days/weeks/months, the user can make form title and form version changes without re-uploading an updated XLSForm (xlsx) file. This avoids the round trip of the user downloading the XLSForm they uploaded, making modifications to it, and then re-uploading.

lognaturel:
My apologies, I took a quick look at this and didn't think through the implications and the broader work in progress!

As far as I can tell, there was never an effort to document or clarify what should be intended as a stable interface to pyxform beyond the XML output for clients. As @lindsay-stevens describes, that has lead to different systems making different assumptions around what should be considered public vs private. As we work to improve performance and make some aspects of the system easier to evolve, we're bound to break some integrators, and we're trying to signal that with major release bumps and discussion in PRs as they come in. Please join those conversations!

@ukanga did you close this because you've found a satisfactory solution for now? Given that you are updating specific, well-known fields, maybe you don't need pyxform to make those changes? Or maybe a public interface specifically to make those changes could be provided? I'm not seeing the connection between choices and the needs you've mentioned, though!

lognaturel:
> This has been manageable and was almost done.

Thanks, that's great to hear! Do keep us posted on any other issues you may run into.

> identifying and fixing tests on onadata

Should some of those tests be in pyxform instead?

> For now, I will pin us to https://github.com/XLSForm/pyxform/releases/tag/v3.0.0 + https://github.com/XLSForm/pyxform/commit/6524d17d167a4458d85598856f0a11942803caa6 which is not as chaotic for us for the time being

That sounds great. @lindsay-stevens are syncing up shortly and will discuss how to help accommodate your usage. Do you actually use OMK still? My understanding was that the app had been long deprecated and that `select_one_from_map` in ODK Collect replaces most desired usage.

ukanga:
> That sounds great. @lindsay-stevens are syncing up shortly and will discuss how to help accommodate your usage. Do you actually use OMK still? My understanding was that the app had been long deprecated and that `select_one_from_map` in ODK Collect replaces most desired usage.

It's okay to remove OMK. I am not aware of any current usage. However, we do have plenty of data collected with OSM. None of it is active anymore, but I may need to check. There are no reservations about removing the functionality; the use case evolved, and the `select_one_from_map` is only one way of the different ways it evolved.

Issue #752, opened 2025-01-28
AttributeError: OsmUploadQuestion includes SELECT_QUESTION_EXTRA_FIELDS


ukanga:
The `OSMUploadQuestion` includes `SELECT_QUESTION_EXTRA_FIELDS` which results in an exception when calling the `to_json_dict()` function for the question. This raises AttributeError exception.

```sh
python -m unittest tests.test_osm                                                                                               (pyxform)  ✘ 1 master ✱
.E.
======================================================================
ERROR: test_osm_type_to_json_dict (tests.test_osm.OSMWidgetsTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ukanga/ona/pyxform/tests/test_osm.py", line 64, in test_osm_type_to_json_dict
    self.assertEqual(osm_question.to_json_dict(), {})
  File "/home/ukanga/ona/pyxform/pyxform/question.py", line 234, in to_json_dict
    result = super().to_json_dict(delete_keys=to_delete)
  File "/home/ukanga/ona/pyxform/pyxform/survey_element.py", line 284, in to_json_dict
    result = self.copy()
  File "/home/ukanga/ona/pyxform/pyxform/survey_element.py", line 276, in copy
    return {k: self[k] for k in self}
  File "/home/ukanga/ona/pyxform/pyxform/survey_element.py", line 276, in <dictcomp>
    return {k: self[k] for k in self}
  File "/home/ukanga/ona/pyxform/pyxform/survey_element.py", line 54, in __getitem__
    return self.__getattribute__(key)
AttributeError: 'OsmUploadQuestion' object has no attribute 'itemset'

----------------------------------------------------------------------
Ran 3 tests in 0.005s

FAILED (errors=1)

```

This line might be including a typo. I believe what was meant was to use `OSM_QUESTION_EXTRA_FIELDS`.

ukanga:
Perhaps not necessary due to the discussion to deprecate the functionality - https://github.com/XLSForm/pyxform/issues/745.

lindsay-stevens:
Closing as it seems the issue was a dev bug that was fixed. The above linked fix PRs for pyxform proposed to add a missing field `constants.CHILDREN` to the `OSM_QUESTION_FIELDS` collection, and this was incorporated via another PR. Similarly the onadata PR for schema migration seems to have been closed with no changes needed. Please re-open though if it is still a bug / blocker. Thanks

PR #753, opened 2025-01-28
fix: OsmUploadQuestion fields

ukanga:
Closes https://github.com/XLSForm/pyxform/issues/752

No.


lindsay-stevens:
I agree there's a typo there but please:

- add a test using PyxformTestCase. If an error can't be triggered that way then the bug is not user-facing. It's not necessarily worthwhile making the code resilient to / accommodating of invalid or unsupported internal API usages. That is, the current test is baking in an assumption that it's OK to pass in that particular data structure and it can be expected from `workbook_to_json`.
- base the PR on the branch in #746 so I don't have to rebase it repeatedly - assuming there will be other PRs incoming soon as well. Some things may be fixed there already, for example the OSM tests pass on that branch.

Thanks

ukanga:
The JSON/Dict is a public API to us and not internal one. We are not just using the output as XML, XML is only relevant where ODK Collect/Enketo/Other form players are concerned. But when it comes to other manipulations not limited to what I have stated [here](https://github.com/XLSForm/pyxform/pull/750#issuecomment-2618925434) then the "internal" API's are very relevant to us.

ukanga:
@lindsay-stevens should I do a PR on your repo, or will you pick the changes from here?

ukanga:
@lindsay-stevens Made a PR against your branch here  https://github.com/lindsay-stevens/pyxform/pull/2.

Issue #760, opened 2025-02-27
BUG: big-image without image does not throw error on choice option


tiritea:
Specifying a big-image media type in the main survey form without a corresponding image will (correctly) result in a fatal error:

"To use big-image, you must also specify an image for the survey element named {self.name}."

You can use this form to check:
[BigImage_error1.xlsx](https://github.com/user-attachments/files/18999325/BigImage_error1.xlsx)

However, no such check appears to be being made when you specify a big-image without an image for a choice option; no error is reported and pyxform will happily create a form with a big-image media type but no image for these options. This form will reproduce the problem:
[BigImage_error2.xlsx](https://github.com/user-attachments/files/18999357/BigImage_error2.xlsx)

which generates a XML XForm with:

```
<?xml version="1.0"?>
<h:html
	xmlns="http://www.w3.org/2002/xforms"
	xmlns:h="http://www.w3.org/1999/xhtml"
	xmlns:ev="http://www.w3.org/2001/xml-events"
	xmlns:xsd="http://www.w3.org/2001/XMLSchema"
	xmlns:jr="http://openrosa.org/javarosa"
	xmlns:orx="http://openrosa.org/xforms"
	xmlns:odk="http://www.opendatakit.org/xforms">
	<h:head>
		<h:title>Big Image (missing choices image)</h:title>
		<model odk:xforms-version="1.0.0">
			<itext>
				<translation lang="default" default="true()">
					<text id="animals-0">
						<value>whale</value>
						<value form="big-image">jr://images/a.jpg</value>
					</text>
					<text id="animals-1">
						<value>frog</value>
						<value form="big-image">jr://images/b.jpg</value>
					</text>
					<text id="animals-2">
						<value>crocodile</value>
						<value form="big-image">jr://images/c.jpg</value>
					</text>
					<text id="animals-3">
						<value>eagle</value>
						<value form="big-image">jr://images/d.jpg</value>
					</text>
...

```

tiritea:
Please note, this may be fixed with PR for https://github.com/XLSForm/pyxform/issues/759, which has to include an almost identical check on choice options for the proposed new `image-description` form type, which also prerequisites having an `image`.

Issue #759, opened 2025-02-26
add support for new 'image-description' XForm form media type


tiritea:
As discussed on [ODK Forum](https://forum.getodk.org/t/placeholder-alt-text-for-images-eg-screen-readers/50868/20), this is to add XLSForm support for a adding a new `form` type to ODK XForm spec to allow a (translatable) string description to be assocaited with media image, ostensible for the purpose of supporting such things as screen readers (eg Enketo, Web Forms) but possibly also to provide dynamic form content descriptors in Collect.

This new string field can be mostly just passed straight (and localizable) thru pyxform today as just another supported 'media' type without significant changes., provided the new type (and there seems a reasonable consensus to call it `image-description`) is added to the list of supported medi types. As with the existing `big-image` media type, a new `image-description` media type should prerequisite the existence of a corresponding `image`; that is, you cannot specify `image-description` (or for that mater `big-image`) without already specifying an `image`. However, the converse is not true; you can still specify an `image` without an `image-description` (or `big-image`).

As discussed, it also seems reasonable to provide a suitable shortcut/alias in XLSForm for the full `media::image-description(::lang),` eg `image-description` or `image-description::French (fr)`

The proposed XLSForm extensions and expected resulting XML XForm definitions are:

[ImageDescription_Proposed.xlsx](https://github.com/user-attachments/files/18999184/ImageDescription_Proposed.xlsx)

[ImageDescriptionTranslated_Proposed.xlsx](https://github.com/user-attachments/files/18999186/ImageDescriptionTranslated_Proposed.xlsx)

[ImageDescription_Proposed.xml.zip](https://github.com/user-attachments/files/18999188/ImageDescription_Proposed.xml.zip)

[ImageDescriptionTranslated_Proposed.xml.zip](https://github.com/user-attachments/files/18999203/ImageDescriptionTranslated_Proposed.xml.zip)

Providing an `image-description` without specifying an `image`, either in the main survey or when associated with a choice list option, should result in a suitable fatal error, much the same as happens today if you specify a `big-image` without an `image` in the survey:

> pyxform.errors.PyXFormError: To use big-image, you must also specify an image for the survey element named {self.name}.

PR #761, opened 2025-03-04
Add support for new form type image-description

tiritea:
Closes #https://github.com/XLSForm/pyxform/issues/759
Closes #https://github.com/XLSForm/pyxform/issues/760

This solution and others considered were discussed extensively on the ODK Forum. See  https://forum.getodk.org/t/placeholder-alt-text-for-images-eg-screen-readers/50868

This adds new new (optional) translatable form type, to the existing `image`, `big-image`, `audio` and `video` media types. It should not impact the existing functionality of these existing media types. XForm clients that do not support or recognize the new `image-description` form type can safely ignore it with no impact on existing XForm functionality.

This PR also adds a new testcase for a choice option with a `big-image` specified but without a corresponding `image`, because it is almost identical to the check now needed for `image-description`. This error condition is currently *not* being checked, and results in an invalid XForm being generated [note, this image prerequisite is being checked in the main survey, but not for choice options. It was probably overlooked]. The new `big-image` check will mean previously invalid XLSForms with this inconsistency will now fail to be processed, but I believe this is actually the desired behavior.

Yes, this new ODK form type will need to be documented in the XLSForm Docs.
See https://github.com/XLSForm/xlsform.github.io/issues/279

- [ ] verified that any code or assets from external sources are properly credited in comments - **Not Applicable**

#### Other important notes:
This is my first pyxform PR; please be gentle... :-)
Attn: @RuthShryock


lognaturel (changes requested):
Thanks for this! The implementation looks good to me but the tests should not introduce real XLSX files or xml files 😬 We've tried to document this at https://github.com/xlsform/pyxform?tab=readme-ov-file#writing-tests and all new tests follow that approach. Hopefully it will be a relatively quick conversion to that format.

@lindsay-stevens please let me know whether you'd like to have a look too or whether you're ok with my review. Thanks!

tiritea:
>We've tried to document this..

Sorry, I'd completely missed that! Will do.

lognaturel:
We just merged https://github.com/XLSForm/pyxform/pull/746 which causes some conflicts. If you can address them, that'd be great. But I can also take over the PR once you've updated the tests and do that part.

tiritea:
It looks one of the conflicts is due to the parallel change that happened regarding you specify aliases and survey headers. So the new 'image-description' now just needs to be reformatted to match.

lognaturel:
Closing for now, feel free to re-open to complete.

PR #764, opened 2025-04-30
fix: no labels visible for randomized selects with translations

lindsay-stevens:
Per report in https://forum.getodk.org/t/54492

- regression seems to be from #746 where the choices may be referenced by the survey only and the select question does not have a reference. This happens if the question is randomized, or external select, or search appearance.
  - this would suggest that maybe the choices should be in-line for randomized choices - however the existing tests for randomize expect an itemset in the body. Also, these new tests pass on tag 3.0.1 (prior to #746), and fail on master (e0b350b5)
  - since the user noticed this via https://getodk.org/xlsform/ and it seems that the converter uses https://staging.getodk.cloud I assume that means master is deployed to staging (i.e. not just tagged releases)
  - The form posted in the forum converts as expected (choice labels shown in enketo preview) without warnings/errors when uploaded to a project on https://staging.getodk.cloud

As mentioned above I wasn't totally sure that these combinations are supported, considering https://github.com/enketo/enketo/issues/322 but it seems like it works in Collect (per https://forum.getodk.org/t/42506/4) and presumably will in Webforms too. Also I wasn't able to locally reproduce the ODK Validate warnings mentioned in the reporting forum thread but maybe the Validate version on xlsform-online is out of sync?

This should be a fix since the existing tests pass and the new code satisfies tests that worked on earlier code (as described above), and the code committed in between didn't intend to change randomize or choice_filter behaviour for users.

No

### DIFF
--- a/pyxform/question.py
+++ b/pyxform/question.py
@@ -370,6 +370,11 @@ class MultipleChoiceQuestion(Question):
             raise PyXFormError("""Invalid value for `self.bind["type"]`.""")
 
         result = self._build_xml(survey=survey)
+        choices = None
+        if survey.choices:
+            choices = survey.choices.get(self.itemset, None)
+        if not choices:
+            choices = self.choices
 
         # itemset are only supposed to be strings,
         # check to prevent the rare dicts that show up
@@ -390,7 +395,7 @@ class MultipleChoiceQuestion(Question):
 
             if file_extension in EXTERNAL_INSTANCE_EXTENSIONS:
                 pass
-            elif self.choices and self.choices.requires_itext:
+            elif choices and choices.requires_itext:
                 itemset = self.itemset
                 itemset_label_ref = "jr:itext(itextId)"
             else:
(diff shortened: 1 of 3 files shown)


lognaturel (approved):
Thanks for spotting this on the forum and for the speedy fix!

lognaturel:
> maybe the choices should be in-line for randomized choices

Do you mean directly as children of the question? That wouldn't work with the way randomization currently works -- it relies on an `itemset` declaration that references a secondary instance being able to include a function call. It's possible to randomize nodes from external secondary instances and even repeats!

> master is deployed to staging

Yes, exactly. It's to catch issues like this one as early as possible. Should we add this to the readme, maybe? The only place I can think of where it's documented is [here](https://docs.google.com/document/d/1bJ40w6VyugVcC1JlC_PyQX6Wcvjpp74Bg82D3yTO-wk/edit?tab=t.0#heading=h.u506283eq3tl) (ODK internal)

> when uploaded to a project on https://staging.getodk.cloud/

Ahhh, we have various staging servers. 🙈 That one reflects the latest Central but doesn't also update pyxform. The user was using the pyxform staging server.

PR #768, opened 2025-06-03
756: support for / test with python 3.13

### LINKED ISSUE #756: Add Python 3.13 to tests and supported versions
Align with pyodk: https://github.com/getodk/pyodk/pull/108

lindsay-stevens:
Closes #756

To align with pyodk python version support. According to the [python website](https://devguide.python.org/versions/), 3.14 is due out 2025-10 and 3.10 goes EOL 2026-10.

The XML escaping algo changed in python 3.13 to what seems to be a more correct way (referring to [normalization rules in the XML spec](https://www.w3.org/TR/REC-xml/#AVNormalize)) - namely, to escape `\r\n\t` in attribute values. So pyxform now does that too, for all supported python versions. Pyxform already was overriding some escaping behaviour, because prior to python 3.13 double quotes were escaped in element text but they only need to be escaped in attributes. Unfortunately python implemented the change in a [non-backwards compatible way](https://github.com/python/cpython/blame/3.13/Lib/xml/dom/minidom.py#L303), by adding a new required `attr` argument, with no default, to the `_write_data` func that pyxform was using. In any case pyxform would have needed to do something to maintain compatibility with 3.10 through 3.13 since the new output is different.

The regression risk then is perhaps tab or newlines characters might have some kind of impact on Collect/Webforms/Enketo behaviour e.g. if they were normalised out before and now will be preserved due to escaping, or if any of the form clients don't recognise the escape codes and suddenly users see the escape codes instead of whitespace or nothing. There is the workaround for users to remove these characters from the XLSForm but generally pyxform tries to smooth over friction points like that. If these regression scenarios are a concern we could do some testing to check, and/or not escape `\r\n\t` in attributes in pyxform, or not emit `\r\n\t` values in attributes, etc.

The dev dependency update to formencode drops support for python 3.7 and 3.8 but that should not affect users. Could be an issue for library users but it's difficult to avoid since the update is required for python 3.13 compatibility. The alternative is to remove this dev dependency, but that would require refactoring all of the remaining "old style" tests that use formencode for XML document comparison.

Updated readme.

### DIFF
--- a/pyxform/utils.py
+++ b/pyxform/utils.py
@@ -12,7 +12,7 @@ from io import StringIO
 from itertools import chain
 from json.decoder import JSONDecodeError
 from xml.dom import Node
-from xml.dom.minidom import Element, Text, _write_data
+from xml.dom.minidom import Element, Text
 
 from defusedxml.minidom import parseString
 
@@ -29,6 +29,8 @@ PYXFORM_REFERENCE_REGEX = re.compile(r"\$\{(.*?)\}")
 SPACE_TRANS_TABLE = str.maketrans({" ": "_"})
 XML_TEXT_SUBS = {"&": "&amp;", "<": "&lt;", ">": "&gt;"}
 XML_TEXT_TABLE = str.maketrans(XML_TEXT_SUBS)
+XML_ATTR_SUBS = {'"': "&quot;", "\r": "&#13;", "\n": "&#10;", "\t": "&#9;"}
+XML_ATTR_TABLE = str.maketrans(XML_ATTR_SUBS)
 
 
 class DetachableElement(Element):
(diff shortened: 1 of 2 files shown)


lognaturel (dismissed):

lognaturel (approved):
Breaking Python 3.7 and 3.8 feels like no problem.

I really doubt tabs and newlines are common in attributes. The only context in which I think they could make sense is in constraint and required messages. Let’s go with this and if there are any issues we will learn something.

lognaturel:
I was wrong! Some users do use newlines in calculations and constraints for readability. I should have thought of that because I do it too. Those aren't typically a problem but there seem to be cases in which they crash Enketo: https://github.com/XLSForm/pyxform/issues/771

PR #772, opened 2025-06-14
771: restore pre-py3.13 xml attribute escaping behaviour for \r\n\t

### LINKED ISSUE #771: Some forms with escaped newlines crash Enketo
Following https://github.com/XLSForm/pyxform/pull/768, we've had reports of two forms that don't load correctly in Enketo. One errors on launch, the other has a select in which all choices are displayed without labels.

The issue has to do with escaped newlines because that's the only escaped character the forms contain. We've asked for the form source to try to find the root cause.

Escaped newlines work fine in Enketo, Collect, ODK Web Forms for everything I've tried so far: [multiline-attributes.xlsx](https://github.com/user-attachments/files/20684831/multiline-attributes.xlsx)

lindsay-stevens:
Closes #771

In py3.13, minidom started escaping `\r\n\t` in attributes in order to comply with the XML spec, which pyxform adopted for all python versions in an earlier commit. But it was found that this change caused a HTTP 500 error for Enketo (not Collect/Webforms). If a user had choice filter in a multi-language form, and the filter expression used `select()`, then Enketo could not display the form but displayed an error saying "Cannot read properties of undefined ( reading length)". Details in XLSForm/pyxform/#771 - many test cases were explored such as constraint, calculate, relevant, single or multi-language forms, choice filter expressions and operators, but only the one mentioned above seemed to trigger the error. Since it seems a reasonably common trigger (using newlines to format expressions, and using `selected()` in choice filters, within a multi-language form), and it may take some time for both Enketo to fix the issue and for that release to be adopted by users, pyxform reverts to the previous behaviour.

It seems like there are a few XForm generator tools out there so if they adopt python 3.13 they may hit the same issue, but that is not really a pyxform problem as such - they could follow/comment/fix https://github.com/enketo/enketo/issues/1408

No, it seems a bit too detailed an unlikely an issue to mention anywhere that pyxform isn't escaping `\r\n\t` in attribute values. If needed, a forum post could reference this PR and it's linked issues.

### DIFF
--- a/pyxform/utils.py
+++ b/pyxform/utils.py
@@ -29,8 +29,6 @@ PYXFORM_REFERENCE_REGEX = re.compile(r"\$\{(.*?)\}")
 SPACE_TRANS_TABLE = str.maketrans({" ": "_"})
 XML_TEXT_SUBS = {"&": "&amp;", "<": "&lt;", ">": "&gt;"}
 XML_TEXT_TABLE = str.maketrans(XML_TEXT_SUBS)
-XML_ATTR_SUBS = {'"': "&quot;", "\r": "&#13;", "\n": "&#10;", "\t": "&#9;"}
-XML_ATTR_TABLE = str.maketrans(XML_ATTR_SUBS)
 
 
 class DetachableElement(Element):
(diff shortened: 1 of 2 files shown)


lindsay-stevens:
@ln yes the custom implementations are there to achieve output that is consistent / compliant XML / compatible with Enketo across python versions. Even if there was no Enketo issue, custom implementations would still have been needed:

- element text: don't escape double quotes (they are in < py3.13 standard lib, but are not >= py3.13).
- attributes: don't escape `\r\n\t` (they are not in < py3.13 standard lib, but are >= py.3.13).

Also the standard library escaping algo does repeated searches/calls like `if token in string: string.replace()` while pyxform tries to do less operations with `if any(t in string for t in tokens): string.translate()`.

PR #782, opened 2025-09-17
781: handle None from fullmatch reference regex

### LINKED ISSUE #781: Error when converting form with `count-selected()` in `repeat_count`
Error: 'NoneType' object has no attribute 'group'

I've verified that this doesn't happen with pyxform v4.0.0

Forum report with sample form: https://forum.getodk.org/t/error-converting-xlsform-to-xml-using-odk-xlsform-online-count-selected/56741

lindsay-stevens:
Closes #781

- this code branch is for `is_pyxform_reference` but it didn't handle the case where there is no match (looks like reference but has other characters) - in which case the iteration over `outer_matches` would fail and throw an error like "None has no group"
- added test coverage for relevant usages of is_pyxform_reference:
  - repeat_count, now with tests showing common usage pattern outcomes
  - seed param, which had alternate usages already (constant/single ref)

Fixes a regression, and tries to avoid others of a similar nature by adding more tests.

No

### DIFF
--- a/tests/test_repeat_count.py
+++ b/tests/test_repeat_count.py
@@ -1,0 +1,243 @@
+from tests.pyxform_test_case import PyxformTestCase
+from tests.xpath_helpers.questions import xpq
+
+
+class TestRepeatCount(PyxformTestCase):
+    """
+    Test usages of the survey repeat_count column.
+    """
+
+    def test_single_reference__generated_element_same_name__ok(self):
+        """Should not have a name clash, the referenced item should be used directly."""
+        md = """
+        | survey |
+        | | type         | name     | label | repeat_count |
+        | | integer      | q1_count | Q1    |              |
+        | | begin repeat | r1       | R1    | ${q1_count}  |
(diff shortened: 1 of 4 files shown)


tests/test_repeat_count.py
```diff
@@ -0,0 +1,243 @@
...
+                xpq.body_control("q1", "input"),
+                # no instance element for generated *_count item
+                """
+                /h:html/h:head/x:model/x:instance/x:test_name[not(./x:r1_count)]
+                """,
+                # no binding for generated *_count item
+                """
+                /h:html/h:head/x:model[not(./x:bind[@nodeset='/test_name/r1_count'])]
+                """,
+                # repeat references existing count element directly.
```
lognaturel, line 70: Seems like we can close https://github.com/XLSForm/pyxform/issues/435?

lindsay-stevens: yes I think so, I closed it with a [comment here](https://github.com/XLSForm/pyxform/issues/435#issuecomment-3305574903)

PR #786, opened 2025-11-19
fix: remove warning for groups without label

### LINKED ISSUE #785: Remove warning for group without label
Currently, if a group has no label, a warning is produced. However, it's pretty common to use groups for organization and not just display. In that case, there's no value to a label. Whenever a user brings up that warning, it's to say that it's confusing. Recent example: https://forum.getodk.org/t/support-for-fields-groups-without-label-hint/57111

Any likely downsides? Do others see value in the warning?

jkuester:
Closes #785

This might be premature as I see `#785` is still labeled with "needs discussion". :sweat_smile: However, as the code was already fresh in my mind, I figured I would raise a PR to provide some additional impetus  to the discussion as well as a concrete proposal to discuss.  

The rational for this change is covered in detail [on the forum thread](https://forum.getodk.org/t/support-for-fields-groups-without-label-hint/57111), but the TLDR is:

- There are valid use-cases for groups without labels
- These use cases are supported by the spec
- Pyxform currently allows for transforming groups without a label/hint (but a warning is printed)
- Currently, the warning can be avoided on a group with not label/hint by including the `field-list` appearance (but that appearance cannot be combined with any other appearance or the warning will re-appear). However, semantically it is meaningless to nest `field-list` groups. So, this workaround is not appropriate for all cases.

The main "risk" I see here would be that users could unintentionally add groups with no label/hint (because they did not see any warning about it).  Technically I guess this would not be a "regression" in terms of form functionality, though. 

The [ODK docs](https://docs.getodk.org/form-logic/#groups-of-questions) already note:

> Groups without labels can be helpful for organizing questions in a way that's invisible to the user. This technique can be helpful for internal organization of the form. These groups can also be a convenient way to conditionally show certain questions.

I do not think these changes require any specific modifications to the xlsform docs.

- [X] run `python -m unittest` and verified all tests pass
- [X] run `ruff format pyxform tests` and `ruff check pyxform tests` to lint code

### DIFF
--- a/tests/test_group.py
+++ b/tests/test_group.py
@@ -681,6 +681,7 @@ class TestGroupParsing(PyxformTestCase):
             ],
         )
 
+
 class TestGroupInternalRepresentations(TestCase):
     maxDiff = None
 


lognaturel:
Thanks, @jkuester! This still feels like the right move to me. @lindsay-stevens could you please take a look when you get a chance?

PR #787, opened 2025-11-28
754: fix 'NoneType is not iterable' error for group with no children

### LINKED ISSUE #754: Importing XLSForm with empty group results in confusing error
Example form attached: [2025-02 empty group bug.xlsx](https://github.com/user-attachments/files/18714538/2025-02.empty.group.bug.xlsx)

Result in v3.0.0+:

* The import fails with message The XLSForm could not be converted: 'NoneType' object is not iterable

Prior to v3.0.0, this case did not error in pyxform and instead was caught by Validate:

```
ODK Validate Errors:
Group has no children! Group: ${IsEmpty}. The XML is invalid.
```

Initially reported by @cooperka at https://github.com/getodk/central/issues/886

lindsay-stevens:
Closes #754

Also added test to document behaviour with repeats where a similar 'empty group' error is not raised
Also tidies up group/repeat related tests to make test coverage clearer

In `section.py`, considered adding `if self.children is not None` branches where necessary, but usually a Section subclass will have child elements so instead a default list is created in the `__init__` method.

If an external library subclasses SurveyElement but expects the moved children-related methods to be there, they would need to copy them over to their own implementation, or use Section as a parent class instead.

No just a bug fix.

### DIFF
--- a/tests/test_group.py
+++ b/tests/test_group.py
@@ -620,7 +620,7 @@ class TestGroupParsing(PyxformTestCase):
         """
         self.assertPyxformXform(
             md=md,
-            run_odk_validate=True,  # Error is from Validate only.
+            run_odk_validate=True,  # Error about empty groups is from Validate only.
             odk_validate_error__contains=[
                 "Group has no children! Group: ${g1}. The XML is invalid."
             ],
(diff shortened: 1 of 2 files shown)


lognaturel:
Thanks!

PR #789, opened 2025-12-08
714: Indicate which sheet a ${} replacement error comes from

### LINKED ISSUE #714: Indicate which sheet a ${} replacement error comes from
pyxform v2.0.3

#### Problem description
I started an instance expression with `${` (`${instance('filename')/root/item`). Naturally, that doesn't work so I got an error: `There has been a problem trying to replace ${instance('filename')/root/itemwith the XPath to the survey element named 'instance('filename')/root/item'. There is no survey element with this name.`

This was on the `entities` sheet so it took me a while to find. It's not the first time this has happened to me. It has also happened from the settings sheet because I sometimes write an instance_name expression and then change the form structure without updating that expression.

Include a badly-formed `${}` expression on the entities or settings sheet.

#### Expected behavior
There's an indication of the sheet the error is coming from.

#### Other information 
Things you tried, stack traces, related issues, suggestions on how to fix it...

lindsay-stevens:
Closes #714

For description of approach see commit messages (mainly 64a2a37). PR also includes a couple of minor tidy-ups.

An alternative might be to pass through context info such as row number to the SurveyElement / Survey classes, or assume the sheet name from the class type (e.g. Questions come from the survey sheet), or the column name from the attribute (e.g. Question.hint comes from `hint` (per language though)) . The main problem with that is that it would significantly change the internal data structure expected to pass through `builder.py` which would affect Ona who uses that as a storage format. Also generally have been trying to move validation (especially user-facing) out of SurveyElement into xls2json where the source data context is readily available, and to separate that validation from the transformations (to XML).

The tests for the main #714 related changes are mainly additive so should be minimal. In 0a7bb72 the expressions lexer/parser was replaced (see commit message for details) but a test suite for the old one was added in 8b41399 and the new one meets that plus all other tests - so that approach should mitigate regression risk.

No

### DIFF
--- a/pyxform/xls2json.py
+++ b/pyxform/xls2json.py
@@ -790,9 +790,6 @@ def workbook_to_json(
                 )
         question_name = str(row[constants.NAME])
         if not is_xml_tag(question_name):
-            if isinstance(question_name, bytes):
-                question_name = question_name.decode("utf-8")
-
             raise PyXFormError(
                 f"{ROW_FORMAT_STRING % row_number} Invalid question name '{question_name}'. Names {XML_IDENTIFIER_ERROR_MESSAGE}"
             )


lognaturel (approved):
I can't say I carefully looked at every change at https://github.com/XLSForm/pyxform/pull/789/changes/15ff6879f38a272ef433c4701c0d3e21d8de8b51 but test coverage looks sufficient. I think our best course of action is merging and seeing if anyone reports any issues.

This will be a really nice change, I just ran into the issue again a couple days ago.

PR #792, opened 2025-12-10
fix: improve error message when entity repeat doesn't exist (fixes #784)

### LINKED ISSUE #784: Improve error message when Entity repeat specified doesn't exist
When declaring an Entity, if you reference a repeat name that doesn't exist in the `repeat` column, you get an error like:

> On the 'survey' sheet, the 'save_to' value 'xyz' is invalid. The entity property populated with 'save_to' must be inside of the entity repeat.

This is technically correct but doesn't help find the issue. It would be more helpful to get a message like 

> On the 'entities' sheet, the 'repeat' value is set to '${foo}' but there is no repeat by that name.

Anuragkrosuru:
Closes #784 

### DIFF
--- a/pyxform/entities/entities_parsing.py
+++ b/pyxform/entities/entities_parsing.py
@@ -346,6 +346,7 @@ def validate_entities_columns(row: dict):
 def validate_entity_repeat_target(
     entity_declaration: dict[str, Any] | None,
     stack: Sequence[dict[str, Any]] | None = None,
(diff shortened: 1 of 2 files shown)


Anuragkrosuru:
I've created PR #792 to address this issue. The PR adds validation to check if the entity repeat exists in the survey and provides a clearer error message pointing to the entities sheet.

PR #795, opened 2025-12-16
747: update identifier validation message now that leading colons aren't allowed

### LINKED ISSUE #747: Update identifier validation message now that leading colons aren't allowed
Follow up to https://github.com/XLSForm/pyxform/issues/630

lindsay-stevens:
Closes #747

The majority of these commits are organising error/warning messages into the `errors.py` module, which is what I had in mind for the `ErrorCode` pattern which was added to pyxform a few months ago. After this PR about 30 error messages are organised there, but pyxform raises an error at about 170 locations and many of those have distinct messages.

The goal of this stage of organisation is to try and identify all the errors/warnings which deal with identifiers, so that the validation approach and message content can be more consistent with each other and with the documentation. Also to take the opportunity to move over error/warning messages that are already using the `errors.Detail` class or seemed easy to move. There are still some name-related errors that have not been moved (e.g. `Audits must always be named 'audit.'`, etc), but they don't specifically deal with identifier grammar. There is also still a name-like regex check for the "app" parameter but that requires an android package name which has different rules to XML names.

If library or script usages were looking for specific error strings then these changes would break them. But putting the errors into an enum/object should make it easier to avoid looking for specific strings - as is shown in the tests that use `ErrorCode`.

Maybe? #630 says ODK tools don't support having a colon as the first character in identifier names but that doesn't seem to be stated in the ODK XForms Specification. On one hand maybe it is not a hard requirement from a spec perspective, but on the other hand it is functionally required via pyxform validation and not being supported by ODK form clients.

### DIFF
--- a/pyxform/errors.py
+++ b/pyxform/errors.py
@@ -58,7 +58,7 @@ class ErrorCode(Enum):
     """
 
     ENTITY_001 = Detail(
-        name="Invalid entity repeat reference",
+        name="Entities - invalid entity repeat reference",
         msg=(
             "[row : 2] On the 'entities' sheet, the 'repeat' value '{value}' is invalid. "
             "The 'repeat' column, if specified, must contain only a single reference variable "
@@ -66,28 +66,28 @@ class ErrorCode(Enum):
         ),
     )
     ENTITY_002 = Detail(
-        name="Invalid entity repeat: target not found",
+        name="Entities - invalid entity repeat: target not found",
         msg=(
             "[row : 2] On the 'entities' sheet, the 'repeat' value '{value}' is invalid. "
             "The entity repeat target was not found in the 'survey' sheet."
         ),
     )
     ENTITY_003 = Detail(
-        name="Invalid entity repeat: target is not a repeat",
+        name="Entities - invalid entity repeat: target is not a repeat",
         msg=(
             "[row : 2] On the 'entities' sheet, the 'repeat' value '{value}' is invalid. "
             "The entity repeat target is not a repeat."
         ),
     )
     ENTITY_004 = Detail(
-        name="Invalid entity repeat: target is in a repeat",
+        name="Entities - invalid entity repeat: target is in a repeat",
         msg=(
             "[row : 2] On the 'entities' sheet, the 'repeat' value '{value}' is invalid. "
             "The entity repeat target is inside a repeat."
         ),
     )
     ENTITY_005 = Detail(
-        name="Invalid entity repeat save_to: question in nested repeat",
+        name="Entities - invalid entity repeat save_to: question in nested repeat",
         msg=(
             "[row : {row}] On the 'survey' sheet, the 'save_to' value '{value}' is invalid. "
             "The entity property populated with 'save_to' must not be inside of a nested "
@@ -95,7 +95,7 @@ class ErrorCode(Enum):
         ),
     )
     ENTITY_006 = Detail(
-        name="Invalid entity repeat save_to: question not in entity repeat",
+        name="Entities - invalid entity repeat save_to: question not in entity repeat",
         msg=(
             "[row : {row}] On the 'survey' sheet, the 'save_to' value '{value}' is invalid. "
             "The entity property populated with 'save_to' must be inside of the entity "
@@ -103,7 +103,7 @@ class ErrorCode(Enum):
         ),
     )
     ENTITY_007 = Detail(
-        name="Invalid entity repeat save_to: question in repeat but no entity repeat defined",
+        name="Entities - invalid entity repeat save_to: question in repeat but no entity repeat defined",
         msg=(
             "[row : {row}] On the 'survey' sheet, the 'save_to' value '{value}' is invalid. "
             "The entity property populated with 'save_to' must be inside a repeat that is "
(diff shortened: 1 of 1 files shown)


lognaturel:
> that doesn't seem to be stated in the ODK XForms Specification

Not explicitly, you're right, that's deferred to the underlying specs, in this case going down to the XML spec. I think these two StackOverflow explanations make it clear that colons should not be allowed as leading characters: https://stackoverflow.com/questions/40445735/is-a-colon-a-legal-first-character-in-an-xml-tag-name/40447829

pyxform/errors.py
```diff
@@ -59,28 +59,36 @@ class ErrorCode(Enum):
...
-    NAMES_005 = Detail(
+    NAMES_005: Detail = Detail(
         name="Invalid duplicate meta name in the survey",
         msg=(
             "[row : {row}] On the 'survey' sheet, the 'name' value 'meta' is invalid. "
             "The name 'meta' is reserved for form metadata."
         ),
     )
+    NAMES_006: Detail = Detail(
+        name="Invalid missing name in the choices sheet",
```
lognaturel: Could this just be "Missing name"?

pyxform/errors.py
```diff
@@ -75,6 +75,22 @@ class ErrorCode(Enum):
             "Learn more: https://xlsform.org/en/#setting-up-your-worksheets"
         ),
     )
+    INTERNAL_001: Detail = Detail(
+        name="Internal error: Incorrectly Processed Question Trigger Data",
```
lognaturel: Should be sentence case to match other names.

pyxform/errors.py
```diff
@@ -178,6 +178,13 @@ class ErrorCode(Enum):
             "'{q}' appears more than once."
         ),
     )
+    SURVEY_001 = Detail(
+        name="Survey Sheet Unmatched Group/Repeat/Loop End",
```
lognaturel: Should be sentence case to match other names

pyxform/errors.py
```diff
@@ -185,6 +185,13 @@ class ErrorCode(Enum):
             "No matching 'begin_{type}' was found for the name '{name}'."
         ),
     )
+    SURVEY_002 = Detail(
+        name="Survey Sheet Unmatched Group/Repeat/Loop Begin",
```
lognaturel: Sentence case

pyxform/errors.py
```diff
@@ -49,6 +49,13 @@ class ErrorCode(Enum):
             "(like '${{q1}}'), and the reference variable must contain a valid name."
         ),
     )
+    ENTITY_002 = Detail(
```
lognaturel: This one and the one above are interesting because they will likely be replaced very shortly when we remove the `repeat` column. I wonder whether we should consider using that knowledge and giving them later numbers. But really having gaps is fine too!

tests/test_choices_sheet.py
```diff
@@ -578,3 +578,19 @@ def test_missing_name__error(self):
             errored=True,
             error__contains=[ErrorCode.NAMES_006.value.format(row=2)],
         )
+
+    def test_name_not_validated_as_xml_name(self):
+        """Should not raise an error if a name has invalid XML name characters."""
```
lognaturel: I can't figure this out! Why not? Or is this just demonstrating current behavior?

pyxform/survey_element.py
```diff
@@ -514,7 +514,7 @@ def xml_label_and_hint(self, survey: "Survey") -> list["DetachableElement"]:
             and "big-image" in self.media
         ):
             raise PyXFormError(
-                "To use big-image, you must also specify an image for the survey element named {self.name}."
+                f"To use big-image, you must also specify an image for the survey element named {self.name}."
```
lognaturel: Ah, very interesting. Relates to some uneasiness I have about what we're testing, exactly.

lognaturel (commented):
Thanks for breaking this down into easy commits -- the overall diff felt impossible to review meaningfully!

Overall having a central error message registry feels like a good idea in this context. I have a few notes inline you may or may not want to act on.

I have some uneasiness about the fact that tests compare against enum references, not the rendered user-facing strings. Things like formatting bugs could ship while all tests still pass and we could fail to notice that an error message became wrong in certain contexts used. Maybe for some really common or critical error messages we could assert on literal strings? I don't have any specific candidates in mind currently but wanted to share this thought in case it gives you ideas.

lindsay-stevens:
> I have some uneasiness about the fact that tests compare against enum references, not the rendered user-facing strings.

Agree it has that potential weakness but each of the tests involves using the same context items, so the risks seem to be a) there is an unused token in the template which makes the result look weird, or b) checking the rendered string with a test case during development was overlooked. Maybe each message could have a "rendering check" test to at least casually confirm what it looks like with realistic context values. Also, the `_ErrorFormatter` will put "unknown" in place of any missing token values so for problem a) the user sees a somewhat cryptic but not totally broken looking message or a runtime error.

The current approach improves on the previous patterns of a) only checking part of the error/warning string and therefore facing the same rendering problems as above, b) only checking part of the message and matching some other message. By using assigned symbols we can more easily see exactly where an error/warning is used and tested, both for coverage and for not having to manually find and update any whole or partial error/warning messages that may be relevant to current work.

One of the next steps could be working on the language of the messages, and so far the format I've been trying to follow is roughly "This is where the problem is. This is why it's a problem. This is what you can do to fix it." Sort of like how a python traceback does context, details, tips in that order. Many messages have been collected in this PR but there are still tons still inline amongst the code, and perhaps they could be reviewed for consistency with Central/Collect/Webforms message phrasing, so maybe that all could be a follow up task? Also could be extended to localisation.

pyxform/errors.py
lindsay-stevens: Addressed in commit 05f4fae2

lindsay-stevens:
Thanks!

PR #818, opened 2026-02-11
800: Allow declaring that multiple Entities should be created/updated

### LINKED ISSUE #800: Allow declaring that multiple Entities should be created
This is a placeholder issue pending establishing scope. Example considerations:
- [ ] Do we start only with multiple Entities as long as they're all defined in different repeat scopes?
- [ ] How do we help make sure that the data is easily linkable in analysis and future form-based usage?

### LINKED ISSUE #819: Don't output Entity id setvalue when there's an entity id expression specified
See [forum thread](https://forum.getodk.org/t/create-and-update-entities-from-repeats/55163/25?u=ln).

https://github.com/XLSForm/pyxform/blob/v2.0.0/pyxform/entities/entity_declaration.py#L67-L68

Explicitly outputs a `setvalue` when a `create_condition` is present. However, that's done with the assumption that the `entity_id_expression` will take precedence. Let's not output the `setvalue` if we expect it not to have impact.

lindsay-stevens:
Closes #800
Closes #819

Initial implementation that passes existing relevant tests and a handful of new ones but still needs:

- [x] validations to be added
- [x] error messages tidied up
- [ ] much more + thorough tests
  - mainly solver regression suite remaining but current tests cover many common patterns
  - remaining existing tests in test_create_repeat.py may be better covered by solver suite
- [ ] more docstrings / comments
  - could be more thorough still but is reasonably clear at the moment
- [x] resolve todo's
- [x] probably some restructuring to improve readability / maintainability e.g.
  - replace mysterious dict/tuple/etc with types (with class methods etc)
  - redundant data processing / call paths

As noted in below comments, the index to this PR is the docstring at the top of `test_entities.py` which assigns a short code to each requirement from the XForm spec, for validations, and for general behaviour. These short codes are then used to tag the relevant test cases. 

New feature but main regression risk is breaking forms for the old entity specs

Yes, there's a spec doc update already but XLSForm side not documented yet

### DIFF
--- a/tests/entities/test_entities.py
+++ b/tests/entities/test_entities.py
@@ -2682,8 +2682,9 @@ class TestEntitiesOutput(PyxformTestCase):
         # ES003 ES005
         md = """
         | survey |
-        | | type | name | label | save_to |
-        | | text | q1   | Q1    | e1p1    |
+        | | type         | name | label | save_to |
+        | | text         | q1   | Q1    | e1p1    |
+        | | csv-external | e1   |       |         |
 
         | entities |
         | | list_name | label | entity_id |
@@ -2701,10 +2702,11 @@ class TestEntitiesOutput(PyxformTestCase):
         # ES003 ES005
         md = """
         | survey |
-        | | type        | name | label | save_to |
-        | | begin_group | g1   | G1    |         |
-        | | text        | q1   | Q1    | e1p1    |
-        | | end_group   | g1   |       |         |
+        | | type         | name | label | save_to |
+        | | begin_group  | g1   | G1    |         |
+        | | text         | q1   | Q1    | e1p1    |
+        | | end_group    | g1   |       |         |
+        | | csv-external | e1   |       |         |
 
         | entities |
         | | list_name | label | entity_id |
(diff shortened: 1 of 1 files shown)


lindsay-stevens:
@ln thanks! I think something that would be good to review initially is the traceability test suite plan at the top of `test_entities.py`. I'm still refining it, but the gist of it is there. What I was going for is a way to organise the test suite for entities, in a lightweight way that can map each test to the relevant requirement(s) and vice versa. There are fancy ways to do this sort of thing with method decorators or gherkin libraries etc. - but as it is currently, at least it allows (in Pycharm or similar) to double-click or search each requirement tag (e.g `ES001`) to see where the relevant tests are.

Another thing that would be good to review early is the `ENTITY_009`, `ENTITY_011`, and `ENTITY_012` errors for topology problems. With those cases, it seemed like the sort of thing where it may be quite confusing for users to resolve for non-trivial forms - even if the message includes every bit of context available (which may be too verbose) where the error is raised - because ultimately pyxform can't know exactly what the user intended to do. Maybe there is a good docs URL that can do the heavy lifting on explaining, rather than a paragraph-sized error? etc.

PR #830, opened 2026-03-16
829: allow entities allocation to consider sibling and/or descendant hierarchies

lindsay-stevens:
For now just a draft with test cases created for investigation of #829 

TBC

TBC

TBC


lognaturel:
Do you think this is worth investigating further? I still don't think there are likely common cases that are being missed.

PR #836, opened 2026-04-11
Insert itemset declaration for range labels

### LINKED ISSUE #835: Generate child itemset for range question with tick labelset
Currently when an XLSForm specifies a `tick_labelset` parameter, that is included in the resulting webform as a `tick-labelset` attribute. However, we made a change to the spec to instead include an itemset child: https://github.com/getodk/xforms-spec/issues/336

lognaturel:
Closes https://github.com/XLSForm/pyxform/issues/835

Verified with Collect v2026.2.0 beta 1 using [rangelabels.xlsx](https://github.com/user-attachments/files/26647989/rangelabels.xlsx)

In xls2json I ended up removing the `tick_labelset` parameter from the json representation and injecting an itemset [using `add_choices_info_to_question`](https://github.com/XLSForm/pyxform/pull/836/changes). That feels like a reasonable resulting json representation and a good way to reuse existing functionality.

For the conversion to xform, I considered whether I should reuse functionality between `MultipleChoiceQuestion` and `RangeQuestion`, maybe pulling out a mixin to represent having items. In the end [I handled itemsets specifically for range](https://github.com/XLSForm/pyxform/pull/836/changes) because choices for selects are more complex. I think we may want to revisit this when we support reference geometry for geopoint/trace/shape which will also involve an itemset child.

I introduced a [`range_itemset`](https://github.com/XLSForm/pyxform/pull/836/changes#diff-3e3971aa9b2cf59967fa85e28a4280c67ffd90c1c8d6df354066b50a08245c33R199) testing function. It just checks the itemset declaration and not the label and value children, that didn't feel necessary.

Limited to the `range` question type, probably really only around tick labels. Feels quite safe to me.

No.


yanokwa (approved):
LGTM

Issue #812, opened 2026-01-13
Leading and trailing spaces in parameters lead to form error


yanokwa:
Works
> label=foo, value=bar
[good-form.xlsx](https://github.com/user-attachments/files/24601588/good-form.xlsx)

Does not work due to leading/training spaces around =
> label = foo, value = bar
[bad-form.xlsx](https://github.com/user-attachments/files/24601587/bad-form.xlsx)
> 
>  Error: [row : 2] Parameter 'value' has a value which is not valid. Values must begin with a letter or underscore. Subsequent characters can include letters, numbers, dashes, underscores, and periods.

Feels like we should trim or ignore the spaces

PR #840, opened 2026-04-25
Leading and trailing spaces in parameters lead to form error #812

Garen6:
Why is this the best possible solution? Were any other approaches considered?
This fixes the bug at the source: parameters_generic.parse() was intentionally preserving case for value and label, but it was also preserving leading and trailing whitespace for those two keys only. That caused forms like value = VAL , label = lBl to fail later with an invalid-name error when used in select_one_from_file / select_multiple_from_file.

The chosen fix keeps the existing intended behavior, which is:

preserve case for value and label
trim incidental outer whitespace the same way other parameters are normalized
I considered leaving parsing unchanged and adding a later validation special case, but that would keep the bug’s root cause in the generic parser and make downstream handling more brittle. Normalizing the values once in the parser is smaller, clearer, and benefits every caller consistently.

What are the regression risks?
Low. The change is narrowly scoped to case-sensitive parameter values in parameters_generic.parse(). It does not change:

parameter key parsing
separator handling
case normalization for non-case-sensitive parameters
validation rules for supported/unsupported parameter names
The main regression risk would be accidentally changing the intended case-preservation behavior for value and label, so I added a regression test to confirm that case is still preserved while outer whitespace is removed.

Does this change require updates to documentation? If so, please file an issue [here](https://github.com/XLSForm/xlsform.github.io) and include the link below.
I do not think this requires a docs update. This change aligns pyxform with the intuitive expectation that incidental leading/trailing spaces around parameter values should not break form compilation.

### DIFF
--- a/pyxform/validators/pyxform/parameters_generic.py
+++ b/pyxform/validators/pyxform/parameters_generic.py
@@ -26,7 +26,9 @@ def parse(raw_parameters: str) -> PARAMETERS_TYPE:
             )
         k, v = param.split("=")[:2]
         key = maybe_strip(k.lower())
-        params[key] = v if key in CASE_SENSITIVE_VALUES else maybe_strip(v.lower())
+        params[key] = (
+            maybe_strip(v) if key in CASE_SENSITIVE_VALUES else maybe_strip(v.lower())
+        )
 
     return params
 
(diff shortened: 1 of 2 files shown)


lognaturel:
Hi @Garen6 and thanks very much for your contributions! We appreciate your patience with us as we work to prioritize review. @lindsay-stevens will taking a look soon.

In the mean time, consider visiting [the ODK forum](https://forum.getodk.org/) if you haven't already. We'd love to learn more about you if you have a moment to [introduce yourself](https://forum.getodk.org/t/introduce-yourself-here/6671). Thank you!

PR #845, opened 2026-06-11
812: Leading and trailing spaces in parameters lead to form error

### LINKED ISSUE #812: Leading and trailing spaces in parameters lead to form error

lindsay-stevens:
Closes #812

As outlined in the commit messages:

> change from regex etc to a lark grammar, which is easier to read
    and could support enhancements like quoted parameter values to
    handle values containing delimiters e.g. "a = 'b c', x = y"

Also goes some way towards #592 in the sense of organising how parameters are handled and tested. I think next steps there could look like 1) moving the parameter handling or validation code blocks out of xls2json.py into a different module(s) or classes, and 2) upgrading the error messages to the new pattern that includes the row number; but these seemed like deviating too far from the motivating issue #812.

Primarily I think the changes to `constants.py`, but that module is not exported at the project level `__init__.py`. I otherwise made minimal compatibility changes to the existing tests to try and avoid regressions.

Not directly, but I noticed that the current XLSForm reference template is missing many accepted parameters (perhaps intentionally). Hopefully collecting the parameter keys in one place will now make it easier to keep the docs in sync.


tests/test_image_app_parameter.py
```diff
@@ -48,31 +51,6 @@ def test_throwing_error_when_invalid_android_package_name_is_used_with_supported
                         ],
                     )
 
-    def test_throwing_error_when_blank_android_package_name_is_used_with_supported_appearances(
```
lognaturel, line 51: Is this now just handled generically as a blank parameter? I have read this test a hundred times and I can't seem to convince myself it had any value so maybe that's why you removed it!

pyxform/xls2json.py
```diff
@@ -1033,54 +1018,54 @@ def workbook_to_json(
...
-                    accepted=select_params_allowed,
-                    row_number=row_number,
-                )
+                    qt_params = constants.ParametersSelectFromFile
+                    pv.validate(
+                        parameters=parameters,
+                        accepted=qt_params,
+                        row_number=row_number,
+                    )
+                    if "value" in parameters:
```
lognaturel: Any particular reason not to introduce constants here and for label?

lognaturel (approved):
Nice! Couple of questions that could prompt action but overall this looks ready to merge and build on.

pyxform/xls2json.py
lindsay-stevens: Just missed this bit (and also `text` param `rows) - fixed in 9fdd143

tests/test_image_app_parameter.py
lindsay-stevens: Yes this test was deleted because the case is a generic parsing failure, now handled by test_parameters.py (though that only had the first one case - added the second one in 9fdd143)

lognaturel (approved):
Fab!

PR #847, opened 2026-06-17
839: Add support for reference geometry for geo questions

### LINKED ISSUE #839: Add support for reference geometry for geo questions
https://forum.getodk.org/t/display-existing-geometry-when-capturing-new-point-trace-shape-on-map/57427/22?u=ln

lindsay-stevens:
Closes #839

- Initial pass for feedback
- Includes refactoring to clarify existing nodeset generation

The refactoring may have missed some usage pattern that is not currently in the test suite.

Spec updates TBC https://github.com/getodk/xforms-spec/issues/343 but outline in linked forum post.

### DIFF
--- a/pyxform/constants.py
+++ b/pyxform/constants.py
@@ -194,11 +194,13 @@ class ParametersAudit(StrEnum):
 class ParametersGeo(StrEnum):
     ALLOW_MOCK_ACCURACY = "allow-mock-accuracy"
     INCREMENTAL = "incremental"
+    REFERENCE_GEO = "reference-geo"
 
 
 class ParametersGeoPoint(StrEnum):
     ALLOW_MOCK_ACCURACY = "allow-mock-accuracy"
     CAPTURE_ACCURACY = "capture-accuracy"
+    REFERENCE_GEO = "reference-geo"
     WARNING_ACCURACY = "warning-accuracy"
 
 
--- a/pyxform/constants.py
+++ b/pyxform/constants.py
@@ -136,6 +136,7 @@ AUDIO_QUALITY_NORMAL = "normal"
 AUDIO_QUALITY_EXTERNAL = "external"
 
 EXTERNAL_INSTANCE_EXTENSIONS = {".xml", ".csv", ".geojson"}
+EXTERNAL_INSTANCE_TYPES = {"csv-external", "xml-external"}
 
 DEFAULT_ITEMSET_LABEL_REF = "label"
 DEFAULT_ITEMSET_VALUE_REF = "name"
--- a/pyxform/constants.py
+++ b/pyxform/constants.py
@@ -194,13 +194,13 @@ class ParametersAudit(StrEnum):
 class ParametersGeo(StrEnum):
     ALLOW_MOCK_ACCURACY = "allow-mock-accuracy"
     INCREMENTAL = "incremental"
-    REFERENCE_GEO = "reference-geo"
+    REFERENCE_GEOMETRY = "reference-geometry"
 
 
 class ParametersGeoPoint(StrEnum):
     ALLOW_MOCK_ACCURACY = "allow-mock-accuracy"
     CAPTURE_ACCURACY = "capture-accuracy"
-    REFERENCE_GEO = "reference-geo"
+    REFERENCE_GEOMETRY = "reference-geometry"
     WARNING_ACCURACY = "warning-accuracy"
 
 
--- a/pyxform/errors.py
+++ b/pyxform/errors.py
@@ -522,10 +522,10 @@ class ErrorCode(Enum):
         ),
     )
     SURVEY_006: Detail = Detail(
-        name="Survey sheet - invalid geo parameter 'reference-geo'",
+        name="Survey sheet - invalid geo parameter 'reference-geometry'",
         msg=(
             "[row : {row}] On the 'survey' sheet, the 'parameters' value is invalid. "
-            "For geo questions, the 'reference-geo' parameter must be either: "
+            "For geo questions, the 'reference-geometry' parameter must be either: "
             "a choices list_name, an entities list_name, a reference variable containing "
             "the name of a repeat, or the name of an attached CSV file."
         ),
--- a/pyxform/errors.py
+++ b/pyxform/errors.py
@@ -521,6 +521,15 @@ class ErrorCode(Enum):
             "The following are invalid parameter key(s): '{rejected}'."
         ),
     )
+    SURVEY_006: Detail = Detail(
+        name="Survey sheet - invalid geo parameter 'reference-geo'",
+        msg=(
+            "[row : {row}] On the 'survey' sheet, the 'parameters' value is invalid. "
+            "For geo questions, the 'reference-geo' parameter must be either: "
+            "a choices list_name, an entities list_name, a reference variable containing "
+            "the name of a repeat, or the name of an attached CSV file."
+        ),
+    )
 
 
 class PyXFormError(Exception):
--- a/pyxform/errors.py
+++ b/pyxform/errors.py
@@ -530,6 +530,27 @@ class ErrorCode(Enum):
             "attached file), or be a reference variable containing the name of a repeat."
         ),
     )
+    SURVEY_007: Detail = Detail(
+        name="Survey sheet - invalid geo parameter 'capture-accuracy'",
+        msg=(
+            "[row : {row}] On the 'survey' sheet, the 'parameters' value is invalid. "
+            "The 'capture-accuracy' parameter must have a numeric value."
+        ),
+    )
+    SURVEY_008: Detail = Detail(
+        name="Survey sheet - invalid geo parameter 'warning-accuracy'",
+        msg=(
+            "[row : {row}] On the 'survey' sheet, the 'parameters' value is invalid. "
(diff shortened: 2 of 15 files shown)


pyxform/validators/pyxform/choices.py
```diff
@@ -67,3 +70,53 @@ def validate_and_clean_choices(
                 option.pop(invalid_header, None)
             option.pop("__row", None)
     return choices
+
+
+def add_choices_info_to_question(
```
lognaturel: So far everything in validators feels much more directly validation-related. Not a big deal but wanted to mention that this file is starting to feel like it could be a choices feature slice rather than just a validator.

pyxform/question.py
```diff
@@ -353,8 +353,154 @@ def get_options(self, choices: Iterable[dict]) -> Generator[Option, None, None]:
         self.requires_itext = requires_itext
 
 
+class ItemsetNode:
+    nodeset_template: str = "instance('{}')/root/item"
+
+    def __bool__(self):
+        return True
+
+    def __init__(
```
lognaturel, line 368: Consider using `@dataclass`?

pyxform/constants.py
```diff
@@ -194,11 +194,13 @@ class ParametersAudit(StrEnum):
 class ParametersGeo(StrEnum):
     ALLOW_MOCK_ACCURACY = "allow-mock-accuracy"
     INCREMENTAL = "incremental"
+    REFERENCE_GEO = "reference-geo"
```
lognaturel: I forgot we had all these existing parameters that use dashes because they match the XForm convention. Typically we use underscores for new parameter names but I agree with your instinct to match the others in this context.

I forgot to update the issue -- Yaw has requested we use the full `reference-geometry` which does fit better with the other parameters as well.

pyxform/xls2json.py
```diff
@@ -407,7 +407,9 @@ def workbook_to_json(
     question_names = set()
     element_names = Counter()
     trigger_references: list[tuple[str, int]] = []
-    repeat_names = set()
+    geo_references: list[tuple[str, int]] = []
+    csv_sources: set[str] = set()
```
lognaturel: It could also be XML or geojson sources -- any named secondary instance -- whether it's internal or external or what the file format is shouldn't matter.

I thought pyxform would have a way to list these but now that I think about it we don't validate the parameter to `instance` calls or anything like that, do we? Can we rely on the code that outputs all the instance declarations by any chance?

pyxform/errors.py
```diff
@@ -521,6 +521,15 @@ class ErrorCode(Enum):
             "The following are invalid parameter key(s): '{rejected}'."
         ),
     )
+    SURVEY_006: Detail = Detail(
+        name="Survey sheet - invalid geo parameter 'reference-geo'",
+        msg=(
+            "[row : {row}] On the 'survey' sheet, the 'parameters' value is invalid. "
+            "For geo questions, the 'reference-geo' parameter must be either: "
+            "a choices list_name, an entities list_name, a reference variable containing "
+            "the name of a repeat, or the name of an attached CSV file."
```
lognaturel: "an attached CSV file" -> "an attached file" (could be XML or geojson)

tests/test_geo.py
```diff
@@ -203,3 +204,407 @@ def test_with_incremental__wrong_type_no_params__ok(self):
...
+                    ],
+                )
+
+    def test_with_reference_geo__select_from_file__ok(self):
+        """Should find that a child itemset is emitted, with default value/label."""
+        md = """
+        | survey |
+        | | type                         | name | label | parameters           |
+        | | select_one_from_file s1.csv  | q1   | Q1    |                      |
+        | | {type}                       | q2   | Q2    | reference-geo=s1.csv |
```
lognaturel: I think it would be more consistent with other usage to use the instance name without the extension. It feels kind of similar to writing an `instance('s1')` expression where we don't use the extension.

It's true that there's an extension for `select_one_from_file s1.csv` but that feels to me like it's because we conflated attaching and querying the file. For `reference-geometry` the file must already be attached.

tests/test_geo.py
```diff
@@ -203,3 +204,407 @@ def test_with_incremental__wrong_type_no_params__ok(self):
...
+                        xpq.model_instance_bind("q2", t),
+                        xpq.body_itemset(
+                            q_name="q2",
+                            nodeset="instance('s1')/root/item[region = 1]",
+                            extra_q_assertions="and count(@*) = 1",
+                        ),
+                    ],
+                )
+
+    def test_with_reference_geo__select_one_external__ok(self):
```
lognaturel: I would prefer not to support this case. I think it's very unlikely someone would want it and in general we can treat the DB-backed CSV querying features as legacy.

lognaturel (commented):
Thanks so much! Folks have been trying out the beta implementation in Collect and are loving this functionality.

I know this isn't your favorite spec but I like how you've introduced a layer in the question hierarchy for itemset-capable question types, it makes a lot of sense and led to some useful cleanup.

Details below, the main things are:
* parameter name should be `reference-geometry`
* any external secondary instance should be supported, regardless of the filename extension
* let's drop support for the special itemsets.csv if it's practical

pyxform/validators/pyxform/choices.py
lindsay-stevens: Moved back to xls2json in 3735635 for reasons explained there.

Separately to that yes the sub-package name "validators" is maybe not the most accurate anymore since some of the modules in there deal with processing as well as validation. But at least it breaks down the xls2json logic into digestible pieces, and in the longer term hopefully aligns to doing something like how all the entities stuff is in one package - maybe each question type gets a module (and finally spell the end of question_type_dict 😱), etc.

pyxform/question.py
lindsay-stevens: The `ItemsetNode` is looking a lot like a `SurveyElement` but I was leaving that til the spec for this feature settles.

pyxform/constants.py
lindsay-stevens: Updated the parameter name in b31e271, and noted the preference for underscores in #592

pyxform/xls2json.py
lindsay-stevens: Updated validation and test cases to allow all external instance file types in 7b1799b. Since it became a bit lengthy I put some follow-up questions on scope [in this comment](https://github.com/XLSForm/pyxform/issues/839#issuecomment-4738892367).

pyxform/errors.py
lindsay-stevens: Updated the error message along these lines in 7b1799b

tests/test_geo.py
lindsay-stevens: Updated the usage pattern and tests accordingly in 7b1799b

tests/test_geo.py
lindsay-stevens: In 7b1799b I removed the special case for this from `validate_parameter_reference_geometry`, although the tests show that the select_one_external / external_choices approach doesn't work on it's own anyway - like with entities it needs a csv-external to trigger the auto-generated secondary instance, which allows the nodeset to reference a file named itemsets.csv.

Unless a regression has snuck in somewhere, the effect of select_one_external / external_choices seems to be just to emit an input control with an instance() expression in a `query` attribute - and then presumably Collect does some magic to auto-wire that to itemsets.csv without a secondary instance.

I'm not sure it's OK to block the select_one_external / external_choices usage pattern entirely because a user might have a file called itemsets.csv that they created independently. In other words we could say in an error "you cannot use the secondary instance name `itemsets` with reference-geometry" or "you cannot (do that) if there is an external_choices sheet", which seems potentially noisy?

pyxform/question.py
lognaturel: Ah yes, I was writing comments commit-by-commit and didn't go back to remove the ones that went stale.

tests/test_geo.py
lognaturel: > Collect does some magic to auto-wire that to itemsets.csv without a secondary instance

Exactly.

> a user might have a file called itemsets.csv that they created independently

That makes sense. And in that case it would be the regular CSV file code path, right? I guess what I'm saying is let's not include `itemsets` as a special case for `select_one_external`, let's let it get considered only if there's e.g. an explicit `select_one_from_file itemsets.csv`.

lognaturel:
Thanks, the functionality is looking complete to me, now!

lognaturel (approved):
Looks great! ✨ 

The least satisfying part for me is the validation of secondary instances but any alternative I can think of feels pretty invasive and not worth it. The best I can come up with is some kind of secondary instance registry that every secondary instance related construct could register into.

pyxform/validators/pyxform/question_types/geo.py
```diff
@@ -16,36 +16,47 @@ def validate_parameter_incremental(value: str) -> None:
...
 ) -> None:
-    """Check that the reference-geometry name can be resolved to valid nodeset target."""
-    for target, row_num in referrers:
+    """
+    Check that the reference-geometry name can be resolved to valid nodeset target.
+
+    Does not support secondary instances that would be generated solely by:
+    - pulldata() calls
+    - search() calls
+    - last-saved usages in variables
```
lognaturel, line 35: Good idea to document this here 👍

lognaturel (approved):

PR #849, opened 2026-06-23
Support vertical and no-ticks for new range parameters

### LINKED ISSUE #846: New range parameters should be available for `vertical no-ticks` appearance
[sliderTest-vertical.xlsx](https://github.com/user-attachments/files/28967113/sliderTest-vertical.xlsx)

The last question results in 

> On the 'survey' sheet, the 'parameters' value is invalid. For the 'range' question type, the parameters 'tick_interval', 'placeholder', and 'tick_labelset' are only supported for the appearances 'vertical', 'no-ticks' and the default (empty) horizontal.

We forgot to account for `vertical no-ticks` or `no-ticks vertical`.

lognaturel:
Closes #846

I considered splitting the appearance string into tokens which would be more correct but I think string membership is easier to read and is fine for a short list of well-known tokens.

Small fix so should be minimal. Worst case is probably that the new range parameters are allowed to pair with appearances they're not supposed to pair with.

No

### DIFF
--- a/pyxform/errors.py
+++ b/pyxform/errors.py
@@ -448,8 +448,8 @@ class ErrorCode(Enum):
         msg=(
             "[row : {row}] On the 'survey' sheet, the 'parameters' value is invalid. "
             "For the 'range' question type, the parameters 'tick_interval', 'placeholder', "
-            "and 'tick_labelset' are only supported for the appearances 'vertical', 'no-ticks' "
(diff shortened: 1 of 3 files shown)


lindsay-stevens:
Looks good! Just some minor tidying. Added #850 for the idea on improving appearances handling.

lindsay-stevens (approved):

PR #853, opened 2026-07-13
Dependencies update

lindsay-stevens:

That time of year again for moving up python versions and doing some housekeeping. Due to the `ruff` rule changes it'd be easier to review per-commit.

- Move python support window from `3.10 to 3.13` to `3.11 to 3.14`. Python 3.10 is fully EOL on 2026-10-31, and 3.14 would be stable now having been first released 2025-10-07.
  - There is a backwards incompatible change in f2171c0e but since .xls support will probably be removed with the next release it'll probably be a new major version for pyxform anyway.
- Update project dependencies and GitHub workflow actions versions to latest available.
  - There is a `xlrd==2.0.2` with a minor fix but this dependency is due to be removed soon as noted above.
- Update / apply ruff rules.
  - There are many more rules that might be useful, but basically I was looking for rules that seemed worthwhile and could be applied reasonably cleanly in the space of a couple of hours.
- Converted readme to markdown to simplify maintenance.

Considered adding test matrix runs for windows and linux `arm64` since these platforms are probably becoming more popular. But the GitHub runners currently don't have a `ubuntu-latest-arm` option so it'd be something else to maintain, and the `macOS` build is already testing `arm64` to some extent, and there's already 13 jobs in the `verify.yml` workflow run on every push. Maybe it'd make sense to add as a pre-release check at some point.

As noted above.

No, the relevant docs are updated in this PR.


lognaturel (approved):
Thank you, nice to have these updates!

No issues from my end so merging now so `master` will be ready for the next PRs.

PR #854, opened 2026-07-13
711: remove support for enketo-validate

### LINKED ISSUE #711: Remove support for enketo-validate
The scope would be to remove anything related to enketo-validate from pyxform. This is mainly in `pyxform/validators/`, but there are other references e.g. to enable the validator from `xls2xform.py` and in the main CLI.

Reasons:
- The `pyxform_validator_update` script seems to have been broken since 2020 (since [this enketo-validate release](https://github.com/enketo/enketo-validate/releases/tag/test-build-62)). 
  - The path to of `xmljs.js` was changed from subpath `libxmljsmt` to `node1-libxmljsmt`.
  - The updater script has no way to specify a release tag, i.e. it does not support dodging broken builds
- Enketo-validate dropped support for binary builds in 2022 [in this PR](https://github.com/enketo/enketo-validate/pull/108) as noted for release v1.15.0 [in the changelog](https://github.com/enketo/enketo-validate/blob/64a7780b389c2f487b708971170b0ee1fb128fd4/CHANGELOG.md#1150---2022-05-30).
  - Reason: "Removed binary building as I believed it is not used and too cumbersome to fix and maintain."
  - Binary builds required so that pyxform users don't have to install node, yarn, etc
- ODK is moving from Enketo to https://github.com/getodk/web-forms 
  - even if we did get it working again, it may not be relevant for very long
  - users that want to continue using enketo-validate could continue to use it separately, as they presumably are now

lindsay-stevens:
Closes #711

- Branch based on `deps-update` / PR 853 branch so only commit 70cfdf78 (or later) are relevant.
- Minimal removal of enketo related mentions to retain passing tests, i.e. did not re-evaluate the code paths or tests as to whether they were still the best possible path for supporting only ODK Validate.

If someone was using enketo-validate and providing a build of it themselves, then this is a breaking change. But the issue #711 has been open for two years now and nobody has noted an objection.

No, the only docs for this were the command-line help text, which is now removed.


lognaturel (approved):
Should we think about trying Github's new stacked pull requests feature? https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

lindsay-stevens:
Thanks I will look at stacked PRs for future situations like this, fortunately the current stack seems to apply pretty cleanly.

PR #855, opened 2026-07-15
Drop XLS support

lindsay-stevens:
Closes https://github.com/getodk/security/issues/139

> A deprecation schedule for XLS support was proposed in Dec 2021 [on the forum](https://forum.getodk.org/t/pyxform-xlsform-proposed-deprecation-of-xls-format-support-in-favour-of-xlsx/35974) and the response there and in the https://github.com/XLSForm/pyxform/issues/501 showed no major objections. In the forum thread Yaw noted the following stats for http://xlsform.getodk.org/
>
>> over the last 1.5 months, we've had ~16k forms uploaded and it's been less than 5% XLS
>
> Arguably 5 years is fair warning to drop XLS support and remove the xlrd dependency in the next release. There are many options available for users to convert to XLSX instead if needed (LibreOffice, Google Sheets, MS Office).

Branch based on deps-update / PR 853 branch so only commit https://github.com/XLSForm/pyxform/commit/b5ba83d9db90bd46b7d399f2f8ff69ddcf3f6955 (or later) are relevant.

Current approach:

- [x] Remove / refactor tests using XLS files. There are lots of these, and many of the test cases are covered by other tests. So each XLS file removal is done in a commit explaining where coverage is, or adding it.
- [x] Remove / refactor tests using functions or arguments specific to XLS.
- [x] Remove / refactor source code for XLS.

Initially I started with the last step, but realised it was going to cause a huge diff or multiple commits with failing tests. Also considered just converting the XLS files to XLSX, but I figured that if the files are being touched anyway then it's probably the most relevant time to consider whether that test is still needed or has to use a file (as opposed to the current markdown test style with PyxformTestCase).

Tools that support or expect XLS files would not work any more.

Probably should mention it somewhere, at least on the Central upload page.


lognaturel (approved):
Nice tidying.

Issue #741, opened 2024-12-05
Remove the survey-level option "flat"


lindsay-stevens:
Re: [this code comment](https://github.com/XLSForm/pyxform/blob/6a405088c066461ae0644c977ba6134b69c6bb30/pyxform/survey_element.py#L241):

> The "flat" setting was added in 2013 to support ODK Tables, and results in a data instance with no element nesting. Not sure if still needed.

Confirmed OK to remove in [this PR comment](https://github.com/XLSForm/pyxform/pull/740#discussion_r1870292274):

> I think this can be safely removed. I don't know of anyone using the ODK suite with Tables (part of the ODK-X suite) these days.

So all code for this setting should be removed. It does not appear to be referenced in any documentation.

lindsay-stevens:
@rgraber / @jnm - in [this comment](https://github.com/XLSForm/pyxform/pull/748#issuecomment-2631003955), Rebecca mentioned that kpi is using the `flat` attribute. If it was removed from pyxform it might be tricky to re-implement, but as noted above it's not clear where and how it's being used (and it's not in the ODK docs). Could you please help add some context for this feature with the following questions (or any other comments you might have):

- What is the use case(s) for this attribute, i.e. why/how is a flat instance used in kpi? Is it mainly to simplify XPath expressions?
- Is the `flat` setting presented to kpi users e.g. in documentation or as part of the user interface, etc? 
- Is it expected to be required long term or are there any plans to stop using it (e.g. if it's for backwards compatibility etc.)?

jnm:
Hi @lindsay-stevens, thanks for the heads-up! We do not use `flat` for anything except our [reimplementation](https://github.com/kobotoolbox/kpi/blob/f65a0dcdd8e4d307fb6235cada8d2587ab1a0452/kobo/apps/openrosa/apps/logger/xform_instance_parser.py#L434-L437) of `get_abbreviated_xpath()`, which will not break even if `flat` is removed altogether.

PR #857, opened 2026-07-21
710: csv-external is inside a repeat reports error with xml_instance

### LINKED ISSUE #710: Warn when csv-external is inside a repeat
pyxform v1.x.x, Python v

#### Other information 
Things you tried, stack traces, related issues, suggestions on how to fix it...

lindsay-stevens:
Closes #710

Branch based on deps-update / PR 853 branch so only commit d4149102 and later are relevant.

This PR includes three commits that allow for selecting the preferred approach:

1. The minimal fix + tests
2. A minor refactor to tidy up the relevant code in section.py
3. Adding a validation check + error when a `csv-external` or `xml-external` is inside a repeat, as described in the commit message:

> There doesn't seem to be any reason for it to be allowed, but it does seem like the sort of thing that might be useful in future for scoping instances or entities so disallowing it now leaves that use case open (as opposed to warning (for no strong reason) and then potentially requiring forms to be updated later).

None / low - fixes a reported bug.

Potentially - if `csv-external` and `xml-external` are not allowed in repeats then the docs / XLSForm template could mention that.


pyxform/section.py
```diff
@@ -138,15 +138,15 @@ def xml_instance(self, survey: "Survey", **kwargs):
 
         for child in self.children:
             repeating_template = None
-            if hasattr(child, "flat") and child.get("flat"):
+            if isinstance(child, Section) and child.flat:
```
lognaturel, line 141: Was trying to remember what `flat` was about: https://github.com/XLSForm/pyxform/issues/741

lognaturel (approved):
I feel like we could just allow the declarations in repeats and hoist them up to the top level but I'm happy with the error message here and it will lead to cleaner XLSForms.

PR #858, opened 2026-07-22
763: adding a label for background-geo should not produce body control 

### LINKED ISSUE #763: Specifying a label for a background-geopoint question should not produce a body element
Currently a `background-geopoint` question with a label gets displayed in unexpected ways by clients (see [forum post](https://forum.getodk.org/t/spec-proposal-expose-xforms-value-changed-event-with-odk-setgeopoint-action-in-xlsform/48655/32?u=ln)). We could either ignore the label (my preference) or produce an error.

We should also check `start-geopoint`.

lindsay-stevens:
Closes #763

Branch based on deps-update / PR 853 branch so only commit b6ec9ffd is relevant.

Per the commit message:

> The cause was that in builder.py, the "trigger" tag maps to the TriggerQuestion class, which calls Question._build_xml(), which creates the control element and label/hint. The alternative "action" tag maps to Question, and the base class Question.build_xml() emits None. Another option is to remove the "control" default completely, which would map background-geo to the base Question class, but it was changed to "action" since a) other adjacent types are doing the same, b) the type is basically an action so it is descriptive. Added label test for each "action" question type.

Considered adding a warning if `label` is specified, and then started thinking that maybe there should be a warning if any fields that aren't supported by that type are present. Then I started imagining a giant table of all the supported question types in rows and all the supported columns, where each cell is true/false to indicates whether or not that feature works for that type. Something of that nature would be useful documentation for pyxform (and Collect/Webforms) and as a regression testing tool, but it might be a bit too noisy to have pyxform emit a warning for every unsupported row/column combination.

Low risk, maybe someone was relying on the old behaviour but #763 suggests it was not intended and undesirable.

No (other than as mentioned above).


lognaturel (approved):
Makes a lot of sense!

PR #861, opened 2026-08-20
Detect non-ambiguous dangling XPath operators

### LINKED ISSUE #860: Reject XLSForm expressions ending with non-ambiguous dangling operators

pyxform v4.5.0, Python 3.11

#### Problem description

pyxform accepts expressions that end with an operator that requires a
right-hand operand when external validation is disabled. For example:

| type    | name               | relevant          |
|---------|--------------------|-------------------|
| text    | source_value       |                   |
| integer | dependent_question | ${source_value} = |

Running conversion with `validate=False` or `xls2xform --skip_validate`
successfully generates an XForm containing:

```xml
<bind
    nodeset="/data/dependent_question"
    relevant="/data/source_value ="
    type="int"
/>
```

The expression is incomplete because the operator has no right-hand value or
expression. The same problem applies to dangling arithmetic, boolean, and union
operators.

This postpones the error until `ODK_Validate.jar` flags the generated XForm as
invalid, without identifying the original XLSForm cell.

1. Create an XLSForm containing the survey rows shown above.
2. Run:

   ```bash
   xls2xform form.xlsx form.xml --skip_validate
   ```

   Alternatively, call `convert(..., validate=False)`.

3. Observe that conversion succeeds and the incomplete expression is emitted
   into the generated XForm.

#### Expected behavior

pyxform should raise a `PyXFormError` before XForm generation, regardless of
whether external validation is enabled.

The error should identify the original sheet, workbook row, and column, for
example:

```text
[row : 3] On the 'survey' sheet, the 'relevant' value is invalid.
An operator must be followed by a value or expression.
```

#### Other information

The targeted validation should:

- Detect comparison operators (`=`, `!=`, `<`, `>`, `<=`, `>=`), arithmetic
  operators (`+`, `-`, `div`, `mod`), boolean operators (`and`, `or`), and the
  union operator (`|`), including trailing whitespace.
- Allow terminal `*` because it can be a valid XPath wildcard, as in `/data/*`.
- Allow operator words used as XPath node names, as in `/data/and`.
- Check expression-bearing survey, settings, and entities columns.
- Ignore disabled survey rows.
- Allow quoted operator characters and static text defaults.
- Preserve original sheet and header names in error messages.
- Run when `validate=False` and with `--skip_validate`.

pyxform's existing expression lexer can identify the final token, with context
used to distinguish operator words from XPath node names.

ukanga:
Closes #860

Expressions ending with an operator that requires a right-hand operand are
incomplete, but pyxform currently emits them into the generated XForm when
external validation is disabled. The resulting error is only reported later by
ODK Validate or a form engine, without the original XLSForm sheet, row, and
column information.

This change adds targeted validation before XForm generation. It uses the
existing expression lexer to inspect the final non-whitespace token and raises a
`PyXFormError` for these dangling operators:

- Comparisons: `=`, `!=`, `<`, `>`, `<=`, `>=`
- Arithmetic: `+`, `-`, `div`, `mod`
- Boolean: `and`, `or`
- Union: `|`

Terminal `*` is intentionally excluded because it can be a valid XPath
wildcard, such as `/data/*`. The keyword operators use token context so XPath
node names such as `/data/and` remain valid.

The validation covers:

- Survey bind expressions: `relevant`, `constraint`, `calculation`, `required`,
  and `readonly`.
- `choice_filter`, repeat count, and dynamic defaults.
- The settings `instance_name` expression.
- Entity `entity_id`, `create_if`, `update_if`, and `label` expressions.
- Aliased and normalized headers while retaining original sheet and column
  names in error messages.

It runs when using both `convert(validate=False)` and
`xls2xform --skip_validate`.

The targeted lexer check addresses unambiguously dangling operators while
leaving broader XPath validation unchanged.

The main risk is incorrectly rejecting literal values or XPath node names that
contain operator text. This is mitigated by checking lexer tokens and the
surrounding token context rather than using a string suffix check.

Tests verify that:

- Every included operator is detected with no trailing whitespace, spaces, and
  mixed trailing whitespace.
- Complete expressions and quoted operator characters remain accepted.
- Terminal `*` wildcards and operator-word node names remain accepted.
- Disabled survey rows are ignored.
- Static defaults ending in an operator remain accepted.
- Dynamic defaults and other recognized expression columns are checked.
- Original sheet, row, and aliased column names are reported.
- ODK Validate is not invoked after this internal validation fails.

There are no public API or type-signature changes. The intentional behavior
change is that previously emitted invalid expressions now raise a
`PyXFormError`, including when external validation is skipped.

No documentation updates are required. This is an internal validation
improvement for already invalid expressions.

#### AI/LLM assistance

This PR was developed with assistance from a coding agent and LLM, including
implementation, tests, and supporting text. The submitted changes remain the
responsibility of the contributor.

### DIFF
--- a/pyxform/parsing/expression.py
+++ b/pyxform/parsing/expression.py
@@ -117,6 +117,51 @@ def parse_expression(text: str) -> tuple[Token, ...]:
     return tuple(_EXPRESSION_LEXER.lex(text))
 
 
+_OPERAND_END_TOKEN_TYPES = {
+    "CLOSE_PAREN",
+    "DATE",
+    "DATETIME",
+    "NAME",
+    "NUMBER",
+    "PARENT_REF",
+    "PYXFORM_REF",
+    "PYXFORM_REF_END",
+    "SELF_REF",
+    "SYSTEM_LITERAL",
+    "TIME",
+    "XPATH_PRED_END",
+}
+_WORD_OPERATORS = {"and", "div", "mod", "or"}
+
+
+def _can_end_operand(token: Token) -> bool:
+    """Return whether a token can end the left operand of a word operator."""
+    return token.type in _OPERAND_END_TOKEN_TYPES or (
+        token.type == "OPS_MATH" and token.value == "*"
+    )
+
+
+def ends_with_dangling_operator(text: str) -> bool:
+    """Return whether an expression ends with an operator requiring an operand."""
+    tokens = tuple(
+        token for token in parse_expression(text) if token.type != "WHITESPACE"
+    )
+    if not tokens:
+        return False
+
+    last_token = tokens[-1]
+    token_value = str(last_token)
+    operator = token_value.strip()
+    if operator in _WORD_OPERATORS:
+        # Operator words can also be XPath node names. They are operators only when
+        # preceded by something that can end a left operand. A path separator, for
+        # example, means the word is a node name such as the `and` in `/data/and`.
+        return len(tokens) > 1 and _can_end_operand(token=tokens[-2])
+    if last_token.type in {"OPS_COMP", "OPS_UNION"}:
+        return True
+    return last_token.type == "OPS_MATH" and operator != "*"
+
+
 def is_xml_tag(value: str) -> bool:
     """Check if the input string contains only a valid XML tag / element name."""
     return value and bool(RE_NCNAME_NAMESPACED.fullmatch(value))
--- a/pyxform/errors.py
+++ b/pyxform/errors.py
@@ -187,6 +187,13 @@ class ErrorCode(Enum):
             "questions using these types and the entity list name."
         ),
     )
+    EXPRESSION_001 = Detail(
+        name="Expression - dangling operator",
+        msg=(
+            "[row : {row}] On the '{sheet}' sheet, the '{column}' value is invalid. "
+            "An operator must be followed by a value or expression."
+        ),
+    )
     HEADER_001: Detail = Detail(
         name="Headers - invalid missing header row",
         msg=(
--- a/pyxform/validators/pyxform/expression.py
+++ b/pyxform/validators/pyxform/expression.py
@@ -1,0 +1,101 @@
+"""Targeted validation for expressions in XLSForm workbook cells."""
+
+from collections.abc import Sequence
+from itertools import islice
+from typing import Any
+
+from pyxform import aliases, constants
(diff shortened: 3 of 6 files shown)


lognaturel:
Thanks for filing this, @ukanga. I think it's worth doing and the tests look reasonable. My immediate reaction is that it feels like the validation should be possible to do with an existing pass over fields that can have expressions. @lindsay-stevens can you please take a look when you get a chance?

pyxform/parsing/expression.py
```diff
@@ -117,6 +117,21 @@ def parse_expression(text: str) -> tuple[Token, ...]:
...
+    return (
+        next(
+            (
+                token.type
+                for token in reversed(parse_expression(text))
+                if token.type != "WHITESPACE"
+            ),
+            None,
+        )
+        == "OPS_COMP"
```
lognaturel: Why would dangling comparison operators be more likely than any other dangling operator?

ukanga: Probably not; I started with an example I encountered. Let me check whether I can identify more examples and whether it makes sense to address them in this PR.

ukanga: It seems we have ambiguous paths and non-ambiguous paths. For the non-ambiguous, the proposed changes seem to align ok.

The `*` operator is considered ambiguous and, as such, is not addressed in this change.

I also tested the expression `${q1} = and ${q1} > 0` which is not flagged at any point including with ODK_Validate.jar. Is this valid?
```XML
<bind
      nodeset="/data/q2"
      type="int"
      relevant=" /data/q1  = and  /data/q1  &gt; 0"
  />
```

ukanga: @lognaturel 

The check has been expanded to cover arithmetic (`+`, `-`, `div`, `mod`), boolean (`and`, `or`), and union (`|`) operators in addition to comparisons.  Terminal `*` is excluded because it can be a valid XPath wildcard such as `/data/*`.

lognaturel:
I briefly discussed this with @lindsay-stevens and we'd like to reconsider it after some of his ongoing work lands. He's looking at the general theme of issues around ${} references (https://github.com/XLSForm/pyxform/issues?q=state%3Aopen%20label%3A%22%24%7B%7D%20references%22) and there will be changes to the code this touches. We'll be sure to use the test and implementation as reference once this gets prioritized. Thanks again for filing the issue.

